PR #89, opened 2015-05-06
Fixed a typo

zuron7:
changed
sudo apt-get install nodejs-legacy build-essentatials 
to
sudo apt-get install nodejs-legacy build-essential


frankrousseau:
Thx @zuron7

PR #100, opened 2015-09-28
Update the npmignore

nono:


aenario:
why do you ignore everything ?

aenario:
ah npm ignore, my bad

PR #102, opened 2015-09-30
Fix rename

nono:
When a file was moved vie the files app on cozy, a new file was created on the local filesystem but the old file was still here. Now, this bug is fixed and the move is correctly done.


frankrousseau:
Thank you for the change. I was surprised you removed many tests but I guess you had a good reason :)

nono:
@frankrousseau yes, the `tests/binary.coffee` file was testing a `binary` module that doesn't exist anymore. Its methods have been moved to other modules and the tests (mostly `db`) have been copied to other tests suites.

The most intriguing thing for me is that several of the moved method have a comment that said:

> # TODO find a better module for this.

frankrousseau:
It's because they should not belong to the `db` module. They are operation that are related with the remote database. So a module named `remote_db` or something like this would be more relevant.

PR #103, opened 2015-09-30
Fix a test for config

nono:


frankrousseau:
Your last addition breaks the test you fixed in the prior-to-last commit. Before merging it would be better to fix this.

nono:
Well, I'm not inspired. The tests are green on my machine and I don't see why it fails on docker. It's probably best to just merge #102 for the moment.

PR #105, opened 2015-10-02
Refactoring

nono:
Hi,

I know it' better to avoid big PR, but splitting the db module in two parts, one for Couch and the other for Pouch was necessary. It's not totally done (there are some calls to pouch in the couch module), but it's a good step in this direction.


frankrousseau:
OK. Is it good to merge?

nono:
Travis build is green, you can merge.

frankrousseau:
Thank you @nono

PR #114, opened 2015-10-15
Big refactoring for readonly mode

nono:
Hi,

this is a very large refactoring. It can be merged now. It implements the readonly mode, where cozy-desktop copies files and folders from a remote cozy to a local folder.


Kloadut:
Nice one :-) I'll take a look

nono:
@frankrousseau can you merge this PR? I'll start a new one for the write-only mode.

frankrousseau:
Thank  you @nono. Before starting the write-only mode. Can you make sure that version is published?

nono:
@frankrousseau I'm not sure what you expect from me. Is it to publish on npm? If yes, I don't have the credentials for that.

PR #126, opened 2015-11-03
Write-only / Push mode

nono:
I'm working on the write-only mode, where files and folders from the local hard drive are copied to a remote cozy instance.

@frankrousseau @aenario this PR can be merged now.

The move/rename detection is not yet here, but I think it's better to put it in another PR. For the moment, they are detected as a deletion and a creation, which kind of works event if it's far from being optimal.


frankrousseau:
@nono Should we merge it?

nono:
@frankrousseau yes, I think it's time to merge it

frankrousseau:
Thank you @nono. That looks like a lot of work. Now it's time to POifiy all the features!

PR #142, opened 2015-11-19
Remove code climate

nono:


frankrousseau:
@nono could you explain why you remvoe Code Climate?

nono:
@frankrousseau as I said on trello, it's not really interesting. It doesn't give much than coffeelint + grep -e 'TODO|FIXME|HACK'

frankrousseau:
Thank your for the details. Information on Trello are not shared with the community. That's why it matters to put them here too.

PR #146, opened 2015-11-23
Fix for images on write-only

nono:
When an image is uploaded on the cozy, its metadata are updated in the
CouchDB to include a thumbnail. So, for the write-only mode, it is
expected that the remote revision can sometimes be desynchronized with
the revision stored in the local PouchDB. This commit adds a way to
bypass this desynchronization and allow to update/remove the images.

### DIFF
--- a/backend/remote/index.coffee
+++ b/backend/remote/index.coffee
@@ -102,6 +102,33 @@ class Remote
                     else
                         @couch.remove doc._id, doc._rev, callback
 
+    # Compare two remote docs and say if they are the same,
+    # i.e. can we replace one by the other with no impact
+    sameRemoteDoc: (one, two) ->
+        for field in ['path', 'name', 'creationDate', 'checksum', 'size']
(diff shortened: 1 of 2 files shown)


backend/remote/index.coffee
```diff
@@ -249,12 +275,25 @@ class Remote
         return callback() unless doc.remote
         remoteDoc = @createRemoteDoc doc, doc.remote
         remoteDoc._deleted = true
-        @couch.put remoteDoc, (err, removed) =>
-            if err
-                callback err, removed
-            else
+        async.waterfall [
```
frankrousseau: It's better to write more functions than using `async.waterfall`

nono: @frankrousseau you are right -> https://github.com/nono/cozy-desktop/commit/54a29cbf5bdfdaf0f9da8b0f495674cd8468660d

frankrousseau:
Thank you!

PR #148, opened 2015-11-24
Fix the deletion of a local folder with files in it

nono:


backend/merge.coffee
```diff
@@ -382,12 +384,14 @@ class Merge
     #   - the folder can be found by its _id
     # Actions:
     #   - delete every file and folder inside this folder
+    # TODO add an integration test where a folder with a lot of files is removed
     deleteFolder: (side, doc, callback) ->
         @pouch.db.get doc._id, (err, folder) =>
-            if err
+            if err?.status is 404
```
frankrousseau: Can you add a comment explaining why you send no error when there is a 404?

nono: @frankrousseau done with 97f254e6535603afc10a8773d440e2721f7be254

frankrousseau:
Thx

PR #151, opened 2015-11-25
No push pull

nono:
Using the same metadata database for one time pushing and other time pulling is dangerous. So, I've blocked that and explained how to use COZY_DESKTOP_DIR if you want to play with both modes.

### DIFF
--- a/backend/config.coffee
+++ b/backend/config.coffee
@@ -86,10 +86,14 @@ class Config
         deviceName ?= @getDefaultDeviceName()
         if deviceName and @devices[deviceName]
             old = @devices[deviceName].mode
-            return true if old is mode
-            throw new Error 'Incompatible mode' if old?
-            @devices[deviceName].mode = mode
-            @save()
+            switch
+                when old is mode
+                    true
+                when old?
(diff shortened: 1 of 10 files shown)


backend/config.coffee
```diff
@@ -79,6 +79,20 @@ class Config
...
             null
 
+    # Set the pull or push mode for this device
+    # It wan throw an exception if the mode is not compatible with the last
+    # mode used!
+    setMode: (mode, deviceName) ->
+        deviceName ?= @getDefaultDeviceName()
+        if deviceName and @devices[deviceName]
+            old = @devices[deviceName].mode
+            return true if old is mode
```
frankrousseau: Please do not use return in the middle of a function. It breaks indentation and makes the function "unpredictable".

frankrousseau: See Cozy Guildelines: https://github.com/cozy/cozy-guidelines#return-shortcuts

nono: @frankrousseau 27f9e5f0ab5ba80d9f12ecaba6e5c31c6642b054

frankrousseau:
Thank you @nono!

PR #154, opened 2015-12-01
Case insensitive fs

nono:
On OSX, the file system (HFS+) is case insensitive. So, we can have a conflict for 2 documents from the remote cozy if their paths differ only by the case, like `foo` and `FOO`. This PR is only for detecting these potentiel conflicts. The resolution of conflicts (in a general sense) is for a next PR.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,6 @@ node_js:
     - 5
 env:
     global:
-        - NODE_ENV=test
         - CXX=g++-4.8
 services:
     - couchdb
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 0.5.0 - 2015-11-??
+## 0.5.0 - 2015-11-30
 
 - Add a write-only mode
 - Fix some bugs with the read-only mode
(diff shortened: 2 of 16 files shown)


aenario:
I am not familiar with cozy-desktop architecture, however I see you replace the _id by path in some places, which I suppose lead to document being created with random _id. 

However, I still see a db.get parent in the merger for instance. 
Moreover, if possible, using the (supposedly unique after normalization) path as the _id prevent the need for a secondary index and save some disk space.

Could you explain more about this _id to path change ?

nono:
The `_id` field is not random. It is constructed from the path and have this property: if two filenames leads to the same file, they should have the same id. On OSX for example, the id is the path put in upper case. I've tried to explain that in the design doc: https://github.com/nono/cozy-desktop/blob/case_insensitive_fs/doc/design.md#differences-between-file-systems. It's not clear, or it's just that you haven't look at it yet?

nono:
@aenario do you think you will have time to review it soon?

frankrousseau:
This PR requires a rebase.

nono:
@frankrousseau this PR is already merged in fact. I've started the big PR from the branch for this PR (a particular case of conflict is from two remote files with same name but distinct case synchronized to a desktop with a case-insensitive file system, so I needed it). I thought this PR would be merged before the other PR.

PR #169, opened 2015-12-16
Local mv

nono:
Detect when a file is moved on the local file system to avoid uploading it again.

### DIFF
--- a/backend/local/watcher.coffee
+++ b/backend/local/watcher.coffee
@@ -15,7 +15,6 @@ log      = require('printit')
 # remote operations triggered by the remoteEventWatcher.
 #
 # TODO add unit tests for move/rename
-# TODO increase timeouts for integration tests
 class LocalWatcher
 
     constructor: (@basePath, @prep, @pouch) ->
@@ -129,10 +128,10 @@ class LocalWatcher
     ### Actions ###
 
     # New file detected
-    # TODO check @pending
     onAdd: (filePath, stats) =>
         log.debug 'File added', filePath
         @paths?.push filePath
+        @pending[filePath]?.done()
         @checksums++
         @createDoc filePath, stats, (err, doc) =>
             if err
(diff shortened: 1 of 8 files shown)


backend/local/watcher.coffee
```diff
@@ -29,6 +25,8 @@ class LocalWatcher
     start: (callback) =>
         log.info 'Start watching filesystem for changes'
         @paths = []
+        @pending = Object.create null  # ES6 map would be nice!
```
frankrousseau: I think you could use: `@pending = {}`.

backend/local/watcher.coffee
```diff
@@ -122,17 +129,42 @@ class LocalWatcher
     onAdd: (filePath, stats) =>
         log.debug 'File added', filePath
         @paths?.push filePath
+        @pending[filePath]?.done()
```
frankrousseau: Isn't that to early to mark it done?

backend/local/watcher.coffee
nono: I could, but it's slightly better to use `Object.create null`. For example, if someone add a file called `toString` or `constructor`, it will fail with `@pending = {}`, but not with `Object.create null`.

backend/local/watcher.coffee
nono: Well, it depends. Detecting moves/renames involves a lot of special cases. If I want to cover more of them, I have to add code that is not trivial and will have bugs. So, a balance is to be found.

In this case, if a file is deleted and then a file is added at the same path, I have to first marks the file as deleted in pouchdb before adding the file. Else, the file will be handled as a conflict. There are 2 solutions for that:

- mark the file as deleted right now ;
- defer the create event.

Defering the create event is very difficult to do right and needs a lot of code to support it. You have to take care of many things, like this add event can be related to a move, or defering it can cause deadlock (if you swap two files for example).

So, I've prefered the first solution for simplicity. The only drawback is that there is a short window of time where a move/rename can be not detected: if a file is renamed and just after that a file is created at the source location, it's possible that we will have 2 create and 1 delete events, instead of 1 create and 1 move.

backend/local/watcher.coffee
frankrousseau: Ok, thank you, I didn't know that.

backend/local/watcher.coffee
frankrousseau: Ok, thank you for the additional explanation.

frankrousseau:
Thx.

PR #171, opened 2015-12-17
Improve first replication by copying docs from a view

nono:


frankrousseau:
Why do you listen to change remotely and not locally?

nono:
I'm not sure about this question. This module is remote, so it listens to changes in couchdb to updates the local pouchdb. The `local` module uses chokidar to listens to the file system and updates pouchdb too.

frankrousseau:
So it means, you don't use the replication tool provided by Couch and Pouch. Do you handle it yourself with change feeds?

nono:
Yes, I handle the replication from CouchDB to PouchDB with the changes feed. For PouchDB to CouchDB, the changes are made one by one via a PUT.

I don't use the built-in replication be the issues make it not worth it:
- performance issue on the first replication
- it forces the have the same documents on both sides (and particulary the not meaningful `_id`)
- conflicts handling is a lot harder to do (some conflicts are the CouchDB conflicts but others are not).

frankrousseau:
Ok. Thank you for the PR and additional information.

PR #172, opened 2015-12-17
Improve cli

nono:
- Add a default for devicename when adding a remote
- Ask confirmation before reset-database
- Validate URL for add-remote-cozy

### DIFF
--- a/backend/app.coffee
+++ b/backend/app.coffee
@@ -25,11 +25,20 @@ class App
         @config = new Config @basePath
         @pouch  = new Pouch @config
 
+
     # This method is here to be surcharged by the UI
     # to ask its password to the user
     #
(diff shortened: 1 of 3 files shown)


backend/app.coffee
```diff
@@ -49,6 +62,8 @@ class App
             if err
                 log.error err
                 log.error 'An error occured while registering your device.'
+                if url[0..6] is 'http://'
+                    log.warn 'Did you try with an httpS URL?'
```
frankrousseau: You have a `url` module with node that is proper to parse an url and analyze it.

nono: Right, I will do that.

frankrousseau:
Thx!

PR #175, opened 2015-12-18
Fix for files deleted on both local and remote sides

nono:
If the user deletes a file on its local folder and its cozy when cozy-desktop is stopped, it should not block cozy-desktop. This commit fixes this.


frankrousseau:
Thank you @nono

PR #177, opened 2015-12-21
Fix mobile images

nono:
Fix for images from mobile: they have no checksum and were wrongly detected as conflicts.

Note: this PR also includes commits from #171.


frankrousseau:
Thx

PR #230, opened 2016-03-10
Improve error handling be retrying up to 10 times when an operation fails

nono:


test/unit/sync.coffee
```diff
@@ -217,6 +217,8 @@ describe "Sync", ->
 
 
     describe 'applied', ->
+        @timeout 5000
```
kosssi: why you add this?

nono: By default, mocha has 2s timeout. The options file make it longer (5s), but I prefer to be explicit where tests are taking more than  2s. Here, `applied` can take 3s before calling its callback in case of errors.

kosssi: ok but I change the default timeout to 10s in the [`test/mocha.opts`](test/mocha.opts)

nono: I thought it was 5s, not 10s in this file. And I think we should modify it to come back to 5s. 5s is already much more than we need. And I breaks the flow to have to wait 10s to see a test fails in local when you forgot a callback. For travis, it won't change anything. When the 5s timeout is reached, it's nearly always because something has broken in a previous test.

PR #231, opened 2016-03-11
Add property based testing

nono:
We launch 2 cozy-desktops in two directories, and make them synchronize via a central cozy instance. During ~10 seconds, we do some operations in both directory (like creating files, or moving folders). After that, we wait 20 seconds and we compare the 2 directories. They should be identical: same files and folders.

### DIFF
--- a/src/remote/index.coffee
+++ b/src/remote/index.coffee
@@ -62,18 +62,23 @@ class Remote
                 @couch.put binary, next
             (created, next) =>
                 binary._rev = created.rev
-                @other.createReadStream doc, next
-            (stream, next) =>
-                {_id, _rev} = binary
-                mime = doc.mime or 'application/octet-stream'
-                @couch.uploadAsAttachment _id, _rev, mime, stream, next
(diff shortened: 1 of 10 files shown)


src/remote/index.coffee
```diff
@@ -62,18 +62,23 @@ class Remote
...
+                    # Don't use async callback here!
+                    # Async does some magic and the stream can throw an 'error'
+                    # event before the next async callback is called...
+                    return next err if err
+                    stream.on 'error', -> next new Error 'Invalid file'
+                    {_id, _rev} = binary
+                    mime = doc.mime or 'application/octet-stream'
+                    @couch.uploadAsAttachment _id, _rev, mime, stream, next
         ], (err) =>
+            [cb, callback] = [callback, ->]  # Be sure to callback only once
```
aenario: took me a few seconds to figure out this line, how about
```coffee
callback = null # Be sure to callback only once
# ...
callback? null, binary
```

nono: It won't work. `callback` will always be null when you try to call it.

PR #232, opened 2016-03-11
Add filter in pouch change to limit bandwidth

kosssi:


src/app.coffee
```diff
@@ -1,4 +1,5 @@
 async = require 'async'
+FilterManager = require 'cozy-device-sdk'
```
nono: I don't like class named in SomethingManager. Cf https://blog.codinghorror.com/i-shall-call-it-somethingmanager/

And why filter instead of views, as it is the terminology in CouchDB world?

src/app.coffee
```diff
@@ -62,6 +63,9 @@ class App
                     deviceName: deviceName
                     password: password
                 Devices.registerDevice options, next
+                filterManager = new FilterManager cozyUrl, deviceName, password
+                filterManager.createOrUpdate {file: true, folder: true}, ->
```
nono: `{` and `}` are not necessary in Coffee.

nono: You should warn the user if there was an error while creating/updating the view.

src/local/index.coffee
```diff
@@ -12,9 +12,10 @@ Watcher = require './watcher'
 # It also applied changes from the remote cozy on the local filesystem.
 class Local
     constructor: (config, @prep, @pouch) ->
+        deviceName = config.getDefaultDeviceName()
         @basePath = config.getDevice().path
         @tmpPath  = path.join @basePath, ".cozy-desktop"
-        @watcher  = new Watcher @basePath, @prep, @pouch
+        @watcher  = new Watcher @basePath, @prep, @pouch, deviceName
```
nono: The changes to this file are useless. The local watcher doesn't use `deviceName`.

src/remote/watcher.coffee
```diff
@@ -10,10 +10,12 @@ log   = require('printit')
 # TODO add comments
 # TODO refactor unit tests
 class RemoteWatcher
-    constructor: (@couch, @prep, @pouch) ->
+    constructor: (@couch, @prep, @pouch, deviceName) ->
         @side = 'remote'
         @errors  = 0
         @pending = 0
+        filterManager = new FilterManager '', deviceName, ''
+        @filterName = filterManager.getFilterName()
```
nono: I don't like passing 2 empty strings here. Maybe we can just a class method:

```coffee
@filterName = FilterManager.getFilterName deviceName
```

src/remote/watcher.coffee
```diff
@@ -10,10 +10,12 @@ log   = require('printit')
 # TODO add comments
 # TODO refactor unit tests
 class RemoteWatcher
-    constructor: (@couch, @prep, @pouch) ->
+    constructor: (@couch, @prep, @pouch, deviceName) ->
         @side = 'remote'
         @errors  = 0
         @pending = 0
+        filterManager = new FilterManager '', deviceName, ''
```
nono: `FilterManager` is not required in this file

nono:
I think it should be Couchdb in PR title and commit message, not Pouchdb.

nono:
The couchdb view created by cozy-device-sdk should also be used for initial replication.

https://github.com/cozy-labs/cozy-desktop/blob/master/src/remote/couch.coffee#L55-L66

src/remote/watcher.coffee
```diff
@@ -77,8 +79,7 @@ class RemoteWatcher
                         @whenReady callback
             else
                 @changes = @couch.client.changes
-                    filter: (doc) ->
-                        doc.docType?.toLowerCase() in ['file', 'folder']
+                    filter: @filterName
```
nono: I think it should be:

```coffee
filter: '_view'
view: @filterName
```

By the way, it's one more reason to call it a view and not a filter :wink:

kosssi: why ? I think is more complicate. It's a filter so the function is on filter field I don't like this way to do this.

nono: I don't understand what you find more complicated. It's just how the PouchDB API works. You can see an example on https://pouchdb.com/api.html#filtered-changes (example 5).

kosssi: yes, it's the same but why I pass by view if filter can do it ? Maybe I don't know a things ?

Exemple 3:

    db.changes({
      filter: 'mydesign/myfilter'
    });

Exemple 5:

    db.changes({
      filter: '_view',
      view: 'mydesign/myview'
    });

src/app.coffee
kosssi: if I write this:

    filterManager.createOrUpdate file: true, folder: true, ->

I don't know if folder: true, is first or second parameter no ?

src/local/index.coffee
kosssi: it's for that I don't add @ before deviceName.
You prefer this: 

    @watcher  = new Watcher @basePath, @prep, @pouch, config.getDefaultDeviceName()

src/remote/watcher.coffee
nono: My bad. I thought filters worked only in the local database and only views worked remotely, but I was wrong. After reading pouchdb doc again, I see that `filter` can also work with a remote design document. It's only with a filter that is a raw function (not in a design document) that the replication transfers all the records and filter them client-side.

src/local/index.coffee
nono: I prefer:

```coffee
@watcher  = new Watcher @basePath, @prep, @pouch
```

The Watcher from this file is the local watcher and doesn't use at all the deviceName.

src/app.coffee
nono: It's the first parameter. If you want to use 2 objects for 2 parameters (which is often not a good idea), you have to use `{` and `}`:

```coffee
filterManager.createOrUpdate {file: true}, {folder: true}, ->
```

aenario:
> The couchdb view created by cozy-device-sdk should also be used for initial replication.
> https://github.com/cozy-labs/cozy-desktop/blob/master/src/remote/couch.coffee#L55-L66

In a perfect world yes, however this is poorly optimised in older versions of couchdb and take several dozens minutes for a cozy with some history.

nono:
@aenario I don't understand. What operation can take several dozens minutes for a cozy with some history? Building the view? If yes, what is the alternative?

frankrousseau:
@kosssi and @nono Could you merge this PR? The sprint is ending tomorrow.

kosssi:
ne pas merger

kosssi:
This PR is on pause, because I work on CozyMobile.

We do debug more travis and understand why token is empty. You can see log here https://travis-ci.org/cozy-labs/cozy-desktop/jobs/117455989#L3606 line 3606

kosssi:
It's green :tada: 
thank you @poupotte to find the bug :dancer:

.travis.yml
```diff
@@ -28,6 +28,7 @@ addons:
 before_install:
     - cd
     - export INDEXES_PATH=/home/travis/build/indexes
+    - export NODE_ENV=test
```
nono: You should put it in `env.global` (line 13)

src/app.coffee
```diff
@@ -66,6 +67,12 @@ class App
                     deviceName: deviceName
                     password: password
                 Devices.registerDevice options, next
+            (credentials, next) ->
+                password = credentials.password
+                config = file: true
+                filterSDK.setDesignDoc cozyUrl, deviceName, password, config, \
```
nono: I don't like to use `\`

.travis.yml
kosssi: the same for INDEXES_PATH?

src/app.coffee
kosssi: It's not the standard?
"You can use \ to escape newlines."

.travis.yml
nono: Well, you can put it in `env.global` or let it here. It is only used in `before_install` to start the data-system (it needs a writable path to write its index).

src/app.coffee
nono: I was not clear. I don't like to be a situation that requires to escape a newline. It breaks the flow of reading when you come back on this code later.

Often, it is possible to extract part of the computation in a variable. Maybe `setDesignDoc = filterSDK.setDesignDoc`

nono:
Oops, this branch has conflicts.

@kosssi can you look at it?

PR #238, opened 2016-03-18
Allow users to ignore somes files listed in .cozyignore

nono:
Cozy-desktop can now ignore somes files and folders when doing the synchronization. The users can list what they want to ignore in the `.cozyignore` file. The syntax for this file is the same as `.gitignore`.


kosssi:
:+1:

PR #243, opened 2016-03-23
Lodash.clone is a dependency, not only for testing

nono:


frankrousseau:
@nono tests are broken.

PR #255, opened 2016-03-30
Try register with a suffix when the device name is taken

nono:


src/devices.coffee
```diff
@@ -30,13 +30,21 @@ module.exports =
         client.post 'device/', data, (err, res, body) ->
             if err
                 callback err
+            else if body.error is 'This name is already used'
+                deviceName = "#{options.deviceName}-2"
+                deviceName = deviceName.replace /-(\d+)-2/, (_, suffix) ->
```
kosssi: We make the same in mobile but it's more readable https://github.com/cozy/cozy-mobile/blob/master/www-src/app/replicator/main.coffee#L138, no ?

kosssi:
else :+1:

src/devices.coffee
nono: Well, it adds a new function, which I wanted to avoid. But I think it's more important to be consistent with cozy-mobile. So, let's do the same.

src/devices.coffee
```diff
@@ -18,6 +19,21 @@ module.exports =
                 err = err?.message or body.error or body.message
             callback err
 
+    # Same as registerDevice, but it will try again of the device name is
+    # already taken.
+    registerDeviceSafe: (options, callback) ->
+        module.exports.registerDevice options, (err, credentials) ->
+            if err is 'This name is already used'
+                unless options.tries?
+                    options = clone options
```
kosssi: why you clone it?

nono: To avoid polluting the options that was given to registerDeviceSafe

PR #268, opened 2016-04-05
Add filter in couch to limit bandwidth

kosssi:
see https://github.com/cozy-labs/cozy-desktop/pull/232

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,7 @@ env:
     global:
         - CXX=g++-4.8
         - COZY_DESKTOP_DIR=tmp
+        - NODE_ENV=test
 services:
     - couchdb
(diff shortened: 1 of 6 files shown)


nono:
In the previous PR, I spoke about using the filter for the initial replication. Currently, we are using the views from files to do that. But it means that if files is not installed, cozy-desktop can't work. We should fallback to the filter you added in this case. For the default, I prefer to use the view that the filter: it should be faster.

https://github.com/cozy-labs/cozy-desktop/blob/master/src/remote/couch.coffee#L56-L76

What do you think of this?

nono:
Let's merge this PR. For the initial replication, we'll do the work later.

PR #276, opened 2016-04-11
Don't try to be responsive

nono:


kosssi:
why ?

frankrousseau:
Same remark @nono. Can you add more explanation to your PR (why and what is changed)?

nono:
It makes things look weird on desktop. The menu takes the full screen and is not something usual on desktop.

nono:
Travis is in a bad day. Many tests had timeout. I've restarted them.

PR #318, opened 2016-05-12
Handle device removed

nono:


gui/app/ports.js
```diff
@@ -10,28 +9,26 @@ const pkg = remote.require('./package.json')
...
+    return `${err}`
+  }
+}
 
 const init = () => {
   elmectron.ports.folder.send(defaultDir)
+  elmectron.ports.version.send(pkg.version)
 }
-init()
+setTimeout(init, 10)
```
kosssi: Why you don't make the `init()` on the bottom of the file?

nono: It doesn't change anything. Elm is loaded asynchronously (I think with a `setimediate`).

PR #350, opened 2016-05-25
Fix unlink (bis)

nono:


kosssi:
the build is failed

nono:
I've restarted the build for node 5 (it looks like a temporary error caused by a slowdown of travis)

PR #378, opened 2016-06-17
Use pouchdb-server instead of couchdb for OSX

nono:
The OSX image for travis has no couchdb, and building Couchdb from brew often fails on travis. This is a try to use pouchdb-server instead of couchdb for integration tests on travis for OSX.


nono:
Well, cradle is not compatible with pouchdb-server (`CouchError: bad_request: only_multipart_accepted`)...

PR #412, opened 2016-07-29
Improve the start synchronisation sequence

nono:


src/local/watcher.coffee
```diff
@@ -140,9 +146,13 @@ class LocalWatcher
             else                    "file"
         return [mimeType, fileClass]
 
-    # Get checksum for given file
+    # Put a checksum computation in the queue
     checksum: (filePath, callback) ->
```
kosssi: This function is not used. You can delete it?

nono: The function is used in other files. But, I can use it above.

PR #415, opened 2016-07-29
Fix CSS for errors in wizard

nono:


gui/styles/_wizard.styl
```diff
@@ -148,9 +148,13 @@
...
+  &.step-error .folder__selector
+    background-color: lighten(red, 30%)
 
   .btn
     color #9164F5
 
 .step-error
   background-color red
+  .btn
+    color red
```
kosssi: it's not a problem background and font are red?

nono: Nope. The background of the page is red, and on this page, the button is white with font in red.

PR #474, opened 2016-12-15
Make sure tests are never run with node_modules not matching package.json

sebn:
I want to make sure I could never have a test pass anymore with the wrong node modules installed.

I was planning to use [installed-check[(https://www.npmjs.com/package/installed-check) to implement this, since it seems fast enought not to add too much delay to our test suites.

I'd like the solution to work in all of those **scenarios**:
1. `npm run test`
1. `npm run test-unit` or `npm run test-integration`
1. `mocha test/**/*.js`

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,6 +58,7 @@
     "coffeelint": "1.16.0",
     "del": "2.2.2",
     "faker": "3.1.0",
+    "installed-check": "^2.1.1",
     "jsverify": "0.7.4",
     "lodash.uniq": "4.5.0",
     "mocha": "3.2.0",
--- a/test/helpers/installed_check.coffee
+++ b/test/helpers/installed_check.coffee
@@ -1,0 +1,19 @@
+installedCheck = require 'installed-check'
+
+maybeErrorMessage = (result) ->
+    if result.errors.length
+        """\n
+        =====================================================================
+        Packages installed in ./node_modules/ don't match package.json:
+
+        #{("- #{error}" for error in result.errors).join "\n"}
+
+        =====================================================================
+        """
+
+module.exports = (done) ->
+    installedCheck()
+        .then (result) -> done(maybeErrorMessage(result))
+        .catch done
+    return
+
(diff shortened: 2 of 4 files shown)


sebn:
## 1) Prefix npm scripts with `installed-check && `

I first wanted to prefix both `test-unit` and `test-integration` scripts with `installed-check && `.
It's quite fast compared to our test suites, so the added delay doesn't matter so much.
But it would make it possible for a test to pass when run directly with the `mocha` command.

## 2) Add `--require .../helpers/installed_check` to `mocha.opts`

I then tried to add a `--require test/helpers/installed_check` to `test/mocha.opts` and run the `installedCheck()` function directly at the root of the module, but since it's async, mocha may run some tests before displaying the `installed-check` errors and exiting.

## 3) Root `before` hook

Last, I set up a proper mocha `before` hook in `test/spec_helper`, which works.
The file doesn't seem to be loaded automatically by mocha, like expected by some people.
I was expecting it to break when required explicitely because of the missing `before` function, but it seems to work...

sebn:
I already spent far more time than expected on this.

So now I need to choose between:
- Going with solution 1, and assume one will at least run one of the test suites before pushing
- Going with solution 3, and add missing `require` statements everywhere.

sebn:
Just thinking that I could just set up some git hook to run `installed-check` before commit/push/whatever.

So no need to run it every time locally, and no need to run it on travis...

sebn:
I'll go with solution 1 because:
- I can't imagine myself running a single test, then pushing without even running the related test suite
- I already spent too much time on this

sebn:
`installed-check` breaks build for node 0.10. It would not really be an issue if we stopped supporting it. But I wonder whether we can when node version in debian jessie [is still 0.10](https://packages.debian.org/search?keywords=nodejs&searchon=names&suite=stable&section=all)

sebn:
[Stopped supporting 0.10](https://github.com/cozy-labs/cozy-desktop/commit/9ed775f5bfe967a9710fa2b9222c257a6b249f43), so not an issue anymore.

PR #480, opened 2016-12-22
Update dependencies

sebn:
Included dependencies:

- coffee-script 1.12.2
- electron-builder 10.9.3 (not sure whether the 10.10.0 is officially published yet)
- pouchdb-server 2.0.0

Fixed some JS code style to make `standard` happy.
Also use yarn and build GUI on Travis CI.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -96,6 +96,11 @@ script:
     - npm run build
     - npm run lint
     - ./scripts/coverage.sh npm run test
+    - cd gui
+    - yarn
+    - npm run lint
+    - npm run build
+    - cd ..
 
 after_success:
     - bash <(curl -s https://codecov.io/bash)
--- a/gui/app/main.js
+++ b/gui/app/main.js
@@ -101,7 +101,8 @@ const openCozyFolder = () => {
 const buildAppMenu = () => {
   const template = [
     {
-      label: translate('AppMenu Edit'), submenu: [
+      label: translate('AppMenu Edit'),
+      submenu: [
         { label: translate('AppMenu Undo'), accelerator: 'CmdOrCtrl+Z', role: 'undo' },
         { label: translate('AppMenu Redo'), accelerator: 'Shift+CmdOrCtrl+Z', role: 'redo' },
         { type: 'separator' },
@@ -113,7 +114,9 @@ const buildAppMenu = () => {
       ]
     },
     {
-      label: translate('AppMenu Window'), role: 'window', submenu: [
+      label: translate('AppMenu Window'),
+      role: 'window',
+      submenu: [
         { label: translate('AppMenu Minimize'), accelerator: 'CmdOrCtrl+M', role: 'minimize' },
         { label: translate('AppMenu Close'), accelerator: 'CmdOrCtrl+W', role: 'close' }
       ]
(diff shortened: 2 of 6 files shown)


.travis.yml
```diff
@@ -91,11 +91,20 @@ before_install:
     - pwd
     - mkdir /tmp/cozy-desktop
     - ln -s /tmp/cozy-desktop tmp
+    - npm install -g yarn
```
nono: Travis detects if you have a yarn.lock file, and if it is the case, it installs yarn by default

.travis.yml
```diff
@@ -91,11 +91,20 @@ before_install:
...
+    - cd ..
 
 script:
     - npm run build
     - npm run lint
     - ./scripts/coverage.sh npm run test
+    - cd gui
+    - npm run lint
+    - npm run build
+    - npm run pack
```
nono: What is the goal of running `npm run pack` on travis?

.travis.yml
sebn: I read that it installed the dependencies with yarn, but that it didn't install yarn itself.
I looked for a blog post about the latter, but couldn't find any...

.travis.yml
sebn: First, to know if it works, because I had absolutely no idea whether it does so 😃 

I was also wondering whether `build` could work while `pack` could fail when upgrading electron stuff. Do you think I'm just testing already tested electron tools?

Also I was thinking about having 1 (not 2) test scenario exercising the UI at some point, which would require building the app. But since it's some future plan, we don't really care.

.travis.yml
sebn: Another question: any concern on your side with the `npm install -g` instead of using the ubuntu repos? I was not very comfortable about mixing deb packaged node from third-party repo with nvm... But I'm not comfortable about using a non-recommended installation method either.

sebn: Ok, found in the [yarn doc](https://yarnpkg.com/en/docs/install-ci#travis-tab):

> If it is available, Travis CI will install yarn if necessary, and execute yarn as the default install command.

.travis.yml
sebn: - [x] No need to run `npm run pack`

.travis.yml
sebn: Just force-pushed without the `yarn` call to know further :)

sebn: Automatic yarn install [confirmed](https://travis-ci.org/cozy-labs/cozy-desktop/jobs/186096259#L1548-L1598).

Waiting for the `gui/` stuff.

sebn: - [x] Let Travis installed yarn automatically
- [x] Run yarn manually before building GUI

nono: Travis will install yarn but you still have to install the modules in gui manually

.travis.yml
nono: Well, in my experience, `npm run pack` will often says OK, even when the package is broken. I still had to test manually the packages. In particular, I had some issues with leveldb (the storage part of pouchdb) that has to be recompiled with some options to work with electron. To test that, I installed the package, launched cozy-desktop-gui, and made a synchronization.

sebn:
Canceled the 2nd build and disabled PR builds.

.travis.yml
sebn: Ok, thanks for the insight :)

PR #491, opened 2017-01-20
V3/feature/log remote changes

sebn:
Early feedback welcome.

The feature itself is trivial, but I'm interested in thoughts regarding:

- the way of handling the migration to the cozy stack v3 (disabling/reenabling code, porting or removing integration tests...)
- the way to introduce new methods to cozy client, by first using mixins (maybe with some `cozy.desktop.changes()` prefix? renaming `changes()` to `remoteChanges()`?), then eventually extracting to cozy-client-js...
- the way to organize v3 tests (integration vs isolation, using homemade test data builders instead of an existing lib...)

Remaining work:

- [x] Set up cozy-stack on Travis
- [x] Document remote/cozy mixins setup
- [x] Document test helpers
- [x] Document test data builders
- [x] Pass cozy client once to some builders factory (instead of passing it to every builder call)
- [x] Rename `Cozy*Builder` to `Remote*Builder` and `builders.*()` to `builders.remote*()` (to match the local/remote convention from lib)
- [x] Move each `*Builder` class to its own file
- [x] Possibly move integration tests setup (`before` / `beforeEach`) to some `test/integration/helper.js` file that would be required once in every integration test (since its always the same)?
- [ ] ~~Add a `TODO` notice with future user story reference for disabled v2 code?~~
- [ ] ~~Instanciate cozy client outside and pass it to `Remote`?~~


sebn:
Some parts of the code probably deserve more comments... (reviewing myself)

sebn:
Just woke up thinking that maybe the remote changes should be replicated and logged from pouch instead of introducing new cozy client API...

sebn:
Build is finally green.

Remaining error in logs (but doesn't fail the test): https://travis-ci.org/cozy-labs/cozy-desktop/jobs/194480793#L897

sebn:
:shipit: anybody?

sebn:
- [ ] Extract hardcoded folder ids
- [ ] Do not try to put trash in itself on cleanup
- [ ] Use a mango view to clean up files/dirs?

src/app.js
```diff
@@ -26,6 +24,8 @@ import Local from './local'
 import Remote from './remote'
 import Sync from './sync'
 
+// TODO: App.Permissions v3
+// eslint-disable-next-line no-unused-vars
```
enguerran: why?

src/app.js
```diff
@@ -153,15 +153,8 @@ class App {
...
-      register(cozyUrl, deviceName, password, Permissions, function (err, res) {
-        if (err) { callback(err) }
-        let config = {file: true};
-        ({ deviceName } = res);
-        ({ password } = res)
-        let setDesignDoc = filterSDK.setDesignDoc.bind(filterSDK)
-        setDesignDoc(cozyUrl, deviceName, password, config, err => callback(err, res))
-      })
+      // TODO: App.registerRemote() v3
+      callback(null, {deviceName, password})
```
enguerran: It is a good time¹ to use *asynchronism* with `promise` or `await/async`, isn't?

*¹: regarding the fact we use ES2015 and babel*

src/app.js
```diff
@@ -274,7 +267,7 @@ class App {
     this.merge = new Merge(this.pouch)
     this.prep = new Prep(this.merge, this.ignore)
     this.local = this.merge.local = new Local(this.config, this.prep, this.pouch, this.events)
-    this.remote = this.merge.remote = new Remote(this.config, this.prep, this.pouch, this.events)
+    this.remote = this.merge.remote = new Remote(this.config)
```
enguerran: Why `this.remote` needs to be equal to `this.merge.remote`?

src/remote/consts.js
```diff
@@ -0,0 +1,6 @@
+// https://github.com/cozy/cozy-stack/blob/master/pkg/consts/consts.go
```
enguerran: I may prefer a full filename like `constants.js`

src/remote/cozy.js
```diff
@@ -0,0 +1,25 @@
+import { Cozy as CozyClient } from 'cozy-client-js'
+import fetch from 'node-fetch'
```
enguerran: Is `node-fetch` still mandatory?

src/app.js
enguerran: more words: why do you disable eslint rules?

src/app.js
sebn: Because there is a design issue (cyclic dependency) in the existing v2 code: `Prep -> Merge -> Remote -> Prep`.

Since refactoring all those components was out of scope, I didn't change the assignment.

enguerran: okay

src/remote/consts.js
sebn: I named it after the existing go module. But I don't mind renaming it.

src/app.js
sebn: Definitely, but I want to introduce it progressively, in methods I can test. Implementing registration will be done in another user story, so right now I just want to disable it so it doesn't crash.

src/remote/cozy.js
sebn: As `cozy-client-js` makes heavy use of the `fetch` API, and the goal was to use it in `cozy-desktop`, it makes sense to use the same HTTP request API everywhere. It may also make it easier to extract some code to the lib when it makes sense.

src/app.js
sebn: The object was used in registration, which I temporarily disabled.
I decided to keep it in order not to forget to set up permissions correctly when bringing back registration. Hence the `TODO`, and the `eslint-*` directive so eslint does complain about unused object.

I agree putting `TODO` notices everywhere is not the best, but I first thought it could be an easy way to keep track of things to port to the v3 stack.

sebn:
For `.travis.yml`, I trusted [cozy-client-js](https://github.com/cozy/cozy-client-js/blob/master/.travis.yml) ;)

sebn:
Can't find a way to make [this](https://github.com/cozy-labs/cozy-desktop/blob/31e4d1dd010b559d066c7e969eb3c995888c0cfa/test/helpers/integration.js#L19-L37) work using [cozy.defineIndex()](https://github.com/cozy/cozy-client-js/tree/master/docs#cozydefineindexdoctype-fields) and [cozy.query()](https://github.com/cozy/cozy-client-js/tree/master/docs#cozyqueryindexreference-query).

sebn:
Finally found a way thanks to @nono

sebn:
:ship:

sebn:
arg broken test... (I was sure it was green)

sebn:
Should be ok now.

PR #492, opened 2017-01-24
build: Set up the flow type checker

sebn:
Add dev dependencies:
- flow-bin
- babel-plugin-transform-flow-strip-types
- babel-eslint
- eslint-plugin-flowtype

Configure everything:
- Set up babel to strip flow types
- Make flow ignore node_modules/jsverify (it fails)
- Set up standard to use the babel-eslint compiler (so it doesn't fail with syntax error)
- Set up flow type linting rules for standard (through eslint plugin)

Introduce npm scripts:
- flow: runs without server startup message, shows all errors
- standard: runs standard as usual
- lint: runs both

This will allow us to progressively introduce type annotations in new code
(and eventually in existing code at some point)

### DIFF
--- a/.flowconfig
+++ b/.flowconfig
@@ -1,0 +1,15 @@
+[ignore]
+<PROJECT_ROOT>/bin/.*
+<PROJECT_ROOT>/coverage/.*
+<PROJECT_ROOT>/gui/.*
+<PROJECT_ROOT>/lib/.*
+<PROJECT_ROOT>/test/.*
+<PROJECT_ROOT>/tmp/.*
+
+<PROJECT_ROOT>/node_modules/jsverify/.*
+
+[include]
+
+[libs]
+
+[options]
--- a/package.json
+++ b/package.json
@@ -22,7 +22,9 @@
   ],
   "scripts": {
     "fauxton": "babel-node ./src/bin/fauxton.js",
-    "lint": "./node_modules/.bin/standard './src/**/*.js' './test/**/*.js'",
+    "flow": "flow status --quiet --show-all-errors",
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -59,14 +67,18 @@
   "devDependencies": {
     "babel-cli": "^6.18.0",
     "babel-core": "^6.21.0",
+    "babel-eslint": "^7.1.1",
```
enguerran: Not sure `babel-eslint` is mandatory for `standardjs`, but it may be mandatory for `flow`. That is weird.

nono: The default parser for standardjs/eslint can't read the flow annotations, you have to switch to the babel parser for that.

.flowconfig
```diff
@@ -0,0 +1,15 @@
+[ignore]
+<PROJECT_ROOT>/bin/.*
+<PROJECT_ROOT>/coverage/.*
+<PROJECT_ROOT>/gui/.*
+<PROJECT_ROOT>/lib/.*
+<PROJECT_ROOT>/test/.*
+<PROJECT_ROOT>/tmp/.*
+
+<PROJECT_ROOT>/node_modules/jsverify/.*
```
nono: I'm surprised that flow doesn't ignore node_modules by default

sebn: According to the `flow ls` output, it doesn't seem to ignore anything... (event images and json files are listed)

Since I didn't know if flow needed to analyse node modules, and it didn't seem to be an issue (except maybe for performance, but it seems ok even with modules), I decided not to ignore them, and see whether it became an issue.

package.json
enguerran: okay, TIL

PR #497, opened 2017-02-06
bin/fauxton: Fix v2's couch helpers import

sebn:
It was broken while moving v2 test helpers to a subdir.


sebn:
I assumed spending time to set up something to prevent this issue to arise again was not worth it.

PR #501, opened 2017-02-06
V3/build/dev setup improvements

sebn:

PR #496, opened 2017-02-02
V3/feature/pull metadata

sebn:
To be merged after https://github.com/cozy-labs/cozy-desktop/pull/503

This completes the pull feature with cozy-stack v3, except pulled files will be empty (fetching their content is another story).

**Known issues**

-  Node's `Buffer` class is permissive regarding base64 padding. Which means we could accidentally store a non-padded digest string, then fail to compare it to the same padded one.

**Not included**

- `Remote` unit test (it's only glue code right now, I'll unit-test it as soon as I start restoring v2's remaining `Watcher` code)
- Extract crypto-related constants (`HASH_FUNCTION = 'md5'`? `DIGEST_ENCODING = 'base64'`? `DIGEST_BYTES_SIZE = 16`?)

PR #502, opened 2017-02-06
V3/refactor/conversion and watcher modules

sebn:
~~To be merged after https://github.com/cozy-labs/cozy-desktop/pull/501~~ (already merged)

This is the first step to port v2 code to the new cozy stack v3

I decided to keep as much existing code as possible, so we don't loose existing behavior.
Ideally, I'd like to bring back the old pull integration tests as soon as files are not empty anymore.
So we could refactor more things safely.

**Not included**

- `conversion.createLocalDoc()` unit test: I extracted it from v2's `RemoteWatcher`, where it was only tested in integration with other extracted methods.
- Rename `RemoteDoc` / `LocalDoc` to `RemoteMetadata` / `LocalMetadata`? (would be better since *Metadata* is a domain thing, while using a CouchDB *document* is an implementation detail... but I didn't rename it yet because *doc* variables are used almost everywhere in the code)
- Rename `onChange/putDoc/removeRemote` methods, or even define metadata changes as a single pure function (I could still do it as soon as I really understand how prep/merge work)
- Promisify pouchHelpers
- Rename local fields to match the remote ones (it may break lots of existing v2 code, better make it work first)
- Store the remote `dir_id` (it seems to work with the `path` right now)


src/pull.js
```diff
@@ -0,0 +1,94 @@
...
+
+import type { RemoteDoc } from './remote/document'
+import type { LocalDoc } from './local/document'
+
+const log = printit({
+  prefix: 'Pull',
+  date: true
+})
+
+// Pull changed files and directories metadata from the remote Cozy, and merge
```
nono: s/changed/get changes for/ ?

src/pull.js
```diff
@@ -0,0 +1,94 @@
...
+// them into the local database.
+export default class Pull {
```
nono: I'm not sure it's a good idea to call this `Pull`. There is a pull command on the command-line that does a bit more:

- it track changes on the remote cozy and update the pouchdb with these changes (like your class)
- it also applies these changes to the file-system

src/remote/document.js
```diff
@@ -0,0 +1,21 @@
...
+  created_at: string,
+  dir_id: string,
+  executable?: boolean,
+  md5sum?: string,
+  mime?: string,
+  name: string,
+  path?: string,
+  size?: string,
+  tags: string[],
+  type: string,
```
nono: You have both `type` and `_type`?

src/conversion.js
```diff
@@ -0,0 +1,45 @@
...
+    case FILE_TYPE: return 'file'
+    case DIR_TYPE: return 'folder'
+    default: throw new Error(`Unexpected Cozy Files type: ${remote}`)
+  }
+}
+
+// Transform a remote document in a local one
+//
+// We are tolerant with the input. For example, we accept files with no
+// checksum (e.g. from konnectors).
```
sebn: Ok, thanks, I wasn't aware of this.

src/local/document.js
```diff
@@ -0,0 +1,26 @@
+/* @flow */
+
+// The local Cozy Desktop metadata, as stored in PouchDB
```
sebn: I thought of pouch as being part of "local", since it's mostly used to implement local metadata storage.
But I don't have a clear view of the reconciliation part and the limit between it and "local".
So ok to rename `local/document` to `document` if it makes more sense.

src/local/document.js
```diff
@@ -0,0 +1,26 @@
...
+export type LocalDoc = {
+  _id: string,
+  // TODO: v3: Rename to md5sum to match remote
+  checksum?: ?string,
```
sebn: Should probably have been `checksum?: string`, so we have either a key with a non-null value, or no key at all.

But if checksums became mandatory, maybe we should just remove the mobile-specific test and corresponding code, and enforce it in the flow declaration.

nono: The checksum is mandatory for files, not folders. And you can remove the mobile-specific test (the stack now generates a md5sum for files when no md5sum was given at upload, so all files in couchdb have a md5sum).

src/pouch.js
```diff
@@ -123,6 +123,16 @@ class Pouch {
     })
   }
 
+  byRemoteIdMaybe (id, callback) {
+    this.byRemoteId(id, (err, was) => {
+      if (err && (err.status !== 404)) {
```
sebn: I extracted this from [v2's RemoteWatcher](https://github.com/cozy-labs/cozy-desktop/blob/master/src/remote/v2/watcher.js#L171-L173).

I assumed this was some PouchDB behavior, but it's actually done in `byRemoteId` (line 119, just before this one).
Maybe I could have used `this.db.query` directly.

src/pull.js
sebn: I'm not sure either.
Right now it only pulls the metadata, delegating part of the work to the `Prep` class.

I was first thinking about making some "pull" domain notion visible by extracting the corresponding logic out of `RemoteWatcher`, including the filesystem update.
Then I figured out that it was actually only updating metadata in pouch, and that including fs update in the same class may actually not be such a good idea.

I assumed I could wait to understand the fs part better, and eventually rename it to something else.

src/remote/document.js
sebn: `_type` is the doctype, while type is the "files" API attribute.

I was a bit confused about `cozy-client-js`'s `find()` method returning this, and the cozy-stack using a nested `attributes` object...

nono: Oh, I forgot about the JSON-API stuff!

src/pull.js
sebn: `MetadataPull`?
I'd like to keep the `Pull` suffix since at least it indicates the direction...
Or were you think of moving this back into the `RemoteWatcher`? (I thought it was to much responsibility for a single class...)

nono: I'd have put this code in RemoteWatcher. But MetadataPull is OK too for me

sebn:
- [x] Stop saying we are tolerant with checksum in comments
- [x] Rename `LocalDoc` to `Metadata`
- [x] Rename `createLocalDoc` to `createMetadata`
- [ ] `checksum: string`
- [ ] Remove mobile-specific test

test/unit/remote/watcher.js
```diff
@@ -0,0 +1,418 @@
...
+
+      await this.pull.onChange(doc)
+
+      this.prep.deleteDocAsync.called.should.be.true()
+      let id = this.prep.deleteDocAsync.args[0][1].path
+      id.should.equal('my-folder/file-1')
+    })
+
+    it('calls addDoc for folder created by the mobile app', async function () {
+      // FIXME: What makes a folder created by the mobile app special?
```
sebn: I think this is the mobile test I was thinking, it's not related to the checksum actually.
Still not sure what makes it special... the folder name?
@nono ?

sebn:
Removed the `FIXME`. Ready to merge as soon as CI is green.

test/unit/remote/watcher.js
nono: @sebn the doctype, it's 'Folder' and not 'folder' (all lowercase)

nono: But it won't happen in cozy-v3, so the test can be removed safely

sebn: Sorry, I missed your comment. I'll fix remove it in https://github.com/cozy-labs/cozy-desktop/pull/496

PR #504, opened 2017-02-07
Update dependencies

sebn:
I did this by hand since Greenkeeper doesn't play well with yarn.

- cozy-client-js was upgraded to 0.0.7, not 0.0.8
- babel-plugin-autobind-class-methods was upgraded to 3.0.0, not 5.0.1

See commit messages for explanations.


nono:
@sebn fyi, there is a known workaround for greenkeeper + yarn: https://gist.github.com/maxrimue/16e272c4e0b8a8ac9dd5570e180e08c0

PR #507, opened 2017-02-07
V3/feature/files download

sebn:


test/unit/remote/index.js
```diff
@@ -6,75 +6,68 @@ import sinon from 'sinon'
...
-          localDoc.should.equal(doc)
-          let stream = fs.createReadStream(fixture)
-          return callback(null, stream)
-        }
+
+      let binary
+      try {
+        binary = await builders.file().named('pillow.jpg')
+          .contentType('image/jpeg').dataFromFile(fixture).build()
+        // FIXME: binary.md5sum.should.equal(expectedChecksum)
```
sebn: Can't find a way to make this work yet.

test/builders/remote/file.js
```diff
@@ -11,17 +13,32 @@ export default class RemoteFileBuilder extends RemoteBaseBuilder {
...
+    return this
+  }
+
+  data (data) {
+    this._data = data
+    return this
+  }
+
+  dataFromFile (path) {
+    this._data = fs.readFileSync(path)
```
sebn: `fs.readFileSync()` [returns a `Buffer`](https://nodejs.org/docs/latest-v4.x/api/fs.html#fs_fs_readfilesync_file_options) since no encoding argument is given.

Node's `Buffer` class [implements the Uint8Array API](https://nodejs.org/docs/latest-v4.x/api/buffer.html#buffer_buffer).

```
> let b = fs.readFileSync('./test/fixtures/cool-pillow.jpg')
> b
<Buffer ...>
> b instanceof Buffer
true
> b instanceof Uint8Array
true
> b.buffer
ArrayBuffer { byteLength: 21102 }
> b.buffer instanceof ArrayBuffer
true
```

But upload fails as if the `ArrayBuffer` was converted to `String`:

```
> b.buffer.toString()
'[object ArrayBuffer]'
```

```
$ cat storage/test.cozy-desktop.local\:8080/pillow.jpg 
[object ArrayBuffer]
```

sebn: Reminder for myself:
- Make sure `cozy-client-js` effectively picks up `node-fetch` as its `fetch` implementation
- I believed `node-fetch` supported buffers for the body... But `cozy-client-js` detects and uses the `.buffer` property. So maybe take a look at the typechecks in `node-fetch`?
- `node-fetch` is supposed to play well with streams, but `cozy-client-js` doesn't seem to, even on the server... PR?

sebn: `node-fetch` actually works with a jpeg buffer... Looking at how it handles it.

sebn:
The build failure on node 7 is a bit strange. I wonder whether the latest node 7 Buffer implementation tries to instanciate some `ArrayBuffer`, which for some unknown reason was not reset.

sebn:
:ship:

PR #512, opened 2017-02-09
V3/feature/new directory push

sebn:
- [x] Bring back log in `src/remote/v2/index.js`
- [x] Add missing type annotations
- [x] Either use new `addFolder`'s promise, or revert back to callback API
- [ ] ~~Overwrite timestamps => in another PR~~

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -1,0 +1,6 @@
+/* @flow */
+
+// Node-style callback
+// FIXME: Could not find a way to define a Callback<T> type with multiple
+//        signatures.
+export type Callback = (?Error, any) => void;
--- a/src/app.js
+++ b/src/app.js
@@ -274,9 +274,6 @@ class App {
 
   // Start the synchronization
   startSync (mode, callback) {
-    // FIXME: v3: Temporarily force pull mode only, until push works
-    mode = 'pull'
-
     this.config.setMode(mode)
     log.info('Run first synchronisation...')
     this.sync.start(mode, err => {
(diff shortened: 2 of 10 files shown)


sebn:
@nono the timestamps aren't fixed yet, but an early review could help me fixing issues faster.

src/utils.js
```diff
@@ -0,0 +1,6 @@
+/* @flow */
+
+// Node-style callback
+// FIXME: Could not find a way to define a Callback<T> type with multiple
+//        signatures.
+export type Callback = (?Error, any) => void;
```
nono: I don't know flow, but did you try this?

```ts
export type Callback = (?Error, any, any, any) => void;
```

sebn: Why 4 args?

I was thinking about something more like this:

```javascript
export type Callback<T> = ((err: Error) => void) & ((err: null, result:T) => void);
```

sebn: But when I use it, flow complains about `result` being nullable, even in the `else` branch of an `if (err)` check.

nono: To support up to 3 arguments (+ error)

sebn: Oh, ok, makes sense.
I would prefer to wait until I encounter the case, so I can make sure the type modification actually works on existing code.

sebn:
Build is green and the feature could already be used (even with the timestamp limitation).
I think we should merge and fix the timestamps in another PR.

@nono are you already working on the date header support in cozy-client-js, or may I take a look at it?

nono:
@sebn you can take it

PR #518, opened 2017-02-14
Update yarn deps

nono:


test/unit/remote/cozy.js
```diff
@@ -21,6 +21,13 @@ describe('RemoteCozy', function () {
 
   beforeEach(function () {
     remoteCozy = new RemoteCozy(COZY_URL)
+    // FIXME: Temporary hack to make cozy-client-js think it has OAuth tokens
+    remoteCozy._authstate = 3
+    remoteCozy._authcreds = Promise.resolve({
```
sebn: I'm not sure to understand this one... Since `remoteCozy` is a `RemoteCozy` instance (and not a `cozy-client-js.Cozy` one), why do we need to do this?

sebn: Let's keep it, we'll remove it with other temporary hacks anyway.

PR #520, opened 2017-02-14
V3/refactor/file upload from stream

sebn:


sebn:
Build broken because of new cozy-stack permissions.

PR #521, opened 2017-02-14
Metadata helpers

sebn:
Copied from v2 code to the metadata module.


sebn:
Maybe I could have added a random `_rev` in unit test instead of removing type annotation. But not so important, we should probably replace fixture objects like this with test data builders anyway. And specify only the relevant fields.

PR #522, opened 2017-02-14
Remote class properties

sebn:
Needed to reimplement `Remote#addFile()` with Flow typechecks.


sebn:
The `FileReader` name is not so great... `FileStreamProvider`? Any other idea?

nono:
`FileStreamProvider` looks fine to me

sebn:
lint & tests pass locally. :ship:

PR #525, opened 2017-02-16
Add new local file on remote Cozy

sebn:


src/timestamp.js
```diff
@@ -0,0 +1,62 @@
+/* @flow */
+
+export type Timestamp = Date
+
+export function InvalidTimestampError (obj: any) {
+  this.name = 'InvalidTimestampError'
+
+  this.message = `
+    Timestamps must be JavaScript UTC Date instances with second-only precision,
```
sebn: Reviewing myself: a Date doesn't have to *be* UTC, it doesn't really make sense.

src/timestamp.js
```diff
@@ -0,0 +1,62 @@
...
+  ensureValid(t1)
+  ensureValid(t2)
+
+  return t1.getTime() === t2.getTime()
+}
+
+function stringify (t: Timestamp) {
+  ensureValid(t)
+
+  return t.toISOString().substring(0, 19) + 'Z'
```
sebn: This will break in year 10000...

sebn:
Except for comments above, tests pass locally, same for manual acceptance testing. WDYT?

sebn:
Unhandled cases:
- File is deleted locally just before being uploaded
- File was already created remotely while offline
- Dir `foo/` and file `foo/bar.baz` created locally, but for whatever reason `foo/bar.baz` is uploaded before `foo/` (so folder doesn't exist remotely)

sebn:
I expected [`bluebird.asCallback`](http://bluebirdjs.com/docs/api/ascallback.html) to be of some help for glueing callbacks/promises, but I couldn't make it work. Maybe because `cozy-client-js` calls don't return Bluebird promises, I wonder...

nono:
> 1. File is deleted locally just before being uploaded
> 2. File was already created remotely while offline

In fact, I think those two cases are handled. We should test them later, but in theory, the upload will fail, the watchers will detect the changes, and the situation will come back to the normal.

> 3. Dir foo/ and file foo/bar.baz created locally, but for whatever reason foo/bar.baz is uploaded before foo/ (so folder doesn't exist remotely)

I think we always create the folder before putting files in them. There is only a race condition if the folder is created in local, then synchronized to the remote, and then at the same time the folder is deleted on the remote and file added on the local. But it is conflict management, not something to consider in this PR.

PR #526, opened 2017-02-16
Fix build with cozy permissions

nono:


test/helpers/password.js
```diff
@@ -1,7 +1,7 @@
 'use strict'
 
 // The password used for integration tests:
-let password = 'CozyTest$1'
+let password = process.env.COZY_PASSPHRASE || 'CozyTest_1'
```
sebn: The goal of this "executable" module was for the default test password/passphrase to be defined in a single place, but now it is defined thrice.

src/app.js
```diff
@@ -152,14 +152,14 @@ class App {
       return
     }
     if (deviceName == null) { deviceName = os.hostname() || 'desktop' }
-    this.askPassword(function (_, password) {
+    this.askPassword(function (_, passphrase) {
```
sebn: `s/askPassword/askPassphrase`? (can be done later)

test/helpers/password.js
sebn: Or do you think of some scenario where still being able to overwrite it is more important?

sebn:
Merging as soon as CI is green (unless you want me to wait).

test/helpers/password.js
nono: Don't think too much about this. I think this module will disappear. We will use a token generated by the stack, not the passphrase, in tests.

sebn:
Strange issue: when using `yarn link cozy-client-js`, [this line](https://github.com/nono/cozy-desktop/blob/f8655c96d371f20077d3585c0392f07792a5666a/test/helpers/cozy.js#L22) fails because `CozyClient` is undefined. As soon as I unlink and install version 0.1.0 it works.

Since we don't use `yarn link` on Travis it should work.

PR #530, opened 2017-02-17
CLI registration

nono:


src/bin/cli.js
```diff
@@ -79,18 +78,17 @@ program
 program
   .command('sync')
   .description('Synchronize the local filesystem and the remote cozy')
-  .option('-k, --insecure', 'Turn off HTTPS certificate verification.')
```
sebn: Is this not something self-hosted users will want?

src/bin/repl.js
```diff
@@ -12,7 +12,7 @@ The following objects are available:
   app     The cozy-desktop app
   config  Your active cozy-desktop configuration`)
 
-if ((config.deviceName != null) && (config.url != null) && (config.path != null)) {
+if (config.url) {
```
sebn: `app.config.isValid()`?

src/config.js
```diff
@@ -18,135 +18,103 @@ class Config {
...
-        case (old == null):
-          throw new Error('Incompatible mode')
-        default:
-          this.devices[deviceName].mode = mode
-          return this.save()
-      }
-    } else {
-      return false
+  saveMode (mode) {
+    let old = this.config.mode
```
sebn: `const`

src/config.js
```diff
@@ -18,135 +18,103 @@ class Config {
...
+    if (old === mode) {
+      return true
+    } else if (old) {
+      throw new Error('Incompatible mode')
```
sebn:
```javascript
throw new Error(`Once you set mode to "${old}", you cannot switch to "${mode}"`)
```

src/config.js
```diff
@@ -18,135 +18,103 @@ class Config {
...
-      return this.save()
-    } else {
-      return false
-    }
+  ///////////////////////////////////////////////////////////
+  // Implement the Storage interface for cozy-client-js oauth
+
+  save (key, value) {
+    this.config.credentials[key] = value
+    return Promise.resolve(value)
```
sebn: Have some kind `Config#oauthStorage()` return the expected interface? (not sure how you expect to use it)

src/config.js
```diff
@@ -18,135 +18,103 @@ class Config {
...
 
-function __guard__ (value, transform) {
-  return (typeof value !== 'undefined' && value !== null) ? transform(value) : undefined
+  clear () {
+    this.config.credentials = Object.create(null)
+    return Promise.resolve()
+  }
 }
+
+export default Config
```
sebn: `export default class Config` directly?

src/bin/cli.js
nono: Probably, but it is a priority for the short term, and the code we had for that will no longer work

test/unit/config.js
```diff
@@ -11,111 +11,52 @@ describe('Config', function () {
...
-  describe('saveConfig', () =>
+  describe('persist', function () {
     it('saves last changes made on the config', function () {
-      this.config.devices['new-cozy2'] = {
-        deviceName: 'new-cozy2',
-        passphrase: 'passphrase',
-        url: 'none'
-      }
-      this.config.save()
+      this.config.setCozyUrl('http://cozy.local:8080/')
```
sebn:
```javascript
const url = 'http://cozy.local:8080/'
```

test/unit/config.js
```diff
@@ -11,111 +11,52 @@ describe('Config', function () {
...
-    it('returns devicename from args', function () {
-      process.argv = ['arg1', '-d', 'test']
-      let name = this.config.getDefaultDeviceName()
-      name.should.equal('test')
-    })
-    it('returns default devicename when no args', function () {
-      process.argv = []
-      let name = this.config.getDefaultDeviceName()
-      name.should.equal('tester')
+      conf.getCozyUrl().should.equal('http://cozy.local:8080/')
```
sebn: I started using the following pattern in new code:

```javascript
should(conf.getCozyUrl()).equal(...)
```

Pros:
- no monkey-patching involved...
- ... which makes it more Flow-friendly

Please let me now in case you disagree, so I stop using it everywhere.

src/bin/cli.js
```diff
@@ -31,37 +31,36 @@ let sync = function (mode, args) {
...
-        })
-      } else {
-        return log.log(`${what} ${filename} (unknown size)`)
-      }
-    })
-    return app.synchronize(mode, function (err) {
-      if (err) { return process.exit(1) }
-    })
-  } else {
+  if (!app.config.hasClient()) {
```
sebn: `app.config.isValid()`

src/bin/repl.js
```diff
@@ -3,7 +3,7 @@ import { start } from 'repl'
 import App from '../app'
 
 const app = new App(process.env.COZY_DESKTOP_DIR)
-const config = app.config.getDevice()
+const config = app.config.config
```
sebn: Do you plan to remove the multi-config support? (just out of curiosity)

nono: Well, there were no multi-config support. I had tried it, but it was too error-prone and even if the config file can have several "devices", only the first one was really used.

src/config.js
nono: The goal is to persist what cozy-client-js puts in this storage in the config file (it is the client informations and the OAuth tokens).

sebn: (and move `save` / `load` / `...` to the new object)

sebn: Was this the reason for the `persist()` method not to be named `save()`?

nono: Yes :-p

src/config.js
```diff
@@ -18,135 +18,103 @@ class Config {
...
-      return this.devices[deviceName]
-    } else if (Object.keys(this.devices).length === 0) {
-      return {} // No device configured
-    } else {
-      log.error(`Device not set locally: ${deviceName}`)
-      throw new Error(`Device not set locally: ${deviceName}`)
-    }
+  // Get the path on the local file system of the synchronized folder
+  getSyncPath () {
+    return this.config.path
```
sebn: I'm not sure whether es getters/setters are supported... Would it make sense to use them?

sebn: I mean:

```javascript
get syncPath() {
   ...
}

set syncPath(path) {
  ...
}
```

package.json
```diff
@@ -2,7 +2,7 @@
   "name": "cozy-desktop",
   "version": "0.16.0",
   "description": "File Synchronization Client for Cozy Cloud",
-  "homepage": "https://github.com/cozy-labs/cozy-desktop",
+  "homepage": "https://docs.cozy.io/en/mobile/desktop.html",
```
sebn: Shouldn't we just disable the cozy-desktop github pages if they are not supposed to be the main entry point for users?

nono: We try to keep documentation for users on cozy.io, and on github for people who want to contribute to our projects

src/remote/registration.js
```diff
@@ -0,0 +1,72 @@
...
+      cozyURL: this.url,
+      oauth: {
+        storage: this.config,
+        clientParams: params,
+        onRegistered: this.onRegistered
+      }
+    })
+    return cozy.authorize().then(
+      (res) => res,
+      (err) => { throw err }
```
sebn: Why not just `.catch(err => { throw err })` ?
Also, since `process()` is not async and it returns the promise, is the `catch` block even needed?

test/unit/remote/cozy.js
```diff
@@ -14,25 +15,37 @@ const cozyStackDouble = new CozyStackDouble()
 describe('RemoteCozy', function () {
   before(() => cozyStackDouble.start())
   beforeEach(deleteAll)
+  before('instanciate config', configHelpers.createConfig)
+  before('register OAuth client', configHelpers.registerClient)
+  after('clean config directory', configHelpers.cleanConfig)
```
sebn: Won't mocha use the function name as a description anyway? (assuming it is named)

sebn:
Looks good to me overall.

The only thing really bothering me is the `Config` class exposing oauth-related `save` / `load` / `delete` methods:
- I assume no cozy-desktop code will ever use them
- `save()` / `load()` can be used to change / read whatever property of the class
- they are not of the same abstraction level (hence the name conflict with `persist()`)

src/config.js
nono: Good idea!

nono:
For the config class that also exports the storage interface, I'm OK with you that it should be refactored. But I'd prefer to keep that for later. I'd like to explore token refresh and device revocation before.

src/remote/registration.js
nono: You are right. I've added them to console.log this step. Then I removed the console.log, but the `then()` can also be removed.

sebn:
> But I'd prefer to keep that for later

👌

PR #536, opened 2017-02-21
Remote overwrite file

sebn:
Works with https://github.com/cozy/cozy-client-js/pull/78 and https://github.com/cozy/cozy-stack/pull/305

### DIFF
--- a/src/remote/index.js
+++ b/src/remote/index.js
@@ -5,6 +5,7 @@ import printit from 'printit'
 import Config from '../config'
 import * as conversion from '../conversion'
 import RemoteCozy from './cozy'
+import { jsonApiToRemoteDoc } from './document'
 import Pouch from '../pouch'
 import Prep from '../prep'
 import Watcher from './watcher'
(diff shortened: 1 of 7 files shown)


sebn:
I'll fix the flow issues.
Tests fail because we need changes from master branch of cozy-client-js.

src/remote/index.js
```diff
@@ -99,9 +102,33 @@ export default class Remote {
...
+
+    doc.remote._rev = updated._rev
+
+    return jsonApiToRemoteDoc(updated)
+  }
+
+  overwriteFile (doc: Metadata, old: Metadata, callback: Callback) {
+    this.overwriteFileAsync(doc, old)
+      .then(updated => callback(null, updated))
+      .catch(callback)
```
nono: We should probably add a try/catch here:

```js
overwriteFile () {
  try {
    this.overwriteFileAsync(doc, old)
      .then(updated => callback(null, updated))
      .catch(callback)
  } catch (err) {
    callback(err)
  }
```

nono: In `overwriteFileAsync`, if the promise returned by `this.remoteCozy.updateFileById` if rejected, the await will throw an error, right?

nono: The mix of callbacks/promises/async-await is not ideal. I hope we can remove most callbacks soon.

sebn: > We should probably add a try/catch here

Isn't this already handled by the catch call on line 121?

> the await will throw an error, right?

Well, reading [this post](https://pouchdb.com/2015/05/18/we-have-a-problem-with-promises.html) I believed it wont, but according to both my unit and my manual tests, errors seems to correctly bubble up.

> The mix of callbacks/promises/async-await is not ideal.

Agreed. Then what about this?

```javascript
async overwriteFile (...) {
  try {
    const updated = await this.overwriteFileAsync(doc, old)
    callback(null, updated)
  } catch (err) {
    callback(err)
  }
}
```

sebn:
> IIRC, you can put a git repository in the dependencies in package.json (to test that travis is green with the master branch of cozy-client-js)

Do you think we should commit the git dependency on master for some time so you don't have to publish a new version of `cozy-client-js` everyday?

We could eventually get back to an existing version later.

nono:
Yes, we can use use a git dependency on master for some time. We just have to be sure to use a released version when we will publish cozy-desktop

src/remote/index.js
nono: LGTM

nono: > Isn't this already handled by the catch call on line 121?

I don't think so. `aPromise.catch` only catch a rejected state for the promise, but not the standard thrown exception. And I think await throw such an exception when its promise is rejected.

sebn: Makes sense.

sebn:
Oops, don't we still need to commit the new build in cozy-client-js? Otherwise we will use an old version...

sebn:
- [x] Rewrite `overwriteFile()` using `async`/`await` style

sebn:
- [x] Use cozy-client-js master branch with latest build

PR #541, opened 2017-02-23
Fix path conversion

sebn:


sebn:
At some point, I'll provide a way to "build" the remote doc data structure without creating the corresponding file/dir on the remote cozy, and use it everywhere in tests instead of duplicated fixtures with non-relevant fields.

sebn:
Last 2 commits should fix the build.

PR #544, opened 2017-02-24
Persist the refreshed access token

nono:


src/config.js
```diff
@@ -93,6 +93,10 @@ export default class Config {
 
   save (key, value) {
     this.config[key] = value
+    if (key == 'creds') {
+      // Persist the access token after it has been refreshed
+      this.persist()
+    }
```
sebn: I'm not very familiar with the auth process yet. But from what I understand, a key `'foo'` could be added to `config` but not persisted, then `'creds'` would be added and both of them persisted...
Then why don't we just persist on every `save()` call?

(Also eslint doesn't like the `==` comparison)

nono: >  from what I understand, a key 'foo' could be added to config but not persisted, then 'creds' would be added and both of them persisted...

It is the case.

> Then why don't we just persist on every save() call?

We don't want to persist transitional states from the oauth registration flow. If the user starts the registration for one cozy instance, and then stops it before saying OK, and then starts for the registration for another cozy instance, we don't to have persisted a config file with irrelevant data that will make the next registration fails if we don't clean them.

sebn: Then maybe update the comment? Something like:

```javascript
// Persist the oauth config once the access token has been refreshed (i.e. setup complete)
```

PR #547, opened 2017-02-27
OAuth registration in GUI

nono:


gui/app/main.js
```diff
@@ -70,6 +69,7 @@ const sendToMainWindow = (...args) => {
 }
 
 const sendErrorToMainWindow = (msg) => {
+  // FIXME
```
sebn: Why?

gui/app/main.js
```diff
@@ -464,17 +446,16 @@ ipcMain.on('choose-folder', (event) => {
 })
 
 ipcMain.on('start-sync', (event, arg) => {
-  if (!device) {
+  if (!desktop.config.client) {
     console.error('No device!')
```
sebn: `'No client!'` ?

sebn: Would `if (!desktop.config.isValid())` make sense?

gui/app/main.js
```diff
@@ -414,25 +423,24 @@ app.on('ready', () => {
...
-})
-
 ipcMain.on('register-remote', (event, arg) => {
-  // TODO use package.json for pkg ({} below)
-  desktop.config.cozyUrl = arg.url
-  desktop.registerRemote(arg.url, {})
+  desktop.config.cozyUrl = arg.cozyUrl
+  const onRegistered = (client, url) => {
+    let resolveP
+    const promise = new Promise((resolve) => { resolveP = resolve })
```
sebn: Oh my :)
I can't think of a better solution yet.

gui/app/main.js
nono: The error messages won't be the same with cozy-client-js

gui/app/main.js
```diff
@@ -454,8 +454,8 @@ ipcMain.on('choose-folder', (event) => {
 })
 
 ipcMain.on('start-sync', (event, arg) => {
-  if (!desktop.config.client) {
-    console.error('No device!')
+  if (!desktop.config.isValid()) {
+    console.error('No client!')
```
sebn: `'No valid config'` then (sorry)

sebn:
👍

PR #549, opened 2017-02-27
Unregister the client

nono:


src/app.js
```diff
@@ -181,22 +181,13 @@ class App {
...
-    //   } else {
-    //     log.info('Current device properly removed from remote cozy.')
-    //     fs.remove(this.basePath, callback)
-    //   }
-    // })
+  removeRemote () {
+    if (!this.remote) {
+      this.instanciate()
+    }
+    this.remote.unregister()
```
sebn: Catch failures?

src/remote/cozy.js
```diff
@@ -42,6 +42,7 @@ export default class RemoteCozy {
     })
 
     // Aliases:
+    this.unregister = this.client.auth.unregisterClient
```
sebn:
```javascript
unregister: () => Promise<*>
```

Should fix flow.

src/app.js
```diff
@@ -151,22 +143,18 @@ class App {
...
+      if (!this.remote) {
+        this.instanciate()
+      }
+      this.remote.unregister()
+      fs.removeSync(this.basePath)
+      log.info('Current device properly removed from remote cozy.')
+    } catch (err) {
+      log.error('An error occured while unregistering your device.')
+      log.error(err)
+    }
```
sebn: Won't this miss failures since `unregister()` is async?

nono: You are right, it's better to make `await` here

src/app.js
```diff
@@ -143,17 +143,19 @@ class App {
 
   // Unregister current device from remote Cozy and then remove remote from
   // the config file
-  removeRemote () {
+  async removeRemote () {
     try {
       if (!this.remote) {
         this.instanciate()
       }
       this.remote.unregister()
```
sebn: `await`?

nono: Yeah, my mistake

PR #550, opened 2017-02-28
Remote#moveFile()

sebn:


sebn:
Last commit (https://github.com/cozy-labs/cozy-desktop/pull/550/commits/453191d63ba4db2f7dc9b863893a1f013d07e76b) seems useless to me, like a mix between testing the language and testing other functions than the current one.

@nono wdyt?

sebn:
I suppose those are conflicts, which we'll need to resolve at some point. Separate story again?

src/utils.js
```diff
@@ -4,3 +4,11 @@
 // FIXME: Could not find a way to define a Callback<T> type with multiple
 //        signatures.
 export type Callback = (?Error, any) => void;
+
+export function composeAsync (f1: (...args: Array<*>) => Promise<*>,
+                       f2: (*) => Promise<*>):
+                      (...args: Array<*>) => Promise<*> {
+  return async function composed (...args: Array<*>): Promise<*> {
+    return f2(await f1(...args))
```
nono: Shouldn't we wrap this code in a `try ... catch`?

sebn:
Also I couldn't find a straightforward way to "unpromisify" a method with Bluebird.

I'm still confused about whether using Bluebird's `Promise#asCallback()` would be sufficient, or should we keep `try` / `catch` blocks?
And then, why isn't there any in the `unpromiser()` function of `cozy-client-js`?

src/remote/index.js
```diff
@@ -165,17 +165,34 @@ export default class Remote implements Side {
...
+      name: newName,
+      dir_id: newDir._id
+    })
+
+    return conversion.createMetadata(newRemoteDoc)
+  }
+
+  moveFile (doc: Metadata, from: Metadata, callback: Callback): void {
+    try {
+      this.moveFileAsync(doc, from).asCallback(callback)
```
nono: What is this `asCallback` method?

src/remote/index.js
```diff
@@ -165,17 +165,34 @@ export default class Remote implements Side {
...
+    } catch (err) {
```
nono: I'm not convinced that the `try/catch` here is useful (we are not in an async function)

src/remote/index.js
sebn: http://bluebirdjs.com/docs/api/ascallback.html

src/remote/index.js
sebn: Makes sense

src/utils.js
sebn: Well, I tried to write another unit test to feed `composeAsync` with some throwing sync function, but it still worked... Hence my confusion :)

nono:
> I'm still confused about whether using Bluebird's `Promise#asCallback()` would be sufficient, or should we keep try / catch blocks?

I think that try / catch are not useful with Bluebird's `Promise#asCallback()`

> And then, why isn't there any in the `unpromiser()` function of cozy-client-js?

Cozy-client-js don't use async/await, so `try / catch` are not useful.

src/utils.js
nono: I'm confused too.

@aenario, @jinroh : maybe one of you can help us understand what happen if `f1` returns a promise that is rejected

sebn: My bad, sorry for the noise.

I "awaited" the assertion promise:
```javascript
await should(composed(...)).be.rejected()
```

Instead of awaiting the result correctly:

```javascript
should(await composed(...)).be.rejected()
```

sebn: Well, actually, it doesn't reject, but it does throw...
So even if the test was broken, doesn't the code still works?
I mean, some unexpected exception is thrown, be it a bug or a custom exception type.
And it bubbles up to the unit test, instead of having the function rejecting the promise...
Is it really an issue?
(I may be missing something)

nono: It makes it hard to use the composed function. If you want to check if there is an error, you have to both check the returned promise and wrap the call with a try/catch. Having only one way to return an error from a function is way nicer.

sebn:
- [x] Remove useless `try` / `catch` block around `asCallback()` call
- [x] Remove useless commit (more tests will be added when handling missing file/directory failures)
- [ ] ~~Call `addFile()` when no `.remote` property~~ => Not sure how to reproduce it and whether it's part of the current story

:ship: for me (as soon as CI is green)

src/utils.js
sebn: The more I think about it, the more I believe either sync exceptions are bugs, or they should be handled in the component function who choosed to mix sync and async calls possibly throwing domain exceptions, not in the code calling the composed function.

Would you mind waiting to actually get the issue before handling this case?

nono: OK, let's try

nono:
> Call addFile() when no .remote property => Not sure how to reproduce it and whether it's part of the current story

Maybe, it's possible to get this case with these steps (to be tested):

1. Add a big file
2. While the big file is uploading, add another file and, 5 seconds later, rename it
3. Wait that the second file is handled

PR #554, opened 2017-03-01
Revoked client

nono:


src/app.js
```diff
@@ -207,44 +207,36 @@ class App {
...
-    if (this.config.isValid()) {
-      this.instanciate()
-      this.startSync(mode, callback)
-    } else {
+  synchronize (mode) {
+    if (!this.config.isValid()) {
       log.error('No configuration found, please run add-remote-cozy' +
                 'command before running a synchronization.')
-      __guardFunc__(callback, f => f(new Error('No config')))
+      throw new Error('No client configured')
```
sebn: Maybe extract `isValid()` call + log/error throwing to some `Config#ensureValid()` method? wdyt?

src/remote/watcher.js
```diff
@@ -56,6 +71,9 @@ export default class RemoteWatcher {
       await this.pouch.setRemoteSeqAsync(changes.last_seq)
       log.lineBreak()
     } catch (err) {
+      if (err.message === 'Client has been revoked') {
```
sebn: I imagine this string being hardcoded in multiple `if` statements...

Either extract it to some constant, or check the error type? (although I'm not even sure there is a reliable way to do so in js...)

src/remote/watcher.js
```diff
@@ -34,15 +34,30 @@ export default class RemoteWatcher {
...
 
   start () {
-    this.watch()
-    this.intervalID = setInterval(this.watch, HEARTBEAT)
+    this.started = new Promise((resolve, reject) => {
+      this.started.resolve = resolve
+      const watch () => {
+        this.watch().catch(err => {
+          reject(err)
+          this.started = null
```
sebn: Isn't `this.started = null` already done in `stop()`?

src/remote/watcher.js
```diff
@@ -34,15 +34,30 @@ export default class RemoteWatcher {
   }
 
   start () {
-    this.watch()
-    this.intervalID = setInterval(this.watch, HEARTBEAT)
+    this.started = new Promise((resolve, reject) => {
+      this.started.resolve = resolve
+      const watch () => {
+        this.watch().catch(err => {
+          reject(err)
```
sebn: I suppose having `start()` rejecting when first `watch()` does may be useful.
But I would have intuitively assumed that `RemoteWatcher` can be considered "started" as soon as `setInterval()` was called, assuming the first `watch()` call was more related to "polling" / "timing", and not really a part of the "starting" process...
Also, doesn't it make `start()` resolution take some time, since `watch()` may await quite a lot of things?

src/app.js
nono: It's used only once for the moment. We will refactor when it will be used again somewhere else

src/app.js
```diff
@@ -238,13 +237,15 @@ class App {
     } else {
       log.error('No configuration found, please run add-remote-cozy' +
                 'command before running a synchronization.')
-      __guardFunc__(callback, f => f(new Error('No config')))
+      callback(new Error('No client configured'))
```
sebn: Makes for another `Config#ensureValid()` candidate :)

src/app.js
```diff
@@ -306,10 +306,3 @@ class App {
 App.initClass()
 
 export default App
-
-function __guardFunc__ (func, transform) {
-  return typeof func === 'function' ? transform(func) : undefined
-}
-function __guard__ (value, transform) {
-  return (typeof value !== 'undefined' && value !== null) ? transform(value) : undefined
-}
```
sebn: 👍

src/remote/watcher.js
nono: I've tried to subclass Error, but babel has some caveats on subclassing Error and it didn't work.

src/remote/watcher.js
nono: Yes, it was fixed in a commit later

src/app.js
```diff
@@ -207,43 +207,34 @@ class App {
...
     log.info('Run first synchronisation...')
-    this.sync.start(mode, err => {
-      this.sync.stop(() => {})
-      if (err) {
-        log.error(err)
-      }
-      callback(err)
-    })
+    this.sync.start(mode)
   }
```
sebn: I may be misunderstanding something, but isn't this supposed to return a promise?

sebn: My bad, just saw `synchronize()` below.

src/app.js
```diff
@@ -207,43 +207,34 @@ class App {
...
-    if (this.config.isValid()) {
-      this.instanciate()
-      this.startSync(mode, callback)
-    } else {
+  synchronize (mode) {
+    if (!this.config.isValid()) {
       log.error('No configuration found, please run add-remote-cozy' +
                 'command before running a synchronization.')
-      callback(new Error('No client configured'))
+      throw new Error('No client configured')
```
sebn: `Config#ensureValid()`

src/bin/cli.js
```diff
@@ -58,9 +58,8 @@ let sync = function (mode, args) {
       return log.log(`${what} ${filename} (unknown size)`)
     }
   })
-  return app.synchronize(mode, (err) => {
-    if (err) { return process.exit(1) }
-  })
+  return app.synchronize(mode)
+    .catch(() => process.exit(1))
```
sebn: Won't this swallow exceptions? Or do we assuming they'll be logged every time?

src/app.js
sebn: Ok

src/app.js
nono: Nope, not another, it's the same line

src/app.js
nono: Still the same line

src/bin/cli.js
nono: It was fixed later

src/remote/watcher.js
nono: `start()` is slow, but starting the `LocalWatcher` is slow too. And, IIRC, it's better to have synchronized all the metadata before starting to poll the sync queue for applying changes.

sebn: Ok, makes sense.

src/remote/watcher.js
sebn: Yeah, I'm discovering that everything error-related in js is a lot of fun...

I was thinking of [es5-style](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error#Custom_Error_Types).

Regarding the *incorrect lineNumber and fileName* issue, I suppose there's a way to make it work on node since some modules seems to do so, e.g. node-fetch:

https://github.com/bitinn/node-fetch/blob/master/src/fetch-error.js

This makes me think that [my own implementation of `DirectoryNotFound`](https://github.com/cozy-labs/cozy-desktop/blob/master/src/remote/cozy.js#L21-L23) is not really production-ready...

nono:
:ship:

gui/app/locales/es.json
```diff
@@ -98,11 +102,11 @@
...
-  "Unlink Title": "Deconnection",
-  "Unlink Message": "Are you sure you want to unlink your Cozy?",
-  "Unlink Detail": "Your file will no longer be synchronized with your Cozy.",
-  "Unlink OK": "Log out",
-  "Unlink Cancel": "Cancel",
+  "Unlink Title": "Desconexión",
+  "Unlink Message": "¿Está usted seguro que quiere desvincular su Cozy?",
+  "Unlink Detail": "Su archivo ya no será sincronizado con su Cozy.",
+  "Unlink OK": "Cerrar sesión",
+  "Unlink Cancel": "Anular",
```
sebn: I'll trust you on this since I don't speak spanish at all 😃

nono: Neither do I. I've just pulled from transifex

PR #555, opened 2017-03-02
Update remote folder and do not overwrite cozy-stack mimetype

sebn:


src/merge.js
```diff
@@ -243,8 +243,9 @@ class Merge {
         doc.creationDate = file.creationDate
         if (this.sameBinary(file, doc)) {
           if (doc.size == null) { doc.size = file.size }
-          if (doc.class == null) { doc.class = file.class }
-          if (doc.mime == null) { doc.mime = file.mime }
+
+          doc.class = file.class
+          doc.mime = file.mime
```
sebn: @nono I'm particularly interested in your feedback on this.

sebn:
Not sure why the push build failed on node 7...

nono:
Yesterday, the build for node 7 failed because it wasn't able to compile leveldown. But it looks like it was fixed since.

src/merge.js
nono: @sebn I don't see how this will work. If local and remote see the same file with different mime-type, the doc will loop with one time the mime-type from local and next time the mime-type from remote, no?

sebn: Actually the looping stops as soon as this fix is enabled.
But you'right it won't work: it will actually prevent changing the mimetype once the file's metadata have been created in pouch.
I'll need to come up with another solution.

sebn: Actually I may be wrong too.
Better see this change's context, including [this line](https://github.com/cozy-labs/cozy-desktop/pull/555/files/b9139c4d29bb61e5fd859563aba789e3ccd1901f#diff-d33c8624f2fb2ac1fc83292e027196dbL241)

sebn: The goal was to prevent changing the mimetype when binary content didn't change.

sebn:
Rebased on master.
Doing some manual tests again to see whether the mimetype fix actually breaks things.

src/merge.js
sebn: You were right: it didn't work.

sebn:
I reverted the mimetype fix.
And I removed the local mimetype detection instead, which solves the mimetype overwriting issue.
The mime module it still in `yarn.lock` because other modules depend on it.

I also fixed a bug in `Remote#moveFile()` (which was causing the metadata to be saved without the remote id/rev).

Tests pass locally.
Everything seems to work for every scenarios.

:ship:

nono:
I'm wondering it is still make sense to keep the mime/class in pouch

nono:
Well, let's keep them. It's probably used in the GUI to display an icon, or something like that

PR #559, opened 2017-03-03
Local watcher renamed methods

sebn:


sebn:
Regarding `hasPendingChild()`, I may possibly need to introduce some `hasPendingAncestor()`, hence the renaming.

PR #561, opened 2017-03-06
Local watcher delay dir deletion

sebn:


sebn:
Not sure why [`done()` was called instead of `check()`](https://github.com/cozy-labs/cozy-desktop/blob/9d116707c2def278be183f7c889f3d04ab85d258/src/local/watcher.js#L304-L312).

From what I understand, `check()` was never used.

PR #566, opened 2017-03-07
Several fixes for the GUI

nono:


gui/app/main.js
```diff
@@ -421,7 +421,8 @@ app.on('ready', () => {
 app.on('window-all-closed', () => {})
 
 ipcMain.on('register-remote', (event, arg) => {
-  desktop.config.cozyUrl = arg.cozyUrl
+  const cozyUrl = desktop.checkCozyUrl(arg.cozyUrl)
+  desktop.config.cozyUrl = cozyUrl
```
sebn: Why does the config need to be changed here (and not in CLI)?

src/app.js
```diff
@@ -93,15 +93,20 @@ class App {
...
     if (!['http:', 'https:'].includes(parsed.protocol) || !parsed.hostname) {
       let err = new Error(`Your URL looks invalid: ${cozyUrl}`)
       log.warn(err)
       throw err
     }
+    return cozyUrl
+  }
+
+  // Return a promise for registering a device on the remote cozy
+  registerRemote (cozyUrl, redirectURI, onRegistered, deviceName) {
```
sebn: Except for the config change above, couldn't we call `checkCozyUrl` in `registerRemote`?
It could even be done in `Registration` since `checkCozyUrl` (and `parseCozyUrl`) doesn't really depend on `App`.

gui/app/main.js
nono: In the CLI, the user gives the URL and the path to sync, and then cozy-desktop opens his browser for OAuth, and when done, it persists all the config. Easy, CLI -> LIB -> CLI.

In the GUI, it is more complicated. First the user gives the URL, then cozy-desktop modifies the electron URL, then the user confirms, then cozy-desktop saves some informations, then the user gives the path, etc. It's a lot of context switching between the GUI and LIB, so there are some artifacts like this.

src/app.js
nono: Well, `checkCozyUrl` also modifies the URL to add `https://` if it is missing. And we need the returned URL in the GUI.

sebn:
Those can be fixed later.

PR #567, opened 2017-03-07
Extract Pending and PendingMap from LocalWatcher

sebn:


src/utils/pending.js
```diff
@@ -0,0 +1,59 @@
...
+      }
+    }
+    return false
+  }
+
+  clear (path: string) {
+    const pending = this.map.get(path)
+    if (pending) {
+      pending.stopChecking()
+      delete this.map.delete(path)
```
nono: I'm surprised by this line. What does the first `delete` do?

sebn:
Although tests pass locally, I already saw the Travis failure, but I couldn't find a way to fix it...

src/local/watcher.js
```diff
@@ -263,53 +255,47 @@ class LocalWatcher {
...
-    this.pending[folderPath] = {
-      clear,
-      done,
-      check,
-      interval: setInterval(check, 350)
+      if (!this.pending.hasPendingChild(folderPath)) {
+        execute()
+      }
     }
+    this.pending.add(folderPath, {stopChecking, execute})
```
nono: For `stopChecking`, it's the pending map that removes `folderPath`. But for `execute`, it is the `execute` function that removes it. I think it could be clearer to remove it from the same class for both cases. I don't mind if it is the `stopChecking` and `execute` functions, or in the `pendingMap`, but I'd prefer something consistent for both use cases.

sebn: Makes sense. I think I just forgot the `execute` case.

nono:
Shouldn't this line be changed?

https://github.com/cozy-labs/cozy-desktop/blob/master/test/unit/local/index.js#L22

src/utils/pending.js
sebn: Oops, refactoring artifact.

nono:
You should also try to remove the `return`s in test/unit/local/index.js, as the tests are using `done()`.

sebn:
Rebased on master, 🚢 as soon as Travis is green.

PR #568, opened 2017-03-07
Sync to trash

sebn:
WIP

Right now, this works because chokidar triggers file events before dir ones (should we rely on this?)

### DIFF
--- a/src/local/watcher.js
+++ b/src/local/watcher.js
@@ -255,6 +255,7 @@ class LocalWatcher {
   // It can be a file moved out. So, we wait a bit to see if a file with the
   // same checksum is added and, if not, we declare this file as deleted.
   onUnlinkFile (filePath) {
+    // TODO: Extract delayed execution logic to utils/pending
     let timeout
     const stopChecking = () => {
       clearTimeout(timeout)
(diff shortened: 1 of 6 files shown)


nono:
> Right now, this works because chokidar triggers file events before dir ones (should we rely on this?)

I don't think it's chokidar but the local watcher that makes sure that file deletions are trigger before dir ones. So, yes, we can rely on this behaviour.

Cf https://github.com/cozy-labs/cozy-desktop/blob/master/src/local/watcher.js#L279-L280

sebn:
> I don't think it's chokidar but the local watcher

My bad, you're right. Forgetting this after pending so much time in the local watcher's code...

Note for myself: maybe we should make sure that the RemoteWatcher enforces the same rule...

sebn:
Main case works for me.
Tests actually pass, but build fails on node 6 & 7 because of a timeout on GUI deps install I believe.
I prefer to handle the `pending/utils` refactoring in another PR.

:ship: ?

nono:
The failure of travis is due to the install of elm packages, nothing related to this PR. Let's merge!

PR #573, opened 2017-03-09
Fix: babel-polyfill is a runtime dependency, not a development one

nono:
We will also need https://github.com/cozy/cozy-client-js/issues/88 for the packaging of cozy-desktop-gui (or include pouchdb-find in our package.json).


sebn:
I may soon make some use of pouchdb-find, I'll include it then.

PR #575, opened 2017-03-10
Extract merge methods as functions

sebn:


sebn:
Tests should be green soon.

PR #578, opened 2017-03-14
Remote cozy cleanup

sebn:


nono:
Looks good to me, but travis is not happy

sebn:
My bad, linter is fixed now.
Not sure why only the PR build fails, and why it only happens on node 7.
I believe it's not the first time the `Sync#pop` test fails only on node 7, but it was possibly already unstable on Travis even before migrating to the cozy stack...

sebn:
I would assume a false-positive for now. 🚢 ?

PR #588, opened 2017-03-20
Debug improvements

sebn:
Needs #587

### DIFF
--- a/src/local/watcher.js
+++ b/src/local/watcher.js
@@ -248,7 +248,6 @@ class LocalWatcher {
     log.chokidar.inspect(stats)
     if (folderPath === '') return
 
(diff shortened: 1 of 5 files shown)


src/local/watcher.js
```diff
@@ -238,6 +244,8 @@ class LocalWatcher {
 
   // New directory detected
   onAddDir (folderPath, stats) {
+    log.chokidar.debug(`${folderPath}: addDir`)
+    log.chokidar.inspect(stats)
```
nono: You keep the log.debug on line 251. Is it normal?

sebn: Fixed

PR #593, opened 2017-03-23
Test builders improvements

sebn:


sebn:
I'll eventually extract the id/rev building logic later.

PR #594, opened 2017-03-27
Windows fixes

sebn:


src/conversion.js
```diff
@@ -48,7 +48,7 @@ export function createMetadata (remote: RemoteDoc): Metadata {
 
 // Extract the remote path and name from a local id
 export function extractDirAndName (id: string): [string, string] {
-  const dir = path.dirname(`/${id}`)
+  const dir = '/' + id.split(path.sep).slice(0, -1).join('/')
```
nono: I'm surprised that `path.basename` works with windows paths, but not `path.dirname`. Maybe it is just the leading `/`

sebn: It does work, but as stated in the method comment, the goal is to extract the remote path, which uses slashes as separators.

sebn: I'll probably rename/split this method at some point to prevent confusion.

PR #597, opened 2017-03-28
Sync trash content metadata and restore dir

sebn:


sebn:
@nono oops, build is broken, please wait before reviewing

sebn:
Build is green now, but I'm still fixing a last weird error.

sebn:
@nono Sorry for the noise, everything should be fixed now. Feedback welcome!

src/local/index.js
```diff
@@ -323,6 +349,10 @@ class Local implements Side {
 
   // Delete a file or folder from the local filesystem
   destroy (doc, callback) {
+    if (inRemoteTrash(doc)) {
+      callback(null)
+      return
+    }
```
nono: Can you explain this case?

I have the impression that we should try to remove the file (and just don't throw an error if the file doesn't exist). For example, while cozy-desktop is uploading a big file, the user moves to the trash another file, and just after delete it permanently, this other file won't be deleted on the local disk after the upload finished.

src/merge.js
```diff
@@ -442,6 +447,27 @@ class Merge {
...
+      oldMetadata = await this.pouch.db.get(doc._id)
+    } catch (err) {
+      if (err.status === 404) {
+        log.debug(`${doc._id}: Nothing to trash`)
+        return
+      }
+      throw err
+    }
+    if (doc.docType !== oldMetadata.docType) {
+      this.resolveConflictAsync(side, doc)
```
nono: `return`?

src/sync.js
```diff
@@ -184,6 +184,11 @@ class Sync {
       case side == null:
         this.pouch.setLocalSeq(change.seq, callback)
         break
+      case (sideName === 'remote' && doc.toBeTrashed && !inRemoteTrash(doc)):
+        // File or folder was just deleted locally
+        this.trashLaterWithParentOrByItself(doc, side)
+        done()
```
nono: Probably just a TODO for later: the change is marked as done right now, but if later the operation fails (being offline, or the cozy-stack is upgrading, etc.), it won't be retried

src/local/index.js
sebn: > Can you explain this case?

When a folder is restored, since restoring is actually a move from the trash to the old location, the old folder will be destroyed.

Which makes me think I got it wrong since it'll actually be trashed, not destroyed...
So maybe this check should be moved to the `trash()` method.

> I have the impression that we should try to remove the file (and just don't throw an error if the file doesn't
> exist). For example, while cozy-desktop is uploading a big file, the user moves to the trash another file, and just
> after delete it permanently, this other file won't be deleted on the local disk after the upload finished.

Makes sense, I'll fix it.

src/merge.js
sebn: Oh my... Thanks!

src/sync.js
sebn: Makes sense.
I believe not calling `done()` immediately was prevent upcoming changes to pop up, making it impossible to "wait" for some trashed parent dir.
Not sure how to fix it yet.

nono: I think that just adding a TODO for the moment is OK.

Yes, we have to call `done()` to move to the next changes. But we can modify the document (`doc.toReallyTrash = true`) to make it pop again in sync later. We have to figure the details (when to do that, how to preserve the side that have to apply this, etc.) and I suspect it won't be easy.

sebn: Ok, thanks for the insight.

src/local/index.js
sebn: > Makes sense, I'll fix it.

Well, I'm not so sure in the end: `inRemoteTrash(doc)` actually checks whether `doc.path` starts with `.cozy_trash`. So removing the test would actually mean trying to delete some file in `~/Cozy/.cozy_trash/...`, while `~/Cozy/.cozy_trash` is supposed to never exist...

sebn:
Fixed issues except for `Local#destroy()`, see comment above.

src/local/index.js
nono: Okay, I really have some difficulties to reason about the trash

sebn: It was not easy to fix a case without breaking another.
Which makes me wonder whether we should bring back some integration tests.

Issue #38, opened 2014-11-23
Upload only file difference


frankrousseau:
That one is probably a big one. But when a file is changed, the whole file is uploaded. That's not very efficient.

nono:
See https://github.com/syncthing/syncthing/issues/107

IvanTurgenev:
documentation is not clear enough

PR #601, opened 2017-03-31
Auto launch on mac

nono:


sebn:
Maybe add some comment regarding the status of Teamwork/node-auto-launch#28?
Or Teamwork/node-auto-launch#38 (which was supposed to fix the former)?

sebn:
@nono or do you want to merge this quickly?

nono:
AFAICT, #38 is not suitable for electron apps, and #28 is still open. So, we can merge as is.

PR #604, opened 2017-04-03
Sync promisification

sebn:
Right now we only promisify `fileChanged()` and `folderChanged()` since this should allow us to progressively remove callback-based wrappers in `Sync`'s dependencies.

### DIFF
--- a/src/sync.js
+++ b/src/sync.js
@@ -499,7 +499,7 @@ class Sync {
           log.debug(`${doc.path}: will be trashed with parent directory`)
         } else {
           log.debug(`${doc.path}: should be trashed by itself`)
-          side.trash(doc, log.errorIfAny)
+          side.trashAsync(doc).catch(log.error)
         }
       }
     })
--- a/src/sync.js
+++ b/src/sync.js
@@ -352,77 +352,87 @@ class Sync {
   // If a file has been changed, we had to check what operation it is.
   // For a move, the first call will just keep a reference to the document,
   // and only at the second call, the move operation will be executed.
-  fileChanged (doc: Metadata, side: Side, rev: number, callback: Callback) {
+  async fileChangedAsync (doc: Metadata, side: Side, rev: number): Promise<void> {
     let from
     switch (true) {
       case doc._deleted && (rev === 0):
-        callback()
-        break
+        return
       case this.moveFrom != null:
         // $FlowFixMe
         from = (this.moveFrom: Metadata)
         this.moveFrom = null
         if (from.moveTo === doc._id) {
-          side.moveFile(doc, from, err => {
-            if (err) { this.moveFrom = from }
-            callback(err)
-          })
+          try {
+            await side.moveFileAsync(doc, from)
+          } catch (err) {
+            this.moveFrom = from
+            throw err
+          }
         } else {
           log.error('Invalid move')
           log.error(from)
           log.error(doc)
-          side.addFile(doc, function (err) {
-            if (err) { log.error(err) }
-            side.destroy(from, function (err) {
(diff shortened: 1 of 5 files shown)


sebn:
One test is still failing, but early feedback is welcome.

src/sync.js
```diff
@@ -352,77 +350,88 @@ class Sync {
...
-              if (err) { log.error(err) }
-              callback(new Error('Invalid move'))
-            })
-          })
+          try {
+            await side.addFileAsync(doc)
+          } catch (err) {
+            log.error(err)
+            try {
+              await side.destroyAsync(from)
```
nono: The nesting seems wrong here. We should call `destroyAsync(from)` even if `addFileAsync(doc)` has suceeded

src/sync.js
```diff
@@ -352,77 +350,88 @@ class Sync {
...
-            // Photos uploaded by cozy-mobile have no checksum,
-            // but it's useless to reupload the binary
-            side.updateFileMetadata(doc, old, callback)
-          } else {
-            side.overwriteFile(doc, old, callback)
-          }
-        })
+        let old
+        try {
+          old = this.pouch.getPreviousRevAsync(doc._id, rev)
```
nono: no `await`?

src/sync.js
```diff
@@ -431,30 +440,43 @@ class Sync {
...
-              if (err) { log.error(err) }
-              callback(new Error('Invalid move'))
-            })
-          })
+          try {
+            await side.addFolderAsync(doc)
+          } catch (err) {
+            log.error(err)
+            try {
+              await side.trashAsync(from)
```
nono: Nesting seems wrong here too

src/sync.js
sebn: Yep, I believe that's the last failing test :)

src/sync.js
```diff
@@ -352,77 +350,88 @@ class Sync {
...
-          } else {
-            side.overwriteFile(doc, old, callback)
-          }
-        })
+        let old
+        try {
+          old = this.pouch.getPreviousRevAsync(doc._id, rev)
+        } catch (err) {
+          // FIXME: overwriteFileAsync never uses old anyway
+          return side.overwriteFileAsync(doc, doc)
```
sebn: Old line 396 bothered me... Better explicitely pass `null` instead of `old`, or `doc` (which is worse)...
I'll need to fix `overwriteFileAsync()`

src/sync.js
sebn: Oh my... Thanks!

sebn: Worse is that `folderChanged()` tests are actually passing. I'll need to make sure they don't.

src/sync.js
sebn: Agreed, same mistake as below.

src/sync.js
```diff
@@ -431,30 +440,43 @@ class Sync {
...
-        break
+        return side.addFolderAsync(doc)
       default:
-        this.pouch.getPreviousRev(doc._id, rev, (_, old) => side.updateFolder(doc, old, callback))
+        let old
+        try {
+          old = await this.pouch.getPreviousRevAsync(doc._id, rev)
+        } catch (_) {
+          // FIXME: Side#updateFolderAsync()'s old param may be null?
+          return side.updateFolderAsync(doc, doc)
```
sebn: Same here, better pass `null`.

sebn: Or maybe call `addFolderAsync()` directly.

sebn:
Fixed issues and rebased on master.
Lint and tests pass locally.

sebn:
:ship: ?

PR #607, opened 2017-04-03
Fix the detection of a revoked device

nono:


src/sync.js
```diff
@@ -237,8 +237,8 @@ class Sync {
       if (err) { log.error(err) }
       if (err && err.code === 'ENOSPC') {
         callback(new Error('The disk space on your computer is full!'))
-      } else if (err && err.message === REVOKED) {
-        callback(err)
+      } else if (err && err.status === 400) {
+        callback(new Error(REVOKED))
```
sebn: Looking at [cozy-stack's VFS API](https://github.com/cozy/cozy-stack/blob/master/docs/files.md#patch-filesfile-id-and-patch-filesmetadata), it seems that some `PATCH /files/123` or `PATCH /files/metadata` request may trigger a `400` error code... Could this be an issue?

sebn: I suppose that, even in the worst case, it would just unnecessarily refresh the token. But the real error would be swallowed...

nono: Cozy-desktop won't try to move one directory inside one of its sub-directories. But, yes, it's not ideal to detect a revocation with a 400 status code. It's most reliable way to detect it for the moment, but not totally safe. I keep in mind to improve it in cozy-stack.

sebn: Sorry, I'm messing up revocation with token renewal... So I suppose a poorly handled 400 error could lead to the client being disconnected from the remote cozy... Or am I wrong?

nono: Yes, it will be detected as a revoked device. The user can then unlink the device and redo the OAuth 2 flow. It's far from ideal, but not a blocker too.

sebn: Makes sense!

PR #617, opened 2017-04-07
local/watcher: Use polling on Windows

sebn:
At least until we find a better solution.


nono:
Polling is really not a good idea on the long term. But let's use it until we find how to fix the open directories issue.

PR #618, opened 2017-04-07
Bunyan

sebn:
- Replace our custom logger with Bunyan
- Logs are now lines of json records
- The `bunyan` command allows us to format/filter them (by piping the command / logfile into it)
- The way we log things could still be improved (by adding fields, e.g. `side`), but this is a working 1st version
- We cannot chose a custom log location from the CLI anymore, we know always use the `.cozy-desktop/logs.txt` path as it was already done in the GUI.


sebn:
sample-bunyan-output

(this is test output, hence the `undefined` path.

sebn:
Build should be fixed now.

sebn:
I'll make sure everything works on Windows before merging.

PR #626, opened 2017-04-11
Detect offline

nono:


gui/app/main.js
```diff
@@ -292,26 +292,23 @@ const removeFile = (info) => {
...
-          total: +res.diskSpace.totalDiskSpace,
-          totalUnit: res.diskSpace.totalUnit
+          used: +res.attributes.used,
+          quota: +res.attributes.quota
         }
         sendToMainWindow('disk-space', space)
-      }
-    })
+      },
+      (err) => console.error(err)
```
sebn: Is the goal not to catch errors from the `(res) => { ... }` callback?

src/remote/index.js
```diff
@@ -106,20 +101,10 @@ export default class Remote implements Side {
       _rev: created._rev
     }
 
+    // TODO do we use the returned values somewhere?
```
sebn: I'm not sure either, but I would definitely prefer doing so instead of changing the given metadata...
But this is another story.

gui/app/main.js
nono: Well, I just transposed the code to replace `getDiskSpace` to `diskUsage` and use promises, without too much thinking. The goal was to have a GUI that doesn't crash

src/remote/watcher.js
```diff
@@ -71,7 +70,7 @@ export default class RemoteWatcher {
       await this.pouch.setRemoteSeqAsync(changes.last_seq)
       log.debug({event: 'end'}, 'No more remote changes for now')
     } catch (err) {
-      if (err.message === REVOKED) {
+      if (err.status === 400) {
         throw err
```
sebn: If we end up testing this in multiple places, maybe we could later extract the `if` + `throw` in some `ensureNotRevoked()` function...

PR #632, opened 2017-04-13
Fix deletion of a local file with the client stopped

nono:


src/merge.js
```diff
@@ -381,20 +382,25 @@ class Merge {
...
+    let docs = await this.pouch.byRecursivePathAsync(folder._id)
     // In the changes feed, nested subfolder must be deleted
     // before their parents, hence the reverse order.
     docs = docs.reverse()
     docs.push(folder)
     for (let doc of Array.from(docs)) {
-      markSide(side, doc, doc)
-      doc._deleted = true
-      delete doc.errors
+      if (doc.sides && (doc.sides[side] || 0) < (doc.sides[other] || 0)) {
```
sebn: Either wrap `(doc.sides[side] || 0)` in some `metadata` function, or ensure `sides[side]` is `0` by default?

(sorry, I didn't do it myself since this was a PoC)

src/merge.js
```diff
@@ -381,20 +382,25 @@ class Merge {
...
+        // TODO we should also preserve the parents of this folder
```
sebn: What happens right now if we don't preserve it?

nono: Bad things, we can lose the synchronization of some files. I will work on it today

PR #653, opened 2017-04-21
Shorten stacktraces unless DEBUG is set or on CI

sebn:


sebn:
`mocha-clean` is quite old but it still seems to work well.

PR #660, opened 2017-04-24
Fix for files overwriten on both sides

nono:


src/remote/index.js
```diff
@@ -119,11 +119,16 @@ export default class Remote implements Side {
...
     log.info(`${doc.path}: Uploading new file version...`)
     const stream = await this.other.createReadStreamAsync(doc)
-    const updated = await this.remoteCozy.updateFileById(doc.remote._id, stream, {
+    const options = {
       contentType: doc.mime,
       checksum: doc.md5sum,
-      lastModifiedDate: new Date(doc.updated_at)
-    })
+      lastModifiedDate: new Date(doc.updated_at),
+      ifMatch: ''
```
sebn: Wouldn't it work without the default empty string?

nono: It works but flow is not happy

src/metadata.js
```diff
@@ -182,17 +182,7 @@ export function sameFile (one: Metadata, two: Metadata) {
...
-  } else if ((one.md5sum != null) && (one.md5sum === two.md5sum)) {
-    return true
-  } else if ((one.remote != null) && (two.remote != null)) {
-    let oneId = one.remote._id
-    let twoId = two.remote._id
-    return (oneId != null) && (oneId === twoId)
-  } else {
-    return false
-  }
+  return one.md5sum === two.md5sum
```
sebn: Not completely sure of the implications, but definitely feels safer.

src/remote/index.js
sebn: Oh yeah... Flow considers object litterals as "closed" by default.
If you declare with `const options: Object = ...`, I believe it should work (although I agree specifying that `options` is an `Object` is not really valuable...).

src/metadata.js
nono: With cozy-v2, it was possible to have a file on the cozy without a checksum (files uploaded by the mobile for example). So, we were doing some dirty tricks.

PR #663, opened 2017-04-27
Fix conflict between a deletion and an update by keeping the updated file

nono:


sebn:
Not sure why Travis is broken. I can't see the logs. Rebuild?
Do you want to wait for a cozy-client-js build?
Maybe update unit tests?
Flow complains locally.

PR #664, opened 2017-04-27
Ci fixes

sebn:


sebn:
I spent quite some time on this.
Some occasionally failing tests seem to work when adding delays everywhere...
I couldn't make `onUnlinkDir` work though.

I'm not sure it's worth spending too much time on this right now.

sebn:
Still not found why PR build seems to fail almost every time on AppVeyor...

sebn:
I checked *Enable secure variables in Pull Requests from the same repository only* in the AppVeyor settings, and it seems to work.

I assume PR from forked repos will fail because of the missing variables to set up the test env using the remote instance.

:ship: ?

PR #669, opened 2017-05-02
Platform incompatibilities

sebn:


gui/app/main.js
```diff
@@ -318,6 +333,47 @@ const chooseSyncPath = () => {
   sendToMainWindow('registration-done')
 }
 
+const incompatibilitiesErrorMessage = (i) => {
+  const reasons = []
+  const docType = translate(`Helpers ${i.docType}`)
+  if (i.reservedChars) {
+    reasons.push(
+      interpolate(
+        translate('Error {0} names cannot include « {1} » characters'),
```
nono: I think that « ... » is used in french, but not in english, where “...” is prefered

src/utils/regexp.js
```diff
@@ -0,0 +1,22 @@
+/* @flow */
+
+// Same as String#match, but return all the matches
```
nono: Isn't the `g` flag on the regexp made for that?

```js
string.match(/[aeiou]/g)
```

gui/app/main.js
```diff
@@ -352,6 +408,12 @@ const startSync = (force) => {
       addFile(info)
       removeFile(old)
     })
+    desktop.events.on('platform-incompatibilities', incompatibilities => {
+      // TODO: Index incompatibilities to prevent notification overload
+      incompatibilities.forEach(i => {
+        sendErrorToMainWindow(incompatibilitiesErrorMessage(i))
```
nono: A debounce on `sendErrorToMainWindow` is a good idea for your TODO

src/remote/watcher.js
```diff
@@ -132,6 +135,11 @@ export default class RemoteWatcher {
     let doc: Metadata = conversion.createMetadata(remote)
     const docType = doc.docType
     ensureValidPath(doc)
+    const incompatibilities = pathPlatformIncompatibilities(doc)
+    if (incompatibilities) {
+      this.events.emit('platform-incompatibilities', incompatibilities)
+      return
+    }
```
nono: I'd have put this check in `Prep`, but it's OK for here

src/utils/regexp.js
sebn: I believe I generalized it without using other match args anymore. Thanks!

sebn:
Added fixes according to feedback. I'll merge as soon as build is green.

sebn:
@nono Do you think we should fix the debounce issue right now?

I suppose we could just add and use `lodash.debounce`.

nono:
@sebn it's probably easy to do, so better to do it now

sebn:
:ship: ?

PR #677, opened 2017-05-05
No trash path warnings

sebn:


sebn:
The hook that fails on both AppVeyor builds has nothing to do with this PR.
Tests pass locally.
I'll merge and fix this issue later.

PR #682, opened 2017-05-09
Windows updates

sebn:


gui/package.json
```diff
@@ -58,5 +61,8 @@
     "watch:elm": "chokidar 'lib/**/*.elm' -c 'npm run build:elm'",
     "watch:js": "chokidar app/main.js app/ports.js -c 'npm run lint'",
     "elm-upgrade": "npm install --save-dev elm@latest elm-upgrade@latest && ./node_modules/.bin/elm-upgrade"
+  },
+  "dependencies": {
+    "electron-builder-squirrel-windows": "^17.4.0"
```
nono: Is it normal to have this dependency on `gui/package.json`, and not `gui/app/package.json`?

sebn: Right file, but wrong section I believe... Thanks!

sebn: Fixed

PR #696, opened 2017-05-15
Fix locale

sebn:
Locale cannot be determined from env vars on Windows.
To be portable, we must use electron's standard app.getLocale() function.
But it cannot be used before the ready event is fired.

### DIFF
--- a/gui/app/main.js
+++ b/gui/app/main.js
@@ -21,19 +21,22 @@ const autoLauncher = new AutoLaunch({
 const desktop = new Desktop(process.env.COZY_DESKTOP_DIR)
 const lastFilesPath = path.join(desktop.basePath, 'last-files')
 
-app.locale = (() => {
-  const env = process.env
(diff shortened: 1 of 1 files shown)


gui/app/main.js
```diff
@@ -21,19 +21,23 @@ const autoLauncher = new AutoLaunch({
...
 
-app.locale = (() => {
-  const env = process.env
-  const envLocale = env.LC_ALL || env.LC_MESSAGES || env.LANG || env.LANGUAGE
-  if (envLocale && envLocale.match(/^fr_/i)) {
-    return 'fr'
+app.locale = 'en'
+const setUpLocale = () => {
+  const locale = app.getLocale()
+  console.log(JSON.stringify(locale))
```
nono: `console.log` spotted

sebn: Oops thanks 😛

sebn: Fixed

PR #697, opened 2017-05-15
Force single instance of the GUI

nono:


sebn:
eslint doesn't approve, though 😄

sebn:
@nono wdyt?

nono:
LGTM

PR #698, opened 2017-05-15
Release 3.0.0 beta.1 again

sebn:


sebn:
I'll fix (unless you do so) and include #698 before releasing.

PR #724, opened 2017-05-30
Send mail to the support

nono:


gui/lib/Help.elm
```diff
@@ -124,6 +124,33 @@ view helpers model =
                     ]
                 ]
             , h2 [] [ text (helpers.t "Help Official Support") ]
+            , if model.status == Success then
+                p [ class "message--success" ]
+                    [ text (helpers.t "Help Your mail has been sent. We will try to respond to it really soon!") ]
```
sebn: Will the `model.status` be reset at some point, allowing me to send another email?
Or should we add some `Send another mail` button?

gui/lib/Help.elm
```diff
@@ -124,6 +124,33 @@ view helpers model =
...
+              else
+                Html.form [ class "send-mail-to-support" ]
+                    [ case model.status of
+                        Error error ->
+                            p [ class "message--error" ]
+                                [ text ("Error: " ++ error) ]
```
sebn: Will this be a low-level error message? Should we show it to the user, or log it and display a more user-friendly message?

gui/lib/Help.elm
nono: You can close and reopen the window ;-)
I think we can improve this after the redesign of the GUI.

sebn: That's what I was thinking of 👍 
It's so nice to get this feature back!

PR #725, opened 2017-05-30
Fix detection of renaming files when the client is stopped

nono:


nono:
Well, it does not work very well. There is a race condition with the `ready` event from chokidar, and computing the checksums.

PR #727, opened 2017-05-31
Fix detection of renaming files when the client is stopped

nono:


src/local/watcher.js
```diff
@@ -233,7 +233,9 @@ class LocalWatcher {
               log.info({path: filePath}, 'file added')
               this.prep.addFileAsync(SIDE, doc).catch(logError)
             } else {
-              const same = find(docs, d => this.pending.hasPath(d.path))
+              const same = find(docs, this.paths
+                  ? d => !fs.existsSync(d.path)
+                  : d => this.pending.hasPath(d.path))
```
sebn: I believe we should at some point either rename `this.path` to `this.initialPaths` and have `this.isInitialScan = () => this.initialPaths != null`, or even have some kind of `this.initialScan = {paths: []}` object (so we can do `if (this.initialScan) this.initialScan.paths.push(...)` and we don't have a remaining useless `this.initialPaths` property once scan is completed).

sebn: (can be done later thought, just wanted to share my thoughts)

src/local/watcher.js
```diff
@@ -338,13 +340,14 @@ class LocalWatcher {
...
+              this.onUnlinkFile(doc.path)
             } else {
-              log.info({path: doc.path}, 'deleted while client was stopped')
-              await this.prep.trashDocAsync(SIDE, doc)
+              this.onUnlinkDir(doc.path)
             }
           }
           delete this.paths
-          callback()
+          setTimeout(callback, 3000)
```
sebn: I assume having the usual unlinked file/dir timeouts + waiting 3 seconds after initial scan prevents the "onready" race condition?

sebn:
Arg... PR rebase & merge takes the last commit message instead of the PR one 😭

sebn:
My bad, it actually did a rebase as usual. Should have squashed instead.

src/local/watcher.js
nono: Exactly

PR #731, opened 2017-05-31
Fix conflict between a local update and a remote trash

nono:


sebn:
Shouldn't we test this? Seems affordable... wdyt?

nono:
Well, a unit test is not interesting. And a functional test is really hard to setup.

sebn:
Ok

PR #732, opened 2017-06-01
Open external links in a browser

nono:


sebn:
Do you think it'll conflict with #736 ?

nono:
No, you just move the files modified in this PR. It should not conflict (and it will be easy to fix anyway).

PR #736, opened 2017-06-02
Electron builder

sebn:


sebn:
Windows tests pass locally.
And fswin is not yet usable from master, so we still need to stick to my fork (I'll open some PR soon).
Better merge this now so we don't get conflict because of the new project tree.

PR #754, opened 2017-06-26
Permanent deletion

sebn:


cli/src/remote/watcher.js
```diff
@@ -82,45 +86,65 @@ export default class RemoteWatcher {
...
   // Pull a single file/dir metadata, given its id
-  async pullOne (id: string): Promise<*> {
-    const doc: ?RemoteDoc = await this.remoteCozy.findMaybe(id)
-
-    if (doc != null) {
-      return this.onChange(doc)
+  async pullOne (change: {id: string, deleted?: true}): Promise<*> {
+    const {id, deleted} = change
+    let doc
+    if (!deleted) doc = await this.remoteCozy.findMaybe(id)
```
nono: @aenario we don't allow the `include_docs` parameter for the changes feed in cozy-stack. Do you see a reason for that? Else, I think it would be simpler to use it here.

sebn: https://github.com/cozy/cozy-stack/pull/777 would definitely help 😄

sebn:
@nono 🚢 ?

sebn:
Oops, doc is quite bad... (I'll fix it)
Do you see any other issue?

sebn:
Fixed doc, also extracted remote file path retrieval logic (was duplicated).

cli/src/remote/cozy.js
```diff
@@ -74,15 +74,16 @@ export default class RemoteCozy {
...
+    const {last_seq, results} = await this.client.data.changesFeed(FILES_DOCTYPE, options)
+    let docs = results.filter(r => !specialId(r.id)).map(r => r.doc)
 
-    return {
-      last_seq: json.last_seq,
-      ids: json.results
-        .map(result => result.id)
-        .filter(id => !specialId(id))
+    for (const doc of docs) {
+      if (doc.type === 'file') await this._setPath(doc)
```
sebn: @nono Maybe we could have retrieved all parent dirs at once...
Not sure which index we should use though... Just `_id`? Passing the parent ids (without duplicates) as a mango query? WDYT?

sebn: s/mango query/mango selector/

nono: It looks like a premature optimization to me

sebn: ok, thanks

nono: Oh, the parents are fetched from the remote couchdb, not the local pouchdb! I see why you want to limit the number of calls

nono: You can call [`/db/_all_docs` with the parent ids as `keys`](http://docs.couchdb.org/en/2.0.0/api/database/bulk-api.html)

PR #769, opened 2017-07-21
Pin synced folder to quick access on Windows 10+

sebn:


gui/main.js
```diff
@@ -518,57 +518,9 @@ const execSync = (cmd) => {
...
-  const {syncRootId, clsid} = windowsConfig
-
-  execSync(`reg delete HKLM\\Software\\Microsoft\\Windows\\CurrentVersion\\Explorer\\SyncRootManager\\${syncRootId} /f`)
-  execSync(`reg delete HKCU\\Software\\Classes\\CLSID\\{${clsid}} /f`)
-  execSync(`reg delete HKCU\\Software\\Microsoft\\Windows\\CurrentVersion\\Explorer\\Desktop\\NameSpace\\{${clsid}} /f`)
-  execSync(`reg delete HKCU\\Software\\Microsoft\\Windows\\CurrentVersion\\Explorer\\HideDesktopIcons\\NewStartPanel /v {${clsid}} /f`)
-}
-*/
+const win10PinToHome = (path) => {
+  execSync(`powershell -Command "(New-Object -COM shell.application).Namespace('${path}').Self.InvokeVerb('pintohome')"`)
```
sebn: Maybe path should be escaped properly...

sebn: More a user concern than a security issue, a legitimate path could contain a single quote.

It would be nice to progressively extract some functions from `gui/main.js` and unit-test them... WDYT?

sebn:
I can confirm it works now with a path containing single-quotes. Double-quotes are forbidden in Windows filenames and I assume other characters won't be interpreted by powershell or the cmd interpreter.

:ship: ?

PR #794, opened 2017-08-28
Fix windows path normalization

sebn:


sebn:
- 2/6 Travis jobs fail on different merge unit tests, I believe this has nothing to do with this PR. But we'll have to investigate why those tests fail randomly at some point.
- The AppVeyor branch build fails while installing npm packages (this happens quite often those days, not sure why...)

PR #796, opened 2017-08-28
Get rid of fswin

sebn:
Better hide directories with a simple Windows command.
1 dependency less.
No need to depend on a fork anymore.
And we won't need to recompile it for electron everytime we run yarn.


sebn:
@aenario Please wait before merging (I want to make sure it actually works)

sebn:
Ok, it works! 🚢

sebn:
(merging as soon as Travis is green)

PR #801, opened 2017-09-07
One-page settings

aenario:
@sebn you can start review 

Still need to do 
- [x] figure out why disk usage does not work 
- [x] put back help as a separate screen.
- [x] minor css improvements.

screenshot from 2017-09-07 10-09-16


aenario:
Ready to squash & merge :rocket:

gui/src/elm/Settings.elm
```diff
@@ -58,48 +82,125 @@ update msg model =
...
-view : Helpers -> Model -> Html Msg
-view helpers model =
+humanReadableDiskValue : Helpers -> Float -> String
+humanReadableDiskValue helpers v =
+    (toString (round (v / 1000000)) ++ " M" ++ (helpers.t "Account b"))
+
+
+diskQuotaLine : Helpers -> Model -> Html Msg
+diskQuotaLine helpers model =
+    if model.disk.quota == 0 then
```
sebn: Maybe introduce some `hasDiskQuota` function?
And reuse it so you don't need to repeat the div/text/values separator?

PR #810, opened 2017-09-14
Local event buffer

sebn:


cli/src/local/watcher.js
```diff
@@ -125,6 +137,26 @@ class LocalWatcher {
...
+          break
+        case 'addDir':
+          this.onAddDir(e.path, e.stats)
+          break
+        case 'change':
+          this.onChange(e.path, e.stats)
+          break
+      }
+    })
+    _.each(unlinkDirEvents, e => { this.onUnlinkDir(e.path) })
```
sebn: @aenario Tests pass with this version, but I'm almost sure it's actually broken.
Here we naively change the order of the events, which may fix some event order issues but probably also breaks other cases...

Passing the events to the LocalWatcher methods as they come from the buffer would probably match the previous behavior more closely, but it breaks the tests.

sebn:
@aenario Code looks good to me (WDYT?).

I'll do a bit more manual testing on both platforms and we should be good to go 😄

sebn:
- [ ] Logging should be updated too, I'll fix it.

cli/src/local/watcher.js
```diff
@@ -125,11 +135,37 @@ class LocalWatcher {
...
+      }
+    }
+  }
+
   stop () {
     if (this.watcher) {
       this.watcher.close()
       this.watcher = null
     }
+    this.buffer.switchMode('idle')
```
aenario: Shouldn't we flush there ?

sebn: The buffer flushes automatically when switching mode, because in both cases we want to flush anyway.

sebn:
- [ ] Not done with manual testing yet.

PR #811, opened 2017-09-19
Systray ui

aenario:
The code needs some cleanup but the function should be here.

### DIFF
--- a/gui/main.js
+++ b/gui/main.js
@@ -2,148 +2,75 @@
 
 require('babel-polyfill')
 
-const childProcess = require('child_process')
 const Desktop = require('cozy-desktop').default
-const electron = require('electron')
 const notify = require('electron-main-notification')
 
 const debounce = require('lodash.debounce')
 const path = require('path')
-// const uuid = require('node-uuid')
 
 const autoLaunch = require('./src/main/autolaunch')
 const lastFiles = require('./src/main/lastfiles')
+const tray = require('./src/main/tray')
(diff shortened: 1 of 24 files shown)


gui/ports.js
```diff
@@ -38,6 +39,7 @@ ipcRenderer.on('registration-error', (event, err) => {
   elmectron.ports.registrationError.send(err)
 })
 ipcRenderer.on('registration-done', (event) => {
+  console.log("registrationDone in port")
```
nono: Debug?

gui/main.js
```diff
@@ -2,302 +2,75 @@
...
-// const uuid = require('node-uuid')
 
-const {spawn} = childProcess
-const {app, BrowserWindow, dialog, ipcMain, Menu, shell, session} = electron
-const autoUpdater = require('electron-updater').autoUpdater
-const autoLauncher = new AutoLaunch({
-  name: 'Cozy-Desktop',
-  isHidden: true
-})
+const autoLaunch = require('./src/main/autolaunch')
```
nono: :+:1 it's cool to have finally split `main.js`. It started as a small file, and then added a function, the next time an if/else to handle a special condition, and after that put another entry in a menu, etc. The file grows and grows but it was never a big change that justify to take time to refactor... until now! Yipee!

sebn: Well, as discussed with @aenario yesterday, last time I tried to split it, it broke electron-builder... 😭

PR #815, opened 2017-09-22
Remove v2 integration tests

sebn:
They are in git history anyway.


sebn:
My bad, relative imports were broken in v3 integration tests,. It should be fixed now.

PR #816, opened 2017-09-22
Remove installed-check

sebn:
I don't remember it showing any warning since we switched to yarn, so I assume it is not so useful anymore.

Looks like updating the `yarn.lock` file also fixes unrelated issues:

Before:

    cd cli/
    rm -rf node_modules/
    yarn
    yarn check # 7 errors

After:

    yarn remove installed-check
    yarn check # 1 error


sebn:
Rebased. Let's wait for travis before merging.

sebn:
3 jobs ok. Merging...

PR #820, opened 2017-09-26
Promisify checksumer

sebn:
So we can await for multiple checksum promises at once.
Useful to process an event batch.

### DIFF
--- a/cli/src/local/index.js
+++ b/cli/src/local/index.js
@@ -202,7 +202,7 @@ class Local implements Side {
       next => {
         if (doc.md5sum != null) {
           // TODO: Share checksumer instead of chaining properties
-          this.watcher.checksumer.push(tmpFile, function (err, md5sum) {
+          this.watcher.checksumer.push(tmpFile).asCallback(function (err, md5sum) {
             if (err) {
               next(err)
             } else if (md5sum === doc.md5sum) {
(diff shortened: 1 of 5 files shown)


sebn:
Ready to squash!

sebn:
Randomly failing test on Travis (already known issue).

cli/src/local/index.js
```diff
@@ -202,10 +202,8 @@ class Local implements Side {
       next => {
         if (doc.md5sum != null) {
           // TODO: Share checksumer instead of chaining properties
-          this.watcher.checksumer.push(tmpFile, function (err, md5sum) {
-            if (err) {
-              next(err)
-            } else if (md5sum === doc.md5sum) {
+          this.watcher.checksumer.push(tmpFile).catch(next).then(md5sum => {
```
aenario: :warning: next may be called twice with this code : the catch returns a promise which is always resolved.

The correct way to transform a promise into a callback is to use Promise#asCallback

sebn: My bad, I messed up trying to use `fromCallback` instead of `asCallback` and gave up since I didn't understand why it was not working.

- [ ] Use `Promise#asCallback()`

sebn:
@aenario Fixed and rebased. Waiting for CI...

sebn:
Bluebird's `asCallback` detects an issue that was fixed in a following PR, hence the cherry-picked commit so the CI stays green.

PR #822, opened 2017-09-26
Promisify more LocalWatcher methods

sebn:
Needs https://github.com/cozy-labs/cozy-desktop/pull/821

PR #823, opened 2017-09-26
Checksum on flush

sebn:
Needs https://github.com/cozy-labs/cozy-desktop/pull/822

### DIFF
--- a/cli/src/local/event_buffer.js
+++ b/cli/src/local/event_buffer.js
@@ -1,7 +1,7 @@
 /* @flow */
 
 type EventBufferMode = 'idle' | 'timeout'
-type FlushCallback<EventType> = (EventType[]) => void
+type FlushCallback<EventType> = (EventType[]) => any
 
 // An event buffer.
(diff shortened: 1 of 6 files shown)


sebn:
Needs https://github.com/cozy-labs/cozy-desktop/pull/822

cli/src/local/event_buffer.js
```diff
@@ -1,7 +1,7 @@
 /* @flow */
 
 type EventBufferMode = 'idle' | 'timeout'
-type FlushCallback<EventType> = (EventType[]) => void
+type FlushCallback<EventType> = (EventType[]) => any
```
aenario: why this change ?

sebn: Needed to make `LocalWatcher#handleEvents()` async and use `await` statements in it.
Otherwise `yarn flow` complains about the return type.

sebn: It could also turn out to be useful in future tests.

PR #824, opened 2017-09-26
Systray ui - codestyles fixes

aenario:


gui/src/main/window_behaviour.js
```diff
@@ -1,29 +1,104 @@
...
 
-module.exports.noMenu = (win) => {
-  win.setMenu(null)
-  win.setAutoHideMenuBar(true)
-}
+const log = require('cozy-desktop').default.logger({
+  component: 'windows'
+})
+
+module.exports = class WindowManager {
```
sebn: Maybe rename `gui/src/main/window{_behaviour -> _manager}.js`?
Or even remove suffix and rename `WindowManager` to `Window`?
(can be done later)

I wonder whether we could have used composition over inheritance (we're just factoring code here, there's no need to provide multiple `WindowManager` implementations to some consumer code)... But it wouldn't be so easy to achieve, event handlers would be particularly hard to extract...

And the new version is definitely a huge improvement anyway 👍

gui/src/main/tray.js
```diff
@@ -17,10 +20,13 @@ module.exports.init = (app, cb) => {
...
-    { label: translate('Tray Quit application'), click: app.quit }
-  ])
-  tray.setContextMenu(cm)
+  if (process.platform !== 'darwin') {
+    const cm = Menu.buildFromTemplate([
+     { label: translate('Tray Quit application'), click: app.quit }
+    ])
+    tray.setContextMenu(cm)
+  }
+  module.exports.setState('idle')
```
sebn: Note for later: mixing export with instance method feels a bit weird...

gui/src/main/tray.js
```diff
@@ -17,10 +20,13 @@ module.exports.init = (app, cb) => {
   tray.on('right-click', clicked)
   tray.on('double-click', clicked)
   tray.setToolTip('loading')
-  let cm = Menu.buildFromTemplate([
-    { label: translate('Tray Quit application'), click: app.quit }
-  ])
-  tray.setContextMenu(cm)
+  if (process.platform !== 'darwin') {
```
sebn: Maybe we should add a comment with a link to the GH issue/workaround?

gui/src/main/tray.window.js
```diff
@@ -12,114 +12,100 @@ const log = require('cozy-desktop').default.logger({
...
-    pos = (process.platform === 'win32') ? 'bottomRight' : 'topRight'
-  } else {
-    pos = (process.platform === 'win32') ? 'trayBottomCenter' : 'trayCenter'
+  create () {
+    log.debug({tcreate: true})
+    let pReady = super.create()
+    log.debug({tcreateafer: true})
+    this.positioner = new Positioner(this.win)
+    this.win.on('blur', this.onBlur.bind(this))
+    log.debug({tcreateafer2: true})
```
sebn: Remove or clarify debug messages?

gui/src/main/tray.window.js
```diff
@@ -12,114 +12,100 @@ const log = require('cozy-desktop').default.logger({
...
-module.exports.show = (trayPos) => {
-  if (!win) module.exports.create()
-  let pos = null
-  if (trayPos === undefined || trayPos.x === 0) {
-    pos = (process.platform === 'win32') ? 'bottomRight' : 'topRight'
-  } else {
-    pos = (process.platform === 'win32') ? 'trayBottomCenter' : 'trayCenter'
+  create () {
+    log.debug({tcreate: true})
+    let pReady = super.create()
```
sebn: Is this some kind of prefixing convention for js promises?

gui/src/main/window_behaviour.js
```diff
@@ -1,29 +1,104 @@
...
-    app.dock.show()
-    win.on('closed', () => { app.dock.hide() })
+  /* abtract */
+  windowOptions () {
+    throw new Error('extends WindowManager before use')
+  }
+
+  /* abtract */
+  ipcEvents () {
+    throw new Error('extends WindowManager before use')
```
sebn: *extend WindowManager before using*?

gui/src/main/window_behaviour.js
```diff
@@ -1,29 +1,104 @@
...
+  hash () {
+    return ''
+  }
+
+  create () {
+    this.win = new BrowserWindow(this.windowOptions())
+
+    // openExternalLinks
+    this.win.webContents.on('will-navigate', (event, url) => {
+      if (url.startsWith('http') && !url.match('/auth/authorize')) {
```
sebn: Isn't `/auth/authorize` onboarding-specific?

gui/src/main/window_behaviour.js
```diff
@@ -1,29 +1,104 @@
...
+    if (process.platform === 'darwin') {
+      this.app.dock.show()
+      this.win.on('closed', () => { this.app.dock.hide() })
+    }
+
+    // dont keep  hidden windows objects
+    this.win.on('closed', () => { this.win = null })
+
+    let resolveRef = null
+    let pReady = new Promise((resolve, reject) => {
```
sebn: `let resolveCreate = null`? `resolveCreation`?
`let createPromise = ...`? `creationPromise`? `readyPromise`? `windowReadyPromise`? `windowPromise`?

gui/main.js
```diff
@@ -11,9 +11,9 @@ const path = require('path')
 const autoLaunch = require('./src/main/autolaunch')
 const lastFiles = require('./src/main/lastfiles')
 const tray = require('./src/main/tray')
-const trayWindow = require('./src/main/tray.window.js')
-const helpWindow = require('./src/main/help.window.js')
-const onboardingWindow = require('./src/main/onboarding.window.js')
+const TrayWindowWM = require('./src/main/tray.window.js')
```
sebn: Either `TrayWindow` or `TrayWindowManager`? I believe the first one would be ok.

sebn:
@aenario I'm lovin' it 😄 

I would be fine with just adding a few notices to the code. Better merge as soon as possible so we can move on!

gui/src/main/window_behaviour.js
aenario: This is for all other links (`!url.match`)

sebn:
(so glad we fixed 2 weird electron issues in 1 PR)

gui/src/main/window_behaviour.js
sebn: Oh, ok!

gui/src/main/tray.window.js
aenario: Not really a convention, just an habit I picked when dealing with mixed promise/cb codebases

gui/src/main/tray.window.js
aenario: will remove

PR #834, opened 2017-10-09
Actions sorting

aenario:
Simple algorithm to sort events for prep.


cli/src/local/watcher.js
```diff
@@ -355,6 +355,7 @@ class LocalWatcher {
       if (process.env.DEBUG) console.log({actions, e})
     }
 
+    // @TODO: tranform into i<l & j<i with 2-way comparison inside.
```
nono: Is it normal that `i` is not decremented in the loop when `j<i`?

cli/src/local/watcher.js
```diff
@@ -364,16 +365,41 @@ class LocalWatcher {
...
+      for (let j = 0; j < i; j++) {
+        if (process.env.DEBUG) console.log('i, j =', i, j)
+        let a = actions[i]
+        let b = actions[j]
+        if (childOf(addPath(a), addPath(b)) || childOf(delPath(b), delPath(a))) {
+          if (process.env.DEBUG) console.log('switch')
+          // put A before B
+          actions.splice(i, 1)
+          i--
+          actions.splice(j, 0, a)
```
nono: Why don't you just swap `a` and `b`?

```js
actions[i] = b
actions[j] = a
```

PR #847, opened 2017-10-24
Batch and squash remote events

sebn:


sebn:
@aenario Ready to merge `--ff` into `remote-ng`.

PR #851, opened 2017-10-27
Use locks to fix #850

### LINKED ISSUE #850: Duplicated File after move
### Description

Race condition leads to duplicated file

### Steps to Reproduce

1. Create file
2. Move file to sub folder
3. Have bad luck

### Expected behavior

File only in subfolder

### Actual behavior

File appears in both root and subfolder

### Informations

* **reproductible**: sometimes
* **OS**: macOS
* **version**: build-developement-20172610

aenario:


aenario:
[[#850] Duplication d'un fichier suite à un déplacement](https://trello.com/c/BcVfIp4s/429-850-duplication-dun-fichier-suite-%C3%A0-un-d%C3%A9placement)

sebn:
Travis fails. Nice you could reproduce & fix this one. Although I wonder whether we should make a non-regression test... Maybe it should be an integration test ensuring metadata/content push/pull steps don't overlap. But we can save this for later since we need the fix to make more testing. 👍

PR #852, opened 2017-10-27
New Scenario : move dirs with same prefix

aenario:


sebn:
@aenario The eslint issues on first Travis build were supposed to be fixed in development...

PR #856, opened 2017-10-31
Fix sync lock

sebn:
Needs https://github.com/cozy-labs/cozy-desktop/pull/857


sebn:
Random failures as usual...  😭 

🚢 ? `merge --ff` after #857?

PR #857, opened 2017-11-02
Fix local watcher unit tests

sebn:


sebn:
https://travis-ci.org/cozy-labs/cozy-desktop/jobs/296126191#L1945 is known to fail randomly (see comment in test). :ship:? `merge --ff` so we can fixup in `development`?

PR #859, opened 2017-11-02
Stopped client testing

sebn:


cli/test/integration/scenarios.js
```diff
@@ -118,6 +120,53 @@ for (let scenario of scenarios) {
...
   } // event files
 
+  const stoppedTestName = `test/scenarios/${scenario.name}/local/stopped`
+
+  if (scenario.disabled) {
+    it.skip(`${stoppedTestName} (${scenario.disabled})`, () => {})
+  } else {
+    it(stoppedTestName, async function () {
+      this.timeout(3 * 60 * 1000)
+      await Promise.delay(500)
```
aenario: why ?

sebn: To prevent random failures.

- [ ] Add a comment
- [ ] Fix macOS build (broken because logs are too big?)

sebn:
Oops, 1 (random?) failure on macOS: https://travis-ci.org/cozy-labs/cozy-desktop/jobs/297897046#L2862

sebn:
I can't find a way to reproduce the issue locally, which makes me think it's due either to overlapping test events or random event orders.

Possible solutions:
1. Disable stopped client tests on CI
2. Increase delay before each stopped client test
3. Use captures for stopped client tests

Solution 3 would address both concerns but it would take some time to set up, so I'd probably go with solution 1 right now (better disable tests by default and still be able to run them locally when relevant than having a randomly breaking CI).

@aenario WDYT?

sebn:
As discussed, let's merge and see where and how often it fails.

PR #861, opened 2017-11-06
Add code comments & test doc

sebn:


aenario:
The bug you mentionned on last PR is comingback. Let's disable them.

PR #862, opened 2017-11-07
test/scenarios: Fix issue possibly preventing tree/prepCalls assertions

sebn:


sebn:
Let's hope build will still be green...

sebn:
CI should be green now.

PR #865, opened 2017-11-10
Remote already deleted

sebn:


sebn:
@aenario You were not supposed to review this one before next week 😛

PR #874, opened 2017-11-14
gui: Limit oauth window height for small screens

sebn:
Team work with @aenario :)


sebn:
Merge after #875 so #875 doesn't need to be rebased.

PR #875, opened 2017-11-14
Mitigate case & encoding issue

sebn:


sebn:
Reverted to separate unit/integration tests & got rid of all root mocha hooks to prevent random (order-related?) failures.

Tests pass on Linux (Travis), Windows (AppVeyor) and macOS (locally).

@aenario 🚢 ?

PR #878, opened 2017-11-23
Use electron-proxy-agent

aenario:
Tested on windows.

### DIFF
--- a/gui/src/main/proxy.js
+++ b/gui/src/main/proxy.js
@@ -23,16 +23,16 @@ const config = require('yargs')
   .help('help')
   .parse()
 
-log.warn({config}, 'argv')
+log.debug({config}, 'argv')
 
-const printCertificate = (certif) => `Certificate(${certif.issuerName} ${certif.subjectName})`
+const formatCertificate = (certif) => `Certificate(${certif.issuerName} ${certif.subjectName})`
 
 module.exports = (app, session, doneSetup) => {
   const loginByRealm = {}
   if (config['login-by-realm']) {
     config['login-by-realm'].split(',').forEach((lbr) => {
-      const [realm, username, password] = lbr.split(':')
-      loginByRealm[realm] = [username, password]
+      const [realm, username, ...password] = lbr.split(':')
+      loginByRealm[realm] = [username, password.join(':')]
     })
   }
 
@@ -43,9 +43,9 @@ module.exports = (app, session, doneSetup) => {
   session.defaultSession.setCertificateVerifyProc((request, callback) => {
     const {hostname, certificate, verificationResult, errorCode} = request
     if (verificationResult < 0) {
-      log.warn({hostname, certificate: printCertificate(certificate), verificationResult, errorCode}, 'Certificate Verification Error')
+      log.warn({hostname, certificate: formatCertificate(certificate), verificationResult, errorCode}, 'Certificate Verification Error')
     } else {
-      log.debug({hostname, certificate: printCertificate(certificate), verificationResult, errorCode}, 'Certificate Validated')
+      log.debug({hostname, certificate: formatCertificate(certificate), verificationResult, errorCode}, 'Certificate Validated')
     }
     callback(-3) // use chrome validation
   })
@@ -56,12 +56,12 @@ module.exports = (app, session, doneSetup) => {
   })
 
   app.on('certificate-error', (event, webContents, url, error, certificate, callback) => {
-    log.warn({url, error, certificate: printCertificate(certificate)}, 'App Certificate Error')
+    log.warn({url, error, certificate: formatCertificate(certificate)}, 'App Certificate Error')
     callback(false)
   })
 
   app.on('login', (event, webContents, request, authInfo, callback) => {
-    log.warn({request: request.method + ' ' + request.url, authInfo}, 'Login event')
+    log.debug({request: request.method + ' ' + request.url, authInfo}, 'Login event')
     const auth = loginByRealm[authInfo.realm]
     if (auth) {
       event.preventDefault()
--- a/gui/src/main/proxy.js
+++ b/gui/src/main/proxy.js
@@ -1,0 +1,110 @@
+const ElectronProxyAgent = require('electron-proxy-agent')
+const url = require('url')
+const http = require('http')
+const https = require('https')
+
+const log = require('cozy-desktop').default.logger({
+  component: 'GUI:proxy'
+})
+
+
+const config = require('yargs')
+  .env('COZY_DRIVE')
+  .conflicts('proxy-script', 'proxy-rules')
+  .describe('proxy-script', 'The URL associated with the PAC file.')
+  .describe('proxy-rules', 'Rules indicating which proxies to use.')
+  .describe('proxy-bypassrules', 'Rules indicating which URLs should bypass the proxy settings. ' +
+                                 'See https://github.com/electron/electron/blob/master/docs/api/session.md#sessetproxyconfig-callback')
+  .default('proxy-ntlm-domains', '*')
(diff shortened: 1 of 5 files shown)


gui/src/main/proxy.js
```diff
@@ -0,0 +1,107 @@
...
+  .describe('proxy-script', 'The URL associated with the PAC file.')
+  .describe('proxy-rules', 'Rules indicating which proxies to use.')
+  .describe('proxy-bypassrules', 'Rules indicating which URLs should bypass the proxy settings. ' +
+                                 'See https://github.com/electron/electron/blob/master/docs/api/session.md#sessetproxyconfig-callback')
+  .default('proxy-ntlm-domains', '*')
+  .describe('proxy-ntlm-domains', 'A comma-separated list of servers for which integrated authentication is enabled. ' +
+                                  'Dynamically sets whether to always send credentials for HTTP NTLM or Negotiate authentication.')
+  .describe('login-by-realm', 'comma-separated list of realm:user:password')
+  .help('help')
+  .parse()
```
sebn: Maybe move to main later?

gui/src/main/proxy.js
```diff
@@ -0,0 +1,107 @@
...
+
+log.warn({config}, 'argv')
+
+const printCertificate = (certif) => `Certificate(${certif.issuerName} ${certif.subjectName})`
```
sebn: Doesn't really print anything.. `format*`? `stringify`?

gui/src/main/proxy.js
```diff
@@ -0,0 +1,107 @@
...
+  .describe('proxy-bypassrules', 'Rules indicating which URLs should bypass the proxy settings. ' +
+                                 'See https://github.com/electron/electron/blob/master/docs/api/session.md#sessetproxyconfig-callback')
+  .default('proxy-ntlm-domains', '*')
+  .describe('proxy-ntlm-domains', 'A comma-separated list of servers for which integrated authentication is enabled. ' +
+                                  'Dynamically sets whether to always send credentials for HTTP NTLM or Negotiate authentication.')
+  .describe('login-by-realm', 'comma-separated list of realm:user:password')
+  .help('help')
+  .parse()
+
+log.warn({config}, 'argv')
```
sebn: `log.debug`?

gui/src/main/proxy.js
```diff
@@ -0,0 +1,107 @@
+/* eslint standard/no-callback-literal: 0 */
+
+const ElectronProxyAgent = require('electron-proxy-agent')
+const url = require('url')
+const http = require('http')
+const https = require('https')
+
+const log = require('cozy-desktop').default.logger({
+  component: 'GUI:proxy'
```
sebn: Makes me think that CLI won't have proxy support. Not a priority though. We'll just need to make it clear in case we bring it back.

gui/src/main/proxy.js
```diff
@@ -0,0 +1,107 @@
...
+})
+
+const config = require('yargs')
```
sebn: Why not reuse `commander` since it was already used in CLI? Is it obsolete in some way? But same a CLI-related comment above: not a priority (just out of curiosity).

gui/src/main/proxy.js
```diff
@@ -0,0 +1,107 @@
...
+                                  'Dynamically sets whether to always send credentials for HTTP NTLM or Negotiate authentication.')
+  .describe('login-by-realm', 'comma-separated list of realm:user:password')
+  .help('help')
+  .parse()
+
+log.warn({config}, 'argv')
+
+const printCertificate = (certif) => `Certificate(${certif.issuerName} ${certif.subjectName})`
+
+module.exports = (app, session, doneSetup) => {
```
sebn: Do we really export anything?

gui/src/main/proxy.js
```diff
@@ -0,0 +1,107 @@
...
+      const [realm, username, password] = lbr.split(':')
+      loginByRealm[realm] = [username, password]
+    })
+  }
+
+  if (config['proxy-ntlm-domains']) {
+    session.defaultSession.allowNTLMCredentialsForDomains(config['proxy-ntlm-domains'])
+  }
+
+  session.defaultSession.setCertificateVerifyProc((request, callback) => {
```
sebn: Is this for debug only?

gui/src/main/proxy.js
```diff
@@ -0,0 +1,107 @@
...
+  }
+  const _httpsRequest = https.request
+  https.request = function (options, cb) {
+    log.warn(options, 'USING RAW HTTPS REQUEST')
+    if (typeof options === 'string') {
+      options = url.parse(options)
+    } else {
+      options = Object.assign({}, options)
+    }
+    options.agent = options.agent || https.globalAgent
```
sebn: Why doesn't `https.request` need the same `options.headers.host` patching as in `http.request`?

gui/src/main/proxy.js
```diff
@@ -0,0 +1,107 @@
...
+log.warn({config}, 'argv')
+
+const printCertificate = (certif) => `Certificate(${certif.issuerName} ${certif.subjectName})`
+
+module.exports = (app, session, doneSetup) => {
+  const loginByRealm = {}
+  if (config['login-by-realm']) {
+    config['login-by-realm'].split(',').forEach((lbr) => {
+      const [realm, username, password] = lbr.split(':')
+      loginByRealm[realm] = [username, password]
```
sebn: Make sure auth doesn't break when password contains a colon?

```javascript
const [realm, username, passwordParts] = lbr.split(':')
loginByRealm[realm] = [username, passwordParts.join(':')]
```

gui/src/main/proxy.js
```diff
@@ -0,0 +1,107 @@
...
+  })
+
+  app.on('select-client-certificate', (event, webContents, url, list, callback) => {
+    log.debug({url}, 'select-client-certificate')
+    callback()
+  })
+
+  app.on('certificate-error', (event, webContents, url, error, certificate, callback) => {
+    log.warn({url, error, certificate: printCertificate(certificate)}, 'App Certificate Error')
+    callback(false)
```
sebn: Why `false`?

gui/src/main/proxy.js
sebn: (not repeating, there are a few others below)

gui/src/main/proxy.js
aenario: It was in the CLI package and I am scared of the dark voodoo between cli & gui if they use the same package 😃

gui/src/main/proxy.js
aenario: yep, this cant be initialized until `app.on('ready')`

gui/src/main/proxy.js
aenario: yep. Hopefully will give us more data for the bug we were speaking about tonight.

gui/src/main/proxy.js
aenario: Need to callback(false) to not accept any certificate.

gui/src/main/proxy.js
aenario: `https.request` calls `http.request` https://github.com/nodejs/node/blob/edb03cb65d51e336713d6af1134a7394c0776635/lib/https.js#L239

gui/src/main/proxy.js
sebn: Of course, we are exporting a function... Sorry, my bad :)

sebn: (My brain was messing up with requirejs-like syntax...)

gui/src/main/proxy.js
sebn: Ok, thanks for the explanation!

sebn:
👍 👌 🙈 ❤️  🎉 🚢 :shipit: 🚀

PR #896, opened 2017-12-14
Cosy vs cozy  and Quit button

aenario:


gui/src/elm/Address.elm
```diff
@@ -61,6 +61,8 @@ update msg model =
                 setError model "Address You don't have filled the address!"
             else if contains "@" model.address then
                 setError model "Address No email address"
+            else if contains "mycosy.cloud" model.address then
```
sebn: `.mycosy.cloud`? (leading dot)

gui/src/main/onboarding.window.js
```diff
@@ -42,7 +42,12 @@ module.exports = class OnboardingWM extends WindowManager {
 
   onRegisterRemote (event, arg) {
     let desktop = this.desktop
-    const cozyUrl = desktop.checkCozyUrl(arg.cozyUrl)
+    let cozyUrl
+    try {
+      cozyUrl = desktop.checkCozyUrl(arg.cozyUrl)
+    } catch (err) {
+      return event.sender.send('registration-error', translate('Address Invalid address!'))
```
sebn: `return` on next line, so we know it doesn't return anything?

PR #905, opened 2018-01-03
NOFIX : Fix build process to support older linux distros

aenario:


aenario:
Possible solutions : 


Option A: bundling `libstdc++.so` 
- Add `libstdc++.so` to AppImage build : https://github.com/electron-userland/electron-builder/issues/1985
- But AppImage does not recommend it, as it might break on newer distro https://github.com/AppImage/AppImageKit/wiki/Creating-AppImages#libstdcso6
- So AppImage recommends to use https://github.com/darealshinji/AppImageKit-checkrt/ to only include the bundled `.so` if the distro is too old.
- But we need to figure out how to include this within `electron-builder`
- This add a few Mo to the build, but insignificant compare to the whole bundle.

Option B: Building on an older distro should give us a working binary.
But we will need to duplicate the works done on https://github.com/electron-userland/electron-builder/tree/master/docker to get a working build environment.

Optionc C: Give up on older distribution (including Jessie :unamused: ) for now.

To be discused when @sebn comes back.

aenario:
Also an issue with Linux Mint 17.1 Rebecca. cf https://github.com/cozy/cozy-drive/issues/667

sebn:
I suppose we can't support every old stable / LTS distrib out there while we just released our first GNU/Linux version.

But maybe we could at least make it easier for people to support them by providing / improving instructions to build the app from sources? WDYT @aenario @jsilvestre?

PR #916, opened 2018-01-18
Improve onboarding UX

aenario:
image


sebn:
Isn't *cloudy* a bit confusing as an example since it's too close to *cloud*? /cc @jsilvestre

gui/elm/Address.elm
```diff
@@ -48,6 +49,35 @@ setError model message =
...
+ensureNoHttps address =
+    if String.startsWith "https://" address then
+        String.dropLeft 8 address
+    else
+        address
+
+
+correctAddress : String -> String
+correctAddress address =
+    dropAppName (ensureNoHttps address)
```
sebn: *Pipe* operators are great to write things in a more natural way:

```elm
address |> ensureNoHttps |> dropAppName
```

gui/elm/Address.elm
```diff
@@ -56,6 +86,9 @@ update msg model =
         FillAddress address ->
             ( { model | address = address, error = "", busy = False }, Cmd.none )
 
+        CorrectAddress ->
+            ( { model | address = correctAddress model.address }, Cmd.none )
```
sebn: I expected it to be done in *core/app* since I believe there is already some validation in there. Although I definitely prefer the elm implementation :trollface:

gui/elm/Address.elm
```diff
@@ -96,15 +131,25 @@ view helpers model =
...
-                , input
-                    [ placeholder ("cloudy.mycozy.cloud")
-                    , class "wizard__address"
-                    , type_ "text"
-                    , value model.address
-                    , onInput FillAddress
-                    , onEnter RegisterRemote
+                , div [ class "https-input-wrapper" ]
+                    [ span [ class "address_https" ]
+                        [ text ("https://") ]
```
sebn: No need for parenthesis

PR #925, opened 2018-01-19
remote/cozy: Recover when changesfeed dir_id points to invalid doc

sebn:


core/remote/cozy.js
```diff
@@ -103,8 +103,8 @@ export default class RemoteCozy {
         // File docs returned by the cozy-stack don't have a path
         const parent = fileParentsById[doc.dir_id]
 
-        if (parent.error) {
```
aenario: Does the parent.error case really never happens ?

PR #929, opened 2018-01-19
Fix onboarding address

sebn:


sebn:
Already done in #927, my bad, closing...

PR #938, opened 2018-01-22
dev: Introduce setup script & env file

sebn:
To make it easier to test the app without trashing your real data/config.


sebn:
@aenario You'll still have to be careful when selecting the synced dir during onboarding. But we can eventually introduce some `COZY_DESKTOP_SYNCED_DIR=$PWD/tmp/Cozy Drive` environment variable later, that would be used as the default choice. And when doing so, renaming a few vars/properties in `app` (e.g. `baseDir` vs `configDir`) could also be a good idea.

PR #944, opened 2018-01-23
core/sync: Sync events by batch

sebn:
So we can lock/release Pouch for the whole batch.
So moved src/dst docs don't get interleaved with unrelated ones.

We didn't include defensive checks to prevent infinite looping
because it actually may fix an existing bug.

Also temporarily disabled sync/pop implementation related unit tests.


sebn:
Already in master.

PR #949, opened 2018-01-27
Clean up

sebn:


sebn:
Linux & Windows builds are green, macOS build is already broken on master... :ship: ?

PR #959, opened 2018-01-31
help: add app version to logs & mail

aenario:


core/app.js
```diff
@@ -179,6 +180,7 @@ class App {
   // Send an issue by mail to the support
   sendMailToSupport (content: string) {
     const logs = fs.readFileSync(LOG_FILE, 'utf-8')
+    content = JSON.stringify(this.debugInformations()) + '\r\n' + content
```
sebn: Did you give it a try?
Won't the debug line be too near the content? Maybe put 2 '\r\n'?
Won't it be confusing when replying to a user? Maybe put it at the end of the message, with a clear separation and some kind of title?

PR #960, opened 2018-02-01
gui: Ensure synced folder is valid

sebn:


gui/js/onboarding.window.js
```diff
@@ -42,9 +43,11 @@ module.exports = class OnboardingWM extends WindowManager {
     // through ports so we can trigger 'registration-done' without relying
     // on timeouts
     this.send('registration-done')
-    this.win.once('dom-ready', () => {
-      setTimeout(() => this.send('registration-done')
-      , 20)
+    this.win.webContents.once('dom-ready', () => {
+      setTimeout(() => {
+        this.send('registration-done')
+        this.checkSyncPath(defaults.syncPath, this)
```
aenario: Using `this` as a `eventSender` is kind of hacky.

gui/js/onboarding.window.js
```diff
@@ -113,19 +121,29 @@ module.exports = class OnboardingWM extends WindowManager {
...
+    eventSender.send('folder-chosen', {
+      folder: result.syncPath,
+      error: result.error ? `Folder ${result.error}` : null
+    })
+    return result
+  }
+
   onStartSync (event, syncPath) {
+    const {error} = this.checkSyncPath(syncPath, event.sender)
+    if (error) return
```
aenario: `log.warn` it,  just in case ?

gui/js/onboarding.window.js
sebn: Agreed.

I was feeling wrong about passing `this.win.webContents` without first checking it still exists, which `this.send()` already does (well actually `this.win` could be `null`, but `this.win` is not supposed to exist without `this.win.webContents` since `this.win` is a `BrowserWindow`...).

Another solution could be to pass a `send` function...

What were you thinking of?

gui/js/onboarding.window.js
sebn: Makes sense. What about doing so directly in `checkSyncPath`?

aenario: It will log for everyone who attempt to sync an existing folder (before they are given a warning it's not possible), whereas doing it here will only log if the user manage to ignore the error and click "start sync" with a bad folder?.

gui/js/onboarding.window.js
```diff
@@ -113,19 +121,29 @@ module.exports = class OnboardingWM extends WindowManager {
...
-      event.sender.send('folder-chosen', {
-        folder: result.syncPath,
-        error: result.error ? `Folder ${result.error}` : null
-      })
+      this.checkSyncPath(folders[0], event.sender)
     }
   }
 
+  checkSyncPath (syncPath, eventSender) {
+    console.log('checkSyncPath', syncPath, eventSender.send)
```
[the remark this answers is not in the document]
aenario: good self catch ;-)

gui/js/onboarding.window.js
aenario: I'am thinking this is hacky, but still a lot less than the whole dom-ready+timeout that surrounds it. Maybe just a `// XXX` to be considered if we ever get around o fix `elm<->ports<->renderer<->main` pipeline.

gui/js/onboarding.window.js
sebn: :+1:

PR #967, opened 2018-02-02
core: Support email improvements

sebn:
So we can use our own email and stop polluting support in dev.
Defaults to the real one.


sebn:
:ship: as soon as CI is green!

sebn:
Added missing unit tests (should make Codecov happy)

PR #972, opened 2018-02-06
Add more perf probes

aenario:
Not sure if we should merge this into master.


sebn:
Closing after discussion.

PR #988, opened 2018-02-17
Added instructions to setup a development environment on Ubuntu

hleroy:
Tested with Ubuntu 17.10


sebn:
Thanks again @hleroy!
@aenario Does it match your setup?

PR #1004, opened 2018-02-28
Survive when chokidar miscategorized an event

aenario:


sebn:
I have a few concerns... Since we only saw it once, is it really worth fixing? I mean it could hide bugs in our own code... And we should at least log a warning. WDYT?

sebn:
Also, in `LocalWatcher#prepareEvents`, we actually rely on the id(path) since we don't have stats for unlink* events. So assuming chokidar is wrong may not be such a good idea, it could actually hide a conflict case (e.g. replacing a file with a directory). We would at least need a move destination add event to detect a chokidar issue.

core/local/analysis.js
```diff
@@ -68,6 +68,17 @@ function analyseEvents (events: LocalEvent[], pendingChanges: LocalChange[]): Lo
 
   for (let e: LocalEvent of events) {
     try {
+      // chokidar make mistakes
+      if (e.type === 'unlinkDir' && e.old && e.old.docType === 'file') {
+        // $FlowFixMe
+        e.type = 'unlink'
```
sebn: - [ ] log a warning

core/local/analysis.js
```diff
@@ -68,6 +68,17 @@ function analyseEvents (events: LocalEvent[], pendingChanges: LocalChange[]): Lo
...
+      }
+
+      if (e.type === 'unlink' && e.old && e.old.docType === 'folder') {
+        // $FlowFixMe
+        e.type = 'unlinkDir'
```
sebn: - [ ] log a warning

PR #1010, opened 2018-03-05
 core/sync: Actually bring back child file on compat->incompat->compat move

sebn:
Courtesy of @sebn & @aenario


sebn:
Code reported as non-covered by Codecov was actually test-driven.

PR #1017, opened 2018-03-12
jq: Introduce filters & scripts

sebn:
Useful to analyse huge logs.


sebn:
I dedicate this commit to @clochix :smiley:

PR #1022, opened 2018-03-15
all: Replace es6 modules with node ones

sebn:
Another step to get rid of babel.
Moved type imports to the top so eslint doesn't complain.
Not sure why I get errors regarding ReadableWithContentLength type import, though.

### DIFF
--- a/core/app.js
+++ b/core/app.js
@@ -1,33 +1,34 @@
 /* @flow */
 
-import EventEmitter from 'events'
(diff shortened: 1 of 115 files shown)


sebn:
Well Travis didn't catch [this one](https://github.com/cozy-labs/cozy-desktop/blob/master/gui/main.js#L5) :sob:

PR #1023, opened 2018-03-15
core/metadata: Add test & fix to not update file if only mtime has changed

aenario:
Waiting for Travis to see if this breaks anything else.


sebn:
AppVeyor is already broken on master (nothing to do with this PR). Merging.

PR #1026, opened 2018-03-19
core: Do not recompute checksum when file (mtime) has not changed

sebn:


test/support/builders/metadata/file.js
```diff
@@ -4,13 +4,32 @@ import type { Metadata } from '../../../../core/metadata'
...
+    super(pouch)
+    this.fileOpts = {
+      size: 0,
+      md5sum: ''
+    }
+  }
+
+  data (data: string): this {
+    this.fileOpts.size = Buffer.from(data).length
+    // TODO: md5sum
```
sebn: Oops, 6pm commit :)

sebn:
Should be ready now.

test/unit/local/watcher.js
```diff
@@ -50,6 +51,50 @@ describe('LocalWatcher Tests', function () {
...
+        .create()
+
+      await fs.outputFile(changedPath, 'Changed file NEW content')
+      this.prep.addFileAsync = sinon.stub().resolves()
+      sinon.spy(this.watcher, 'checksum')
+
+      try {
+        await this.watcher.start()
+
+        should(this.watcher.checksum.args).deepEqual([[changedFilename]])
```
aenario: [nitpick-style] how about 
```
checksumer = sinon.spy(this.watcher, 'checksum')
should(checksumer.calledOnce && checksumer.calledWith(changedFilename))
````
 ?

File: test/unit/local/watcher.js [hunk 3]
```diff
@@ -50,6 +51,50 @@ describe('LocalWatcher Tests', function () {
...
+        should(await this.pouch.db.get(unchangedDoc._id)).have.properties(unchangedDoc)
+        should(await this.pouch.db.get(changedDoc._id)).have.properties({
+          ...changedDoc
+        })
+      } finally {
+        await this.pouch.db.remove(unchangedDoc)
```
aenario: Shouldn't the cleanpouch helper takes care of that ?

sebn: I believe it's in an `after` block, not in an `afterEach` one, and since I lost quite some time fixing unrelated failures because of coupled tests, I didn't want to break everything again...

test/unit/local/watcher.js
sebn: Isn't current version actually shorter? Although it would definitely make sense for more than 1 assertion...

sebn: Well I agree your solution is actually more readable...

sebn:
Shipping as there's no blocker and I don't really want to deal with another build round...

PR #1028, opened 2018-03-21
Don't allow remotewatcher calls to be buffered

aenario:


test/unit/scheduling.js
```diff
@@ -0,0 +1,72 @@
...
+    const cozy = cozyHelpers.cozy
+    builders = new Builders(cozy, this.pouch)
+
+    helpers = new IntegrationTestHelpers(this.config, this.pouch, cozy)
+    await helpers.local.setupTrash()
+    await helpers.remote.ignorePreviousChanges()
+
+    // remote watcher gets nothings
+    this.sandbox.stub(helpers._remote.watcher, 'pullMany').callsFake(() => {
+      console.log("pullMany called")
```
sebn: Remaining double quotes :wink:

test/unit/scheduling.js
```diff
@@ -0,0 +1,72 @@
...
+
+  after('sandbox restore', function () {
+    this.sandbox.restore()
+  })
+
+  it('does not allow remote watcher to buffer more than one change', async function () {
+    // syncBatch takes too long
+    this.sandbox.stub(helpers._sync, 'syncBatch')
+                .callsFake(() => {
+                  console.log("syncBatch called")
```
sebn: Remaining double quotes :wink: (last one to fix eslint)

sebn:
Couldn't this have been solved on the `RemoteWatcher` side only? E.g. by storing the current pull promise at the instance level, then checking whether it already exists to cancel subsequent pulls? Or am I missing something?

aenario:
This fix itself is only in the remote watcher. 

The first commit and biggest changes (a6f849eab1c53c0e6373dc92c742c039afd24432) is to separate scheduling mechanisms (locks & timeouts) from actual actions (`syncBatch` and `pullMany`), so we can stub actions in testing and focus only on scheduling.

sebn:
:+1: Ok, makes sense. Let's ship since we could maybe do better but it would not be so easy since we already spent quite some time to fix it.

PR #1035, opened 2018-03-27
Fix recents file list

aenario:


aenario:
Leaving as wip, there is still some styles to touch up.

PR #1039, opened 2018-03-27
core/sync: Stop logging already synced seq

sebn:
It always happens, it's normal, and generates a lot of noise in logs.
We still need to figure out a way to prevent infinite loops though.


aenario:
Shouldn't we ssave remote seq instead ?

sebn:
I don't know, I just tried to keep the existing behaviour without thinking more about it...

PR #1044, opened 2018-04-03
core/sync: send transfer-move events for moves

aenario:


core/sync.js
```diff
@@ -280,10 +280,10 @@ class Sync {
       if (from.incompatibilities) {
         await this.doAdd(side, doc)
       } else if (from.childMove) {
+        this.events.emit('transfer-move', clone(doc), clone(from))
```
sebn: Shouldn't we emit only when `assignNewRev` was successful?

sebn: I fixed it to trigger another build.

sebn:
AppVeyor failures have nothing to do with this PR, will fix them in another one.

PR #1062, opened 2018-06-08
gui/elm: Refactoring

sebn:
Summary:

- Group UI components per window (easier to grasp)
- Rename *Wizard* to *Onboarding*
- Rename *page* to *hash* (and keep leading `#` to match js values)
- Have 1 model/update/subs/view per window (less typecasts)
- Move data structures used in multiple places (windows, ports...) to `Data/`

Remaining work:
- Merge other `Msg` / `update` sub-layers into their parent window layer
- Ensure tests layout match implementation layout

Other thoughts:
- Maybe `Icons` & `Util.Keyboard` could move to `View.*`
- Maybe `View.Helpers` could be extracted from `Locale`
- Maybe `Locale` could move to `Data.Locale`
- Note sure of the `Data/` vs `Model/` wording (I followed the elm-spa-example convention)
- Maybe we could have separate `Model/Update/View` modules per window (this may even be needed for view submodules)
- In case we want to be able to display the settings tab on init, we could later introduce some `Route` module wrapping a `#window/tab` hash for example
- Maybe we could have to sections in`Ports` content, e.g. `-- INPUTS` & `-- OUTPUTS`, although having them all together may help spotting name inconsistencies (would be far more useful on the `ports.js` side though, so we could just iterate on js/elm consistant names to bind everything)


aenario:
> Maybe we could have separate Model/Update/View modules per window (this may even be needed for view submodules)

I prefer to have everything in a single module

> In case we want to be able to display the settings tab on init, we could later introduce some Route module wrapping a #window/tab hash for example

I'd rather have a multi-main elm (with 1 main/windows) and keep the hash for tab.

> Maybe we could have to sections inPorts content, e.g. -- INPUTS & -- OUTPUTS, although having them all together may help spotting name inconsistencies (would be far more useful on the ports.js side though, so we could just iterate on js/elm consistant names to bind everything)

I'd rather remove all these steps and replace the the ports.js by something like 
`for k in Object.keys(elm.ports) mainRenderer.on(k, elm.ports[k])`

sebn:
👌 Agreed with your comment

PR #1066, opened 2018-06-12
Handle required user action completion

sebn:


sebn:
@aenario Still struggling to make timeout tests pass...
Also, there is probably a smarter way to handle decreasing ticks (to be improved later).

PR #1071, opened 2018-06-20
chore: Extract automatedRegistration() and test App#removeRemote()

sebn:


sebn:
CI is green => :ship:

sebn:
Regarding coverage, `automatedRegistration()` is actually called on AppVeyor, but not on Travis were coverage is checked. So coverage will improve as soon as we find a way to make it work on Travis too, not just locally & on AppVeyor.

PR #1074, opened 2018-06-22
jq: More filters & documentation [ci skip]

sebn:
- Add stack trace example
- Add debug level filters
- Add Metadata component filter
- Add test-related filters
- Add Pouch component filter


sebn:
`debug` is probably not such a good name for a filter since it's a built-in. But we can change it later.

PR #1078, opened 2018-06-22
core: Handle overlapping editions

sebn:


sebn:
Let's see whether it breaks something else...

sebn:
Not sure why flow didn't complain before... CI should be green now.

sebn:
@aenario Rebased, cleaned history, CI should be green!

aenario:
LGTM when green CI.

PR #1081, opened 2018-06-26
macos: Add favorite item using embedded swift binary

sebn:


sebn:
- [ ] Don't break `yarn build` on other platforms

sebn:
CI was green but binary wasn't embedded into the app... 😭 
Fixing...

PR #1082, opened 2018-06-27
Bring back synced dir in Finder's favorite items on macOS

sebn:


sebn:
🎉 The packaged app finally works!

PR #1083, opened 2018-06-29
Improve logs

aenario:


core/local/watcher.js
```diff
@@ -132,7 +133,7 @@ module.exports = class LocalWatcher {
     const started = new Promise((resolve) => {
       for (let eventType of ['add', 'addDir', 'change', 'unlink', 'unlinkDir']) {
         this.watcher.on(eventType, (path /*: ?string */, stats /*: ?fs.Stats */) => {
-          log.chokidar.debug({path}, eventType)
+          log.chokidar.debug({path, initialScan: !!this.initialScanEvents}, eventType)
```
sebn: Wouldn't initializing `this.initialScan` before the `for` loop and using it for the condition do the trick?
I believe we [are][L183] [already][L222] [using][L258] [it][L280] to decide whether an initial scan is currently running.

[L183]: https://github.com/cozy-labs/cozy-desktop/blob/5a651744793946c56481339e7a2b59194f596957/core/local/watcher.js#L183
[L222]: https://github.com/cozy-labs/cozy-desktop/blob/5a651744793946c56481339e7a2b59194f596957/core/local/watcher.js#L222
[L258]: https://github.com/cozy-labs/cozy-desktop/blob/5a651744793946c56481339e7a2b59194f596957/core/local/watcher.js#L258
[L280]: https://github.com/cozy-labs/cozy-desktop/blob/5a651744793946c56481339e7a2b59194f596957/core/local/watcher.js#L280

core/metadata.js
```diff
@@ -240,16 +238,25 @@ function sameFile (one /*: Metadata */, two /*: Metadata */) {
...
   let fields = ['_id', 'docType', 'md5sum', 'remote._id', 'tags', 'size', 'trashed', 'ino']
   one = _.merge({executable: !!one.executable}, pick(one, fields))
   two = _.merge({executable: !!two.executable}, pick(two, fields))
   const same = isEqual(one, two)
-  if (!same) log.trace({path, diff: {one, two}})
+  if (!same) log.trace({path, one, two, diff: explainDiff(one, two)})
   return same
 }
 
+function explainDiff (one /*: Metadata */, two /*: Metadata */) {
```
sebn: What about having a `diff()` function that returns an object with differences if any or nothing otherwise?
It would ensure there is no differences between the way we explain the comparison and the way we actually do it.
Also we wouldn't have to compare fields twice.

core/local/watcher.js
sebn: Got it: we want the `initialScan` property to be `false` as soon as `onFlush` is called so events occurring during analysis are not logged with `initialScan: true`.

sebn: I'm still a bit uncomfortable with the addition of a new property to the already-bloated `LocalWatcher` class, though...

core/metadata.js
sebn: Since this may be quite a critical change, maybe we should unit test the `core/metadata.same*()` functions before doing so... wdyt?

core/metadata.js
```diff
@@ -1,8 +1,9 @@
 /* @flow */
 
 const _ = require('lodash')
-const { clone, isEqual, pick } = _
+const { clone } = _
 const mime = require('mime')
+const delta = require('deep-diff').diff
```
sebn: Is the trailing `.diff` required? E.g. wouldn't `const delta = require('deep-diff')` work?

Also what about `deepDiff` instead of `delta`, to match the lib name, so we don't introduce another one?

```js
const deepDiff = require('deep-diff')
```

core/metadata.js
```diff
@@ -217,37 +218,49 @@ function isUpToDate (side /*: SideName */, doc /*: Metadata */) {
...
+
 function sameFolder (one /*: Metadata */, two /*: Metadata */) {
   const {path} = two
-  let fields = ['_id', 'docType', 'remote', 'tags', 'trashed', 'ino']
-  one = pick(one, fields)
-  two = pick(two, fields)
-  const same = isEqual(one, two)
-  if (!same) log.trace({path, diff: {one, two}})
-  return same
+  let diff = delta(one, two, sameFolderFilter)
```
sebn: Why not just keep `one` and `two` and `deepDiff()` those? Wouldn't this prevent the need for the various filters?

Or even simpler:

```js
function partialDiff(one, two, keys) {
  one = pick(one, keys)
  two = pick(two, keys)
  return deepDiff(one, two)
}
```

Wouldn't this work?

core/metadata.js
```diff
@@ -217,37 +218,49 @@ function isUpToDate (side /*: SideName */, doc /*: Metadata */) {
...
+  if (!diff || diff.length === 0) return true
```
sebn: Quoting the doc:

> Returns either an array of changes or, if there are no changes, undefined. This was originally chosen so the result would be pass a truthy test [...]

Is it broken?

I wouldn't mind logging a `null` diff, so we don't end up wondering whether there is no change or we missed the log entry:

```js
log.trace({path, diff})
return !diff
```

core/metadata.js
```diff
@@ -217,37 +218,49 @@ function isUpToDate (side /*: SideName */, doc /*: Metadata */) {
...
-  two = _.merge({executable: !!two.executable}, pick(two, fields))
-  const same = isEqual(one, two)
-  if (!same) log.trace({path, diff: {one, two}})
-  return same
+  one = _.merge({executable: !!one.executable}, one)
+  two = _.merge({executable: !!two.executable}, two)
+  let diff = delta(one, two, sameFileFilter)
+  if (!diff || diff.length === 0) return true
+  log.trace({path, diff})
+  return false
```
sebn: Same comments as `sameFolder`

core/metadata.js
```diff
@@ -217,37 +218,49 @@ function isUpToDate (side /*: SideName */, doc /*: Metadata */) {
...
+  let diff = delta(one, two, sameFolderFilter)
+  if (!diff || diff.length === 0) return true
+
+  log.trace({path, diff})
+  return false
 }
 
 // Return true if the metadata of the two files are the same
-// For updated_at, we accept up to 3s of differences because we can't
-// rely on file systems to be precise to the millisecond.
```
sebn: :+1:

core/metadata.js
```diff
@@ -217,37 +218,49 @@ function isUpToDate (side /*: SideName */, doc /*: Metadata */) {
...
 function sameFile (one /*: Metadata */, two /*: Metadata */) {
   if (!sameFileIgnoreRev(one, two)) return false
 
   if ((one.remote && one.remote._rev) === (two.remote && two.remote._rev)) {
     return true
   } else {
-    log.trace({path, diff: {one, two}})
+    log.trace({path, diff: delta({remote: one.remote}, {remote: two.remote})})
```
sebn: Won't this also log `.remote._id` differences?

core/local/watcher.js
```diff
@@ -132,7 +133,8 @@ module.exports = class LocalWatcher {
     const started = new Promise((resolve) => {
       for (let eventType of ['add', 'addDir', 'change', 'unlink', 'unlinkDir']) {
         this.watcher.on(eventType, (path /*: ?string */, stats /*: ?fs.Stats */) => {
-          log.chokidar.debug({path, stats}, eventType)
+          const initialScan = this.initialScan && this.initialScan.flushed
```
sebn: Isn't the condition reversed?

```js
const initialScan = this.initialScan && !this.initialScan.flushed
```

Also, out of the scope of this PR, but what about `const isInitialScan` to prevent confusion with the actual object (my bad)?

core/local/watcher.js
```diff
@@ -170,12 +172,12 @@ module.exports = class LocalWatcher {
   async onFlush (rawEvents /*: ChokidarEvent[] */) {
     log.debug(`Flushed ${rawEvents.length} events`)
 
+    if (this.initialScan) this.initialScan.flushed = false
```
sebn: Should be `true`?

core/metadata.js
aenario: Might be premature optimization, but this is much more efficient and as readable.

core/metadata.js
```diff
@@ -217,37 +218,49 @@ function isUpToDate (side /*: SideName */, doc /*: Metadata */) {
...
   const {path} = two
-  let fields = ['_id', 'docType', 'md5sum', 'remote._id', 'tags', 'size', 'trashed', 'ino']
-  one = _.merge({executable: !!one.executable}, pick(one, fields))
-  two = _.merge({executable: !!two.executable}, pick(two, fields))
-  const same = isEqual(one, two)
-  if (!same) log.trace({path, diff: {one, two}})
-  return same
+  one = _.merge({executable: !!one.executable}, one)
+  two = _.merge({executable: !!two.executable}, two)
+  let diff = deepDiff(one, two, sameFileFilter)
```
sebn: `const`?

core/metadata.js
```diff
@@ -217,37 +218,49 @@ function isUpToDate (side /*: SideName */, doc /*: Metadata */) {
...
+
 function sameFolder (one /*: Metadata */, two /*: Metadata */) {
   const {path} = two
-  let fields = ['_id', 'docType', 'remote', 'tags', 'trashed', 'ino']
-  one = pick(one, fields)
-  two = pick(two, fields)
-  const same = isEqual(one, two)
-  if (!same) log.trace({path, diff: {one, two}})
-  return same
+  let diff = deepDiff(one, two, sameFolderFilter)
```
sebn: `const`?

PR #1088, opened 2018-07-09
core/merge: Keep unapplied move source

sebn:
So we don't end up with doc.moveFrom.moveFrom... in Pouch


sebn:
Closing in favor of #1090

PR #1089, opened 2018-07-11
core: Fix 422 Invalid time given on remote sync

sebn:


sebn:
Same as #1087 with cleaned-up history and unit tests for `metadata.assignMaxDate()`

PR #1090, opened 2018-07-11
core: Extract move model & fix nested moveFrom props

sebn:


sebn:
@nono I'm interested in your feedback regarding the `move()` function extraction. Don't feel the need to look at the `moveFrom` & `childMove` stuff (unless you want to!).

sebn:
@nono Sorry, please wait until I split the first commit for easier review.

core/move.js
```diff
@@ -0,0 +1,47 @@
...
+// Modify the given src/dst docs so they can be merged then moved accordingly
+// during sync.
+function move (src /*: Metadata */, dst /*: Metadata */) {
+  // moveTo is used for comparison. It's safer to take _id
+  // than path for this case, as explained in doc/developer/design.md
+  src.moveTo = dst._id
+  src._deleted = true
+
+  // Make sure newly moved docs have their fill of sync attempts
+  delete src.errors
```
sebn: This was done in `moveFile` but not in `moveFolder`, not sure why... But I assume counting errors on a doc to be deleted doesn't make sense, so better remove it.

core/move.js
```diff
@@ -0,0 +1,47 @@
...
+
+// Modify the given src/dst docs so they can be merged then moved accordingly
+// during sync.
+function move (src /*: Metadata */, dst /*: Metadata */) {
+  // moveTo is used for comparison. It's safer to take _id
+  // than path for this case, as explained in doc/developer/design.md
+  src.moveTo = dst._id
+  src._deleted = true
+
+  // Make sure newly moved docs have their fill of sync attempts
```
sebn: This comment should actually be for `delete dst.errors`

core/move.js
```diff
@@ -0,0 +1,47 @@
...
+  delete src.errors
+  delete dst.errors
+
+  // TODO: Find out wether or not it would make sense to also delete the
+  // trashed property on the source, or explain why it doesn't.
+  delete dst.trashed
+
+  dst.moveFrom = _.omit(src, 'moveFrom')
```
sebn: `moveTo` should only be set on deleted docs, so it shouldn't be set anyway.
Maybe we could omit `errors` too.

@aenario The more I think about it, the more I believe we should actually remove any non-useful property to prevent confusion. And introduce some corresponding `MoveSource` type. We're mostly need `path`, `incompatibilities`. For some reason we're also using `remove._id` and `remote._rev` in `side.move*()` but I assume we could set those on the destination doc and use them from there... This is for a future PR though.

PR #1092, opened 2018-07-16
Don't detect incompatible docs as unlinked on initial scan

sebn:
So they don't end up in the Cozy trash.


sebn:
Rebased on master. Merging.

PR #1093, opened 2018-07-17
Assert metadata is deleted when remote incompatible doc is trashed

sebn:


sebn:
I made sure that this specific assertion was actually failing when doc was not trashed.

sebn:
@aenario :ship: ?

PR #1103, opened 2018-07-20
Move and replace dir

aenario:


sebn:
(sorry for the late review)

👍 for making it work!

I restarted AppVeyor and it seems to be green now.

PR is quite big though. Maybe we could extract & squash a few PR out of it, like:
- Pouch error swallowing
- One-side scenarios
- Logs improvements
- New capture script

Also I couldn't find a way to reproduce the Pouch error swallowing issue in a test... Do you remember in which case we were getting it?

sebn:
Here is my (failing) attempt regarding `Pouch#bulkDocs()`:

https://github.com/cozy-labs/cozy-desktop/compare/2b643219...pouch-bulk-docs

test/scenarios/move_overwriting_dir/scenario.js
```diff
@@ -0,0 +1,53 @@
...
+      {method: 'trashFolderAsync', path: 'dst/dir'},
+      {method: 'moveFileAsync', dst: 'dst/dir/file', src: 'src/dir/file'},
+      {method: 'moveFileAsync', dst: 'dst/dir/file2', src: 'src/dir/file2'},
+      {method: 'moveFileAsync', dst: 'dst/dir/subdir/file6', src: 'src/dir/subdir/file6'}
+    ],
+    tree: [
+      'dst/',
+      'dst/dir/',
+      'dst/dir/file',
+      'dst/dir/file2',
```
sebn: I assume `file3` is not there because it's an *overwrite*, not a *merge*?
Maybe we could rename some files to `file-to-be-deleted` or `to-be-deleted-by-dir-overwrite` without specifying their content?
So their name would just appear in the diff in case we messed up and it would make it easier to understand wether they should be there or not.

test/scenarios/move_overwriting_dir/scenario.js
```diff
@@ -0,0 +1,53 @@
...
+      'dst/dir/subdir/',
+      'dst/dir/subdir/file',
+      'dst/dir/subdir/file6',
+      'src/'
+    ],
+    // remoteTrash: [
+    //   'file',
+    //   'file (__cozy__:'
+    // ],
```
sebn: Remove dead code? It would be nice if we could ellipsize trash suffixes same as we're doing with conflict suffixes in integration tests (maybe add it to future improvements?)

test/scenarios/move_overwriting_dir/scenario.js
```diff
@@ -0,0 +1,53 @@
...
+    { ino: 2, path: 'dst/dir/' },
+    { ino: 3, path: 'dst/dir/subdir/' },
+    { ino: 4, path: 'dst/dir/file', content: 'overwriten' },
+    { ino: 5, path: 'dst/dir/file3', content: 'not-overwriten' },
+    { ino: 6, path: 'dst/dir/subdir/file', content: 'sub-overwriten' },
+    { ino: 7, path: 'dst/dir/subdir/file5', content: 'sub-not-overwriten' },
+    { ino: 8, path: 'src/' },
+    { ino: 9, path: 'src/dir/' },
+    { ino: 10, path: 'src/dir/subdir/' },
+    { ino: 11, path: 'src/dir/file', content: 'overwriter' },
```
sebn: `overwriter` & `overwriten` are hard to distinguish (should actually be `overwritten`).
What about `old file content` & `new file content`?

PR #1111, opened 2018-07-30
test/world: FS unicode normalization examples

sebn:


sebn:
Passes locally on High Sierra with APFS:

test output

Should be green on CI.

PR #1112, opened 2018-07-30
core/metadata: Rename .assignIdHFS() to assignIdApfsOrHfs()

sebn:


core/metadata.js
```diff
@@ -65,12 +65,14 @@ export type Metadata = {
 
 let assignId /*: (doc: *) => void */ = (_) => {}
 
+// See [test/world/](https://github.com/cozy-labs/cozy-desktop/blob/master/test/world/)
+// for file system behavior examples.
```
sebn: The relevant tests are actually included in #1111

core/metadata.js
```diff
@@ -110,16 +112,27 @@ function assignIdUnix (doc /*: * */) {
...
+// Build an _id from the path for macOS, assuming file system is either APFS
+// or HFS+.
+//
+// APFS:
+// - case preservative, but not case sensitive
+// - unicode normalization preservative, but not sensitive
+//
+// HFS+:
 // - case preservative, but not case sensitive
 // - unicode NFD normalization (sort of)
```
sebn: `// - unicode normalization insensitive, converts to NFD (sort of) automatically`

wdyt?

PR #1113, opened 2018-07-30
pouch: prevent pouch bulkDocs from swallowing errors

aenario:

- [x] PR is not too big

### DIFF
--- a/core/pouch.js
+++ b/core/pouch.js
@@ -112,13 +112,24 @@ class Pouch {
     return this.db.put(doc).asCallback(callback)
   }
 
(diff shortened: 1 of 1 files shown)


sebn:
I'm still uncomfortable with this one since I couldn't find a way to make a test fail without the fix: the method seems to properly reject without saving anything in case one doc has an invalid rev for example, be it the first or the last doc. Could it be that we're doing something wrong regarding rejection catching? In Sync? Some watcher? WDYT?

sebn:
>  it includes tests matching the implementation changes

Well, not really 😛

sebn:
Closed in favor of #1129

PR #1119, opened 2018-07-31
test/integration/platform_incompatibilities: Fix code style

sebn:
- Return as soon as possible
- No useless indent


sebn:
Tests pass locally, I only dropped the `else` branch, unindented its content and moved the remaining `if` block to the top.

sebn:
(will make some future PR easier to review)

PR #1120, opened 2018-07-31
test: Add & use #incompatibleTree() helper

sebn:


sebn:
`posixifyPath` should probably be moved somewhere else, but I'm lacking inspiration for now...

PR #1121, opened 2018-07-31
test/world: Fix encoding tests on APFS

sebn:


sebn:
AppVeyor doesn't matter: it's a macOS-only change. 🚢 ?

PR #1122, opened 2018-07-31
jq: Filters improvements

sebn:


sebn:
Forgot to make a PR for those, my bad.

PR #1133, opened 2018-08-16
core: Stop abusing writers return types

sebn:
So we may stop wondering:
- wether it is actually used or not
- which doc we should rely upon

All writer methods are actually mutating the input doc.
This is what Sync actually uses currently, while some tests were relying
on writer method results, which was wrong.

Now everybody uses the mutated doc (until we rewrite all this using some
functional pattern).


sebn:
To be squashed & merged

PR #1134, opened 2018-08-20
WIP: Handle identity conflicts

sebn:
Please make sure the following boxes are checked:

- [ ] PR is not too big
- [ ] it improves UX & DX in some way
- [ ] it includes tests matching the implementation changes
- [ ] it includes relevant documentation

PR #1138, opened 2018-08-21
core: Fix type annotations

sebn:


sebn:
Was required by #1134

PR #1139, opened 2018-08-21
core: Extract Pouch#byIdMaybe() from Merge

sebn:


core/merge.js
```diff
@@ -503,16 +464,8 @@ class Merge {
...
-    try {
-      folder = await this.pouch.db.get(doc._id)
-    } catch (err) {
-      if (err.status === 404) {
-        return null
-      } else {
-        throw err
-      }
-    }
+    const folder /*: ?Metadata */ = await this.pouch.db.get(doc._id)
```
aenario: Shouldn't this be `byIdMaybeAsync` ?

sebn: - [ ] Definitely!

sebn:
Fixed call & rebased on master. Waiting for CI.

@nono Can you think of some issue if we decide to throw on non-404 Pouch errors instead of logging a warning in Merge?

nono:
Well, I don't see what errors can return PouchDB on a get (except 404). So, it's hard to tell.

sebn:
I assume any other error would mean the system is currently broken in some way, so better throw instead of still trying to sync.

PR #1143, opened 2018-08-31
chore: Fix test script

enguerran:
I would like to add a commit that changes the tests scripts organization.

`npm run test` runs every tests, at the time I create this Pull Request, it takes more than 15 minutes.
But this classical standard script `npm run test` or `yarn test` is often run as soon as we clone the repository.
So I would suggest to change:
- `"test": "yarn test:world && yarn test:unit && yarn test:elm && yarn test:integration && yarn test:scenarios"` → `"test:all": "yarn test:world && yarn test:unit && yarn test:elm && yarn test:integration && yarn test:scenarios"`
- `"test": "yarn test:world && yarn test:unit && yarn test:elm && yarn test:integration && yarn test:scenarios"` → `"test": "yarn test:unit"`

It would include that we change the `.travis.yml` scripts.

---

- [X] PR is not too big
- [X] it improves UX & DX in some way
- [X] it includes tests matching the implementation changes
- [X] it includes relevant documentation

### DIFF
--- a/package.json
+++ b/package.json
@@ -62,7 +62,7 @@
     "repl": "env-cmd .env.dev node ./dev/repl.js",
     "stack:reset": "sudo ./node_modules/.bin/rimraf tmp/couchdb tmp/cozy-storage",
     "start": "env-cmd .env.dev electron .",
-    "test": "yarn test:world && test:unit && yarn test:elm && yarn test:integration && yarn test:scenarios",
+    "test": "yarn test:world && yarn test:unit && yarn test:elm && yarn test:integration && yarn test:scenarios",
     "test:coverage": "yarn test:unit:coverage && yarn test:integration",
     "test:elm": "elm-test",
     "test:integration": "yarn mocha test/integration",


aenario:
:+1: sur le split test / test:all 

I would put the `test:integration` in `test`, if I recall correctly they are pretty fast. `test:scenarios` are the ones that take forever!

enguerran:
Okay, I'll add the change about the scripts.

sebn:
I would also include `yarn test:elm` because those are Elm unit tests actually, and `yarn test:world` because they are shorter and verify critical assumptions about the current platform.

Regarding `yarn test:integration`, please note it can still take from 30s to 1min depending on the platform. I have not strong preference between including it or not.

enguerran:
Ok, this is very discouraging. I just saw that `.travis.yml` lists the tests to run on the CI. Potentially nobody ever uses the `yarn test` script alone, so let's drop it for now, we'll see later.

sebn:
:ok_hand: Fixing the typo is still useful.

Regarding `.travis.yml`, maybe we did this to still get the output of the various tests when one was failing, I'm not sure...

A solution that was discussed at some point was to have separate builds to make it easier to know what was failing, but we didn't take the time to do anything about it yet.

PR #1144, opened 2018-09-03
core/pouch: add checks for invariants

aenario:

- [x] PR is not too big

### DIFF
--- a/core/metadata.js
+++ b/core/metadata.js
@@ -90,6 +90,7 @@ module.exports = {
   isFile,
   id,
   invalidPath,
+  invariants,
   ensureValidPath,
   detectPlatformIncompatibilities,
   invalidChecksum,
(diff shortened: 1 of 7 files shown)


core/metadata.js
```diff
@@ -178,6 +179,12 @@ function ensureValidPath (doc /*: PathObject */) {
   }
 }
 
+function invariants (doc /*: Metadata */) {
```
sebn: Rename to `validateBeforeSave` or `validateForSaving`, since `Metadata` can still exist without `sides` before `Prep`, the `invariants` name could be confusing?

core/metadata.js
```diff
@@ -178,6 +179,12 @@ function ensureValidPath (doc /*: PathObject */) {
...
+  if (doc.sides.remote && !doc.remote) {
```
sebn: Also test ```if (!doc.sides) throw new Error(`${doc._id} has no sides`)``` ?

PR #1147, opened 2018-09-04
test/scenarios: Improve remote->local assertions

sebn:


sebn:
Although the instrumentation code is not better, the failure output is, definitely.
Instrumentation code will be improved in another PR.

PR #1153, opened 2018-09-07
core/local: Fix moves

sebn:
Stop doing magic like we did in v2, this could mess up with connectors &
sharings:

- Don't delete src when dst exists
- Let it crash on missing src
- Don't create missing dst parent
- Don't fallback to dst creation

Still allow identical rename.

### DIFF
--- a/core/local/index.js
+++ b/core/local/index.js
@@ -283,64 +283,54 @@ module.exports = class Local /*:: implements Side */ {
     log.info({path: doc.path}, 'Local assignNewRev = noop')
   }
 
-  // Move a file from one place to another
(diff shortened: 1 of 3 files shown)


sebn:
Remaining work:

- [ ] Fix move+overwrite
- [ ] Clean up history

sebn:
Let's see wether CI is green when rebased on #1160

sebn:
Build passes except packaging fails complaining about the content of `$TRAVIS_COMMIT_MESSAGE`.

sebn:
:ship:?

PR #1154, opened 2018-09-10
WIP: Remote analysis

sebn:


sebn:
The goal was to extract and remove all side-effects from the remote changes analysis function, to make it easier to test/change.

2nd commit is breaking flow unexpectedly (unit tests were expected to fail though since I just didn't move them yet). I'll try to quickly figure out why, falling back to a revert to the 1st commit.

sebn:
Random `Merge#moveFolder()` test failure fixing attempt in #1155

core/pouch.js
```diff
@@ -228,6 +228,12 @@ class Pouch {
     })
   }
 
+  async allByRemoteIds (remoteIds /*: * */) /* Promise<Metadata[]> */ {
+    const params = {keys: Array.from(remoteIds), include_docs: true}
```
aenario: mmmh, may be safer to split the array into a constant maximum batch size. But this can take place when we limit the change feed. Add a TODO ?

PR #1159, opened 2018-09-11
Fix executable flag handling on windows

aenario:
Ready to merge

### DIFF
--- a/core/metadata.js
+++ b/core/metadata.js
@@ -276,7 +276,7 @@ const makeComparator = (name, interestingFields) => {
   }
   return (one, two) => {
     const diff = deepDiff(one, two, filter)
-    log.trace({path: two.path, diff}, 'sameFolder')
+    log.trace({path: two.path, diff}, name)
     return !diff
   }
 }
--- a/core/metadata.js
+++ b/core/metadata.js
@@ -47,7 +47,7 @@ export type Metadata = {
   class?: string,
   docType: string,
   errors?: number,
-  executable?: true,
+  executable?: true|'na',
   updated_at: string|Date,
   mime?: string,
   moveTo?: string, // Destination id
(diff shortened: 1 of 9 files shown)


test/integration/update.js
```diff
@@ -81,6 +81,45 @@ describe('Update file', () => {
     })
   })
 
+  describe.only('executable change', () => {
```
sebn: oops

sebn:
Except for the exclusive test above, shouldn't we test more cases?
(I'm a bit uncomfortable trying to think about all the possible local / remote executable / not-executable / whatever file addition / permission update scenarios...)

test/integration/update.js
aenario: Fixed, added a commit to avoid repeting this mistake in the future : https://github.com/cozy-labs/cozy-desktop/pull/1159/commits/5a1caa66f8b86baa07808d612d4103f3d642812d

core/metadata.js
```diff
@@ -188,6 +189,18 @@ function invariants (doc /*: Metadata */) {
   }
 }
 
+function copyIfNull (doc /*: Metadata */, fields /*: Array<string> */, was /*: Metadata */) {
```
sebn: I even wonder whether we should actually list fields by hand instead of generalizing the logic inside this function.
E.g.:

- If no was, return
- Copy missing tags/remote/ino fields from was to doc (with default [] for tags)
- If doc is a file, copy missing executable field from was (honoring 'na')
  * if sameBinary, copy missing size, class & mime from was

And just call it at the beginning of every merge method...

sebn:
Maybe we should split the PR?

I mean:
- the `core/metadata: Fix comparator logging` and `Avoid mocha .only making it to master` commits should definitely ship right now, even directly on master.
- the merge refactoring & executable handling are both substential changes

core/metadata.js
sebn: We could maybe even include other fields this way (`_rev`, `moveFrom`...).

PR #1160, opened 2018-09-11
core/remote/watcher: Fix overwriting file moves

sebn:


sebn:
If I remember well, when we implemented overwriting moves, we handled those cases on the local side, but on the remote one it seemed to be working.

This was because we're still doing some v2 black magic in the LocalWatcher to handle missing/existing move source/destination:
https://github.com/cozy-labs/cozy-desktop/blob/master/core/local/index.js#L294-L311
(to be fixed in an upcoming PR)

And since we don't have any notion of identity (only paths) in scenario expectations, the issue could not be spotted (the resulting tree was fine, except we had possibly broken some references).

sebn:
- [ ] Fix `lint:flow`

sebn:
Travis is broken (this has nothing to do with the current PR). But AppVeyor is green and all tests pass locally on GNU/Linux & macOS.

Let's merge.

PR #1162, opened 2018-09-11
core/remote/cozy: #trashById() -> return RemoteDoc

sebn:


aenario:
LGTM, let's wait for travis  !

PR #1166, opened 2018-09-12
ci/travis/linux: Fix docker --env-file error

sebn:
So we stop wondering about multiline env vars occasionnally breaking the build.
I can't think of any multiline one that would be useful anyway... (please correct me if I'm wrong).
Courtesy of @aenario :)


sebn:
:raised_hands: it didn't crash!
https://travis-ci.org/cozy-labs/cozy-desktop/jobs/427732560#L2211-L2221

PR #1177, opened 2018-09-19
Sentry

aenario:


core/sentry.js
```diff
@@ -0,0 +1,112 @@
...
+
+    isRavenConfigured = true
+    log.info('Raven configured !')
+  } catch (err) {
+    log.error({err}, 'Could not load Raven, errors will not be sent to Sentry')
+  }
+}
+
+const handleBunyanMessage = (msg) => {
+  const level = msg.level >= 50 ? 'error' : msg.level === 40 ? 'warning' : 'info'
```
sebn: Bunyan has a few constants for those:

```js
const bunyan = require('bunyan')
bunyan.ERROR // 50
bunyan.WARN // 50
bunyan.INFO // 50
```

core/sentry.js
```diff
@@ -0,0 +1,112 @@
...
+  if (msg.sentry || (msg.err && msg.err.sentry)) {
+    if (msg.err) {
+      const extra = omit(msg, 'err', 'tags')
+      Raven.captureException(bunyanErrToSentry(msg.err), { extra, level })
+    } else {
+      const extra = omit(msg, 'msg', 'tags')
+      Raven.captureMessage(msg.msg, { extra, level })
+    }
+  } else { // keep it as breadcrumb
+    Raven.captureBreadcrumb({
```
sebn: Does breadcrumbs really help even without wrapping with `Raven.context(() => { ... })`?
Or should we wrap a few key parts with it? (e.g. `RemoteWatcher#pull*()`, `LocalWatcher#onFlush(...)`, `Sync#apply(...)`?

core/sentry.js
```diff
@@ -0,0 +1,112 @@
+/* @flow */
+
+const Raven = require('raven')
+const logger = require('./logger')
+const log = logger({
+  component: 'Sentry'
+})
+
+const { omit } = require('lodash')
```
sebn: `const _ = require('lodash')`? (less changes in future patches in case we use more functions)

core/sentry.js
```diff
@@ -0,0 +1,112 @@
...
+
+const DOMAIN_REGEXP = /^https?:\/\/([a-zA-Z0-9-.]+)(?::\d{2,5})?\/?$/
+const SENTRY_REF = `ed6d0a175d504ead84851717b9bdb72e:324375dbe2ae4bbf8c212ae4eaf26289`
```
sebn: Are Cozy webapps refs embedded like this too? I mean, is there a way to prevent anybody from spamming our Sentry without being identified easily?

sebn:
Also, do you believe code is sufficiently covered by existing unit tests so we don't need new ones? (I have no idea)

package.json
```diff
@@ -103,6 +103,7 @@
     "opn": "5.0.0",
     "pouchdb": "^6.2.0",
     "pouchdb-find": "^6.2.0",
+    "raven": "^2.6.4",
```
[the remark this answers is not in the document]
aenario: Essayons!

core/sentry.js
aenario: We do everything sequentially, I don't think we need concept.

core/sentry.js
aenario: Yes and No

core/sentry.js
```diff
@@ -0,0 +1,112 @@
...
+  }
+}
+
+// TODO: make Flow happy with extended error type
+function flag (err/*: Object */) {
+  err.sentry = true
+  return err
+}
+
+if (!process.env.DEBUG && !process.env.TESTDEBUG && !process.env.COZY_NO_SENTRY) {
```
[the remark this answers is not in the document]
aenario: `.env.dev` already have DEBUG=1

core/sentry.js
aenario: I would apply the rule of 3 (if we start using more than 2 functions from lodash, better to import the whole module) but it's true I have a weird illogical dislike of the `_.xxxxx` syntax

core/sentry.js
```diff
@@ -0,0 +1,112 @@
...
+  } else { // keep it as breadcrumb
+    Raven.captureBreadcrumb({
+      msg: msg.msg,
+      category: msg.component,
+      data: omit(msg, ['component', 'pid', 'name', 'hostname', 'level', 'time', 'v', 'msg'])
+    })
+  }
+}
+
+// TODO: make Flow happy with extended error type
```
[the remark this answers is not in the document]
aenario: Could not get it to work without putting a `// $FlowFixMe` on every call site

core/sentry.js
```diff
@@ -0,0 +1,112 @@
...
+const fatalErrorHandler = (err, sendErr, eventId) => {
+  if (sendErr) log.error({err, sendErr}, 'Fatal error, unable to send to sentry')
+  else log.error({err, eventId}, 'Fatal error, sent to sentry')
+  process.exit(1) // eslint-disable-line no-process-exit
+}
+
+const bunyanErrToSentry = (data) => {
+  if (data instanceof Error) return data
+  // TODO: make Flow happy with extended error type
+  const error /*: Object */ = new Error(data.message)
```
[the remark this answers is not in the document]
aenario: Sentry does a ton of magic when the object is not actually an `Error`

https://github.com/getsentry/sentry-javascript/blob/17481d6e25cd76e39d2447a1a968e8d933a410ae/packages/node/src/backend.ts#L33
https://github.com/getsentry/sentry-javascript/blob/17481d6e25cd76e39d2447a1a968e8d933a410ae/packages/utils/src/is.ts#L8

core/sentry.js
```diff
@@ -0,0 +1,112 @@
...
+  error.name = data.name
+  error.stack = data.stack
```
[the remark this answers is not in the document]
aenario: Bunyan apparently has already transformed the error instance into a POJO at this level.

core/sentry.js
```diff
@@ -0,0 +1,112 @@
...
+})
+
+const { omit } = require('lodash')
+
+const DOMAIN_REGEXP = /^https?:\/\/([a-zA-Z0-9-.]+)(?::\d{2,5})?\/?$/
+const SENTRY_REF = `ed6d0a175d504ead84851717b9bdb72e:324375dbe2ae4bbf8c212ae4eaf26289`
+const SENTRY_DSN = `https://${SENTRY_REF}@sentry.cozycloud.cc/91`
+const DOMAIN_TO_ENV = {
+  'cozy.tools': 'development',
+  'cozy.works': 'development',
```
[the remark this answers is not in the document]
aenario: YAGNI

core/sentry.js
```diff
@@ -0,0 +1,112 @@
...
+  try {
+    const { appVersion, cozyUrl } = clientInfos
+    const { domain, instance, environment } = toSentryContext(cozyUrl)
+    Raven.config(SENTRY_DSN, {
+      release: appVersion,
+      environment,
+      tags: { domain, instance }
+    }).install(fatalErrorHandler)
+
+    isRavenConfigured = true
```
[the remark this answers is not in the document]
aenario: Copied this without thinking from `konnector/libs`

core/sentry.js
sebn: :ok_hand:

core/sentry.js
sebn: :ok_hand:

core/sentry.js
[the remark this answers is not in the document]
sebn: Oh, I misread the condition.

core/sentry.js
sebn: Prefixing functions with the module/namespace they're coming from is considered good practice in at least 3 languages I know of: Go, Elm and Clojure.
So I was assuming it would be a good idea to do the same in JS :P

sebn: Knowing lodash myself, I would have guessed an unprefixed `omit(...)` function call would be lodash-related. But knowing is always more comfortable than guessing :see_no_evil:

package.json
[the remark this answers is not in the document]
aenario: So the answer is HELL NO!

PR #1179, opened 2018-09-21
test/builders: Improvements

sebn:

- [ ] PR is not too big


core/metadata.js
```diff
@@ -247,6 +248,15 @@ function isUpToDate (side /*: SideName */, doc /*: Metadata */) {
   return currentRev === lastRev
 }
 
+function markAsUpToDate (doc /*: Metadata */) {
+  let rev = extractRevNumber(doc) + 1
+  for (let s of ['local', 'remote']) {
+    doc.sides[s] = rev
+  }
+  delete doc.errors
+  return rev
```
sebn: I was hoping to reuse this new function in test builders, but looks like it won't be so easy.
Still I kept it in the PR because it was working.

PR #1180, opened 2018-09-25
core: Add md5sum invariant

aenario:


sebn:
- [ ] flag invariant for sentry

PR #1186, opened 2018-10-02
Scenario loop back

aenario:


core/merge.js
```diff
@@ -331,26 +331,25 @@ class Merge {
...
-      docs = await this.pouch.byRecursivePathAsync(was._id)
-    } catch (err) {
-      throw err
-    }
+    const docs = await this.pouch.byRecursivePathAsync(was._id)
     move(was, folder)
     let bulk = [was, folder]
-    for (let doc of Array.from(docs)) {
+
+    const modifiedIDs = docs.map((doc) => doc._id.replace(was._id, folder._id))
```
sebn: `(doc) => doc._id.replace(was._id, folder._id)` logic duplicates line 345 (now 346).

Maybe extract a function? e.g.

```js
const destinationId = doc => doc._id.replace(was._id, folder._id)
const destinationIds = docs.map(destinationId)
...
for (let doc of docs) {
  ...
  dst._id = destinationId(doc)
```

core/merge.js
```diff
@@ -331,26 +331,25 @@ class Merge {
...
+    const result = await this.pouch.db.allDocs({keys: modifiedIDs})
+    const existingsDestinations = {}
+    for (let r of result.rows) if (r.value) existingsDestinations[r.key] = r.value.rev
```
sebn: Maybe rename const, extract revs retrieval to Pouch and prevent PLOP? e.g.

```js
// Pouch

async allRevsByIds (ids /*: string */) /*: Promise<Metadata[]> */ {
  const results = await this.db.allDocs({keys: ids})
  return _.chain(results).filter('value').map(({value}) => [value.id, value.rev]).fromPairs().value()
}

// Merge
const existingDestinationRevs = await this.pouch.allRevsByIds(destinationIds)
```

(this is textarea driven development, commit & deploy without fear :sunglasses:)

sebn: My bad, return type should be `Promise<{[string]: string}>`

sebn: Bonus point if we alias `type Rev = string` & `type Id = string` for documentation purpose :rainbow: 
So we could write `Promise<{[Id]: Rev}>`

PR #1188, opened 2018-10-05
WIP: core/local: Refactor analysis

sebn:


core/local/analysis.js
```diff
@@ -36,34 +36,12 @@ module.exports = function analysis (events /*: LocalEvent[] */, pendingChanges /
...
 
-const panic = (context, description) => {
-  log.error(_.merge({sentry: true}, context), description)
-  throw new Error(description)
-}
-
 function analyseEvents (events /*: LocalEvent[] */, pendingChanges /*: LocalChange[] */) /*: LocalChange[] */ {
   const stopMeasure = measureTime('LocalWatcher#analyseEvents')
-  // OPTIMIZE: new Array(events.length)
-  const changes /*: LocalChange[] */ = []
```
sebn: I wonder why we were listing all changes separately... Isn't it impossible to end up with multiple changes for the same inode? Would it even be possible to have multiple changes for the same (destination) path?

core/local/ChangeMap.js
```diff
@@ -0,0 +1,65 @@
...
+  const changes /*: LocalChangeMap */ = empty()
+  for (const c of pendingChanges) { put(changes, c) }
+  return changes
+}
+
+function put (changes /*: LocalChangeMap */, c /*: LocalChange */) {
+  const previousChange /*: ?LocalChange */ = typeof c.ino === 'number'
+    ? changes.byInode.get(c.ino)
+    : null
+  // FIXME: Is it really needed?
```
sebn: See question on deleted code.

core/local/ChangeMap.js
```diff
@@ -0,0 +1,65 @@
...
+  if (previousChange && previousChange.path !== c.path) {
+    changes.byPath.delete(previousChange.path)
+    // FIXME: Is there anything to remove from changes.withoutInode?
```
sebn: See question on deleted code.

core/local/ChangeMap.js
```diff
@@ -0,0 +1,65 @@
...
+  byPath: Map<string, LocalChange>,
+  withoutInode: LocalChange[]
+}
+*/
+
+function empty () /*: LocalChangeMap */ {
+  return {
+    byInode: new Map(),
+    byPath: new Map(),
+    withoutInode: [] // FIXME: Is it really needed?
```
sebn: See question on deleted code.

core/local/change.js
```diff
@@ -129,13 +90,66 @@ function isChildAdd (a /*: LocalChange */, b /*: LocalChange */) { return childO
...
+        fileMoveFromTo(sameInodeChange, e) ||
+        fileMoveSuccessiveTo(sameInodeChange, e) ||
+        fileMoveIdenticalOffline(e) ||
+        fromEvent(e)
+      )
+    case 'addDir':
+      return (
+        dirMoveSuccessiveTo(sameInodeChange, e) ||
+        dirMoveFromTo(sameInodeChange, e) ||
+        dirMoveIdentical(sameInodeChange, e) ||
```
sebn: Thanks to the new renamed functions, we can easily see differences between branches...
Maybe those differences are normal, in which case we could add a comment explaining e.g. why some case is not relevant for files while it is for dirs...

core/local/change.js
```diff
@@ -129,13 +90,66 @@ function isChildAdd (a /*: LocalChange */, b /*: LocalChange */) { return childO
...
+      return (
+        fileMoveFromTo(sameInodeChange, e) ||
+        fileMoveSuccessiveTo(sameInodeChange, e) ||
+        fileMoveIdenticalOffline(e) ||
+        fromEvent(e)
+      )
+    case 'addDir':
+      return (
+        dirMoveSuccessiveTo(sameInodeChange, e) ||
+        dirMoveFromTo(sameInodeChange, e) ||
```
sebn: Thanks to the new renamed functions, we can easily see differences between branches...
E.g. why are those two in reverse order compared to the (file) `'add'` branch?

sebn: Maybe some of those differences are normal, in which case we could add a comment explaining why some case is relevant on one branch but not on the other.

core/local/change.js
```diff
@@ -286,71 +317,128 @@ const ensureValidMoveEvent = (moveChange /*: LocalMove */, event /*: LocalMoveEv
...
+  if (existingMove && existingMove.type === 'DirMove') {
+    const moveChange = _.clone(existingMove)
+    if (!moveChange.wip &&
+         moveChange.path === e.path &&
+         moveChange.stats.ino === e.stats.ino
+       ) {
+      // FIXME This is based on a bug in chokidar where
+      // an overwriting move have two addDir events on mac+APFS
+      // but no unlinkDir for the overwritten destination.
+      moveChange.overwrite = true
```
sebn: Maybe this branch would deserve its own function (to be called directly).

sebn:
I believe builds should be green soon :tada: 
Although I'm still wondering whether this breaks something, especially the `changes` array removal.
Except for this, I'm quite happy with the git added/deleted ratio Vs the huge readability improvement. :smile:

core/local/change.js
```diff
@@ -88,24 +67,6 @@ function build (type /*: string */, path /*: string */, opts /*: ?{stats?: fs.St
   return change
 }
 
-function maybeAddFile (a /*: ?LocalChange */) /*: ?LocalFileAddition */ { return (a && a.type === 'FileAddition') ? a : null }
-function maybePutFolder (a /*: ?LocalChange */) /*: ?LocalDirAddition */ { return (a && a.type === 'DirAddition') ? a : null }
-function maybeMoveFile (a /*: ?LocalChange */) /*: ?LocalFileMove */ { return (a && a.type === 'FileMove') ? a : null }
-function maybeMoveFolder (a /*: ?LocalChange */) /*: ?LocalDirMove */ { return (a && a.type === 'DirMove') ? a : null }
-function maybeDeleteFile (a /*: ?LocalChange */) /*: ?LocalFileDeletion */ { return (a && a.type === 'FileDeletion') ? a : null }
-function maybeDeleteFolder (a /*: ?LocalChange */) /*: ?LocalDirDeletion */ { return (a && a.type === 'DirDeletion') ? a : null }
```
sebn: I'm so glad to remove those crappy flow-related functions.

core/local/change.js
```diff
@@ -88,24 +67,6 @@ function build (type /*: string */, path /*: string */, opts /*: ?{stats?: fs.St
...
-    const anyChange = changes[i]
-    const rightTypeChange /*: ?T */ = maybeRightType(anyChange)
-    if (rightTypeChange != null && predicate(rightTypeChange)) {
-      if (remove) changes.splice(i, 1)
-      return rightTypeChange
-    }
-  }
-}
-
 function isChildMove (a /*: LocalChange */, b /*: LocalChange */) /*: boolean %checks */ {
```
sebn: This one is actually not used anymore too (was inlined in analysis), but I didn't remove it yet.

core/local/change.js
sebn: Also, in case your editor has an outline view, you'll appreciate the new names natural sorting :smile:

PR #1193, opened 2018-10-09
core/local: Fix non-empty dir moves `a/` to `c/` and `b/` to `a/`

sebn:


sebn:
It actually fails for a more complex case.

core/local/change.js
```diff
@@ -172,7 +172,6 @@ function dirMoveFromUnlinkAdd (unlinkChange /*: LocalDirDeletion */, e /*: Local
     stats: e.stats,
     old: unlinkChange.old,
     ino: unlinkChange.ino,
-    overwrite: e.old,
```
sebn: This is the key change.

aenario: This is a revert of https://github.com/cozy-labs/cozy-desktop/pull/1181/files , no ?

PR #1200, opened 2018-10-11
Update onboarding procedure

taratatach:
For Fedora 28:
  - add transifex-client, docker and docker-compose dependencies

  Add build and run steps, including the creation of a Transifex
  account (i.e. to get an API token).

  Add command to start the included development cozy-stack via docker in
  the testing procedure.

- [x] PR is not too big

### DIFF
--- a/doc/developer/requirements.md
+++ b/doc/developer/requirements.md
@@ -79,13 +79,13 @@ parameters below):
     docker-machine create default
     eval $(docker-machine env default)
 
-### Fedora 27
+### Fedora 28
 
-    sudo dnf install docker gcc-c++ git nodejs
-    cd /etc/yum.repos.d/
-    sudo wget https://dl.yarnpkg.com/rpm/yarn.repo
-    cd -
-    sudo dnf install yarn
+    sudo wget https://dl.yarnpkg.com/rpm/yarn.repo -O /etc/yum.repos.d/yarn.repo
+    sudo dnf install docker gcc-c++ git nodejs yarn transifex-client python3-pip
+    sudo pip install docker-compose
+    sudo groupadd docker
+    sudo usermod -aG docker $USER # Restart your session so this takes effect
 
 ### Ubuntu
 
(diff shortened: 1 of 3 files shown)


edas:
👍

doc/developer/requirements.md
```diff
@@ -79,13 +79,13 @@ parameters below):
...
-    sudo dnf install docker gcc-c++ git nodejs
-    cd /etc/yum.repos.d/
-    sudo wget https://dl.yarnpkg.com/rpm/yarn.repo
-    cd -
-    sudo dnf install yarn
+    sudo wget https://dl.yarnpkg.com/rpm/yarn.repo -O /etc/yum.repos.d/yarn.repo
+    sudo dnf install docker gcc-c++ git nodejs yarn transifex-client python3-pip
+    sudo pip install docker-compose
+    sudo groupadd docker
+    sudo usermod -aG docker $USER # Restart your session so this takes effect
```
sebn: :+1:

doc/developer/requirements.md
```diff
@@ -127,3 +122,33 @@ You can also run any cozy-stack command with the `yarn cozy-stack` script, e.g.:
...
 yarn cozy-stack apps install --domain cozy.tools:8080 drive 'git://github.com/cozy/cozy-drive.git#build-drive'
 ```
+
+## cozy-desktop
+
+### Create a transifex API token
+
+- Create an account on http://www.transifex.com
+- Join the Cozy team
+- Get your API token from the account settings page
```
sebn: It should not really be a requirement to build the app, assuming the translations in git are up-to-date.
And I wonder whether making it a requirement is packager-friendly...

Maybe we could just explain that the build script will try to update the translations when the `transifex` client is on the `PATH`, assuming one already has an account & the client is already set up with a token?
We could later add a doc explaining the transifex setup & workflow and add a link to it from here? WDYT?

doc/developer/requirements.md
```diff
@@ -127,3 +122,33 @@ You can also run any cozy-stack command with the `yarn cozy-stack` script, e.g.:
...
+yarn install
+```
+
+### Build everything
+
+```
+yarn build
+```
+
+The requested token is your transifex API token.
```
sebn: Those are already in `setup.md`.

Initially we only had `doc/developer/setup.md`.
Then @enguerran extracted `doc/developer/requirements.md` from it.
Now we're moving yarn install/build sections from `setup.md` to `requirements.md`...
So I'm a bit confused :slightly_smiling_face:

What do we expect from this `requirements.md` file?
Should it list everything required to build the app? (what I was expected, but maybe I'm wrong)
In which case, `yarn install` could be considered a requirement, but not `yarn build` since it's the actual build?

What else from `setup.md` should or should not go into `requirements.md`?
Should we turn `setup.md` into a `README.md` including links to other docs (or just single commands when possible) in the most logical order? (sounds good to me)

Please note building could deserve its own `build.md` doc, explaining how to build Elm/Stylus & update transifex translations at once (this is `yarn build`) and also how to build the final packages (`yarn dist` and its various options). But this one is definitely out of the scope of this PR.

doc/developer/requirements.md
sebn: (we could also try to fix the npm script)

sebn: Or just change the section title to:

```markdown
## Transifex (optional)
```

taratatach: I added this part because the setup procedure was leading to the question being raised and I had no idea what this was about.
Since this is optional, we should make sure the procedure doesn't lead to it unless we request it.

doc/developer/requirements.md
taratatach: I added the yarn commands because I didn't see anything about checking the build is working. I'm not sure separating requirements from the rest of the setup is best but this was probably done for a reason.

Maybe I could add a link back to `setup.md` from the requirements instead?

PR #1201, opened 2018-10-12
core/local/analysis: Don't confuse addDir(a)+addDir(a) with renaming

sebn:


sebn:
Congrats to @aenario for ending with the exact same conclusion on #1202 :smile:

PR #1202, opened 2018-10-12
core/analysis: multiple adds for same path is not move

aenario:

- [x] PR is not too big


aenario:
Turns out different symptoms add the same cause (see https://github.com/cozy-labs/cozy-desktop/pull/1201)

PR #1204, opened 2018-10-15
core/config: Reset corrupted config file on start

taratatach:
A corrupted config.json file is not parsable and requiring it throws
  an uncaught error, preventing the app to start.
  Since we couldn't figure out how this file gets corrupted and since it
  can get corrupted by any other software, we catch this error and reset
  the file so the app can start in any case.

- [x] PR is not too big

### DIFF
--- a/core/config.js
+++ b/core/config.js
@@ -4,6 +4,11 @@ const _ = require('lodash')
 const path = require('path')
 
 const { hideOnWindows } = require('./utils/fs')
+const logger = require('./logger')
+
+const log = logger({
+  component: 'Config'
+})
 
 // Config can keep some configuration parameters in a JSON file,
 // like the devices credentials or the mount path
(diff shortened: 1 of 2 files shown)


core/config.js
```diff
@@ -20,7 +25,17 @@ module.exports = class Config {
       this.reset()
     }
 
-    this.config = require(this.configPath)
+    try {
+      this.config = require(this.configPath)
+    } catch (e) {
+      if (e instanceof SyntaxError) {
+        log.error(`Could not read config file at ${this.configPath}:`, e)
```
aenario: Would it makes sense to attempt to remove the broken char before giving up? Could be in a separate PR.

taratatach: After discussing it, we decided not to make any attempt at fixing the config file.
:ballot_box_with_check:

PR #1205, opened 2018-10-16
core/config: Persist config via temporary file

taratatach:
To avoid losing any details of the configuration if the app were to crash or be closed during the persistance process, we use a temporary file.

If the temporary file is not copied (and removed) before we close the app, the next time we fire it up, we'll load this temporary config first.

If the temporary config is not valid (e.g. we didn't have the time to finish writing its content), we remove it and load the existing config.

If we load the existing config and it's not valid, we reset it.

To make sure the tmp config path is writable and its content is only visible to the current user, we reuse the existing config path and add a `.tmp` suffix.

- [x] PR is not too big

### DIFF
--- a/core/config.js
+++ b/core/config.js
@@ -16,28 +16,41 @@ module.exports = class Config {
   // Create config file if it doesn't exist.
   constructor (basePath) {
     this.configPath = path.join(basePath, 'config.json')
+    fs.ensureFileSync(this.configPath)
     this.dbPath = path.join(basePath, 'db')
     fs.ensureDirSync(this.dbPath)
     hideOnWindows(basePath)
-    fs.ensureFileSync(this.configPath)
 
-    if (fs.readFileSync(this.configPath).toString() === '') {
-      this.reset()
-    }
+    this.config = this.read()
+  }
 
+  // Load a config JSON file or return an empty object
+  static safeLoad (configPath) {
     try {
-      this.config = fs.readJSONSync(this.configPath)
+      return fs.readJSONSync(configPath)
     } catch (e) {
       if (e instanceof SyntaxError) {
-        log.error(`Could not read config file at ${this.configPath}:`, e)
-        this.reset()
-        this.config = require(this.configPath)
+        log.error(`Could not read config file at ${configPath}:`, e)
+        fs.unlinkSync(configPath)
+        return {}
       } else {
         throw e
       }
     }
+  }
 
-    // FIXME: autoBind(this)
+  // Read the configuration from disk
+  read () {
+    if (fs.existsSync(this.tmpConfigPath)) {
+      const tmpConfig = Config.safeLoad(this.tmpConfigPath)
+
+      if (_.size(tmpConfig) > 0) {
+        this._moveTmpConfig()
+        return tmpConfig
+      }
+    }
+
+    return Config.safeLoad(this.configPath)
   }
 
   // Reset the configuration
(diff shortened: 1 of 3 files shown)


core/config.js
```diff
@@ -2,42 +2,73 @@
...
+
+  // Import the configuration from disk
+  static import (configPath) {
+    let config
+
+    if (fs.existsSync(TMP_CONFIG_PATH)) {
+      config = this.safeLoad(TMP_CONFIG_PATH)
+
+      if (!!config && !!config.configPath) {
+        moveTmpConfig(config.configPath)
```
aenario: :warning: this can throw (for instance if the user does not have permission to write `config.configPath`) and leave the app unusable.

aenario: Idem for persist.

sebn:
@taratatach Needs rebase ;)

core/config.js
taratatach: Actually, I think there's a more important issue here. Writing the file to an unknown location could be used to overwrite completely unrelated files and be used for malicious purposes.
I'll think about what can be done here to make sure a config path change is made only at the user's request.

core/config.js
```diff
@@ -16,28 +16,43 @@ module.exports = class Config {
...
     fs.ensureDirSync(this.dbPath)
     hideOnWindows(basePath)
-    fs.ensureFileSync(this.configPath)
 
-    if (fs.readFileSync(this.configPath).toString() === '') {
-      this.reset()
-    }
+    this.config = this.import(this.configPath)
 
+    // FIXME: autoBind(this)
```
sebn: Should we fix it?

core/config.js
```diff
@@ -49,14 +64,30 @@ module.exports = class Config {
...
   }
 
   // Transform the config to a JSON string
   toJSON () {
     return JSON.stringify(this.config, null, 2)
   }
 
+  // Get the tmp config path associated with the current config path
+  get tmpConfigPath () {
+    return this.configPath + '.tmp'
```
sebn: :+1:

core/config.js
```diff
@@ -16,28 +16,43 @@ module.exports = class Config {
...
+        return {}
       } else {
         throw e
       }
     }
+  }
 
-    // FIXME: autoBind(this)
+  // Import the configuration from disk
+  import () {
```
sebn: Won't the method name collide with the `import` keyword as soon as we upgrade node?

taratatach: I was wondering about this (I got a hint with the syntax highlighting. I'll do some research about this.

core/config.js
```diff
@@ -16,28 +16,43 @@ module.exports = class Config {
   // Create config file if it doesn't exist.
   constructor (basePath) {
     this.configPath = path.join(basePath, 'config.json')
+    fs.ensureFileSync(this.configPath)
```
nono: It's weird that we ensure a file exists before ensuring that its parent directory exists.

core/config.js
```diff
@@ -16,28 +16,43 @@ module.exports = class Config {
...
     fs.ensureDirSync(this.dbPath)
     hideOnWindows(basePath)
-    fs.ensureFileSync(this.configPath)
 
-    if (fs.readFileSync(this.configPath).toString() === '') {
-      this.reset()
-    }
+    this.config = this.read()
 
+    // FIXME: autoBind(this)
```
nono: Maybe we can remove this FIXME as it's here since a long time and nobody have feel the need to fix it.

core/config.js
taratatach: We don't ensure the parent directory exists later on. Besides, ensuring a file exists will ensure the whole path leading to it exists.

core/config.js
taratatach: Sure thing.

PR #1207, opened 2018-10-17
core/merge: Fix move conflicting destination id

sebn:
Conflicting file/dir move destination id was not matching its path (it was missing the conflict suffix).


sebn:
Not sure how adding tests would decrease test coverage... :confused:

sebn:
Maybe confirm we're merging the conflict resolution result to handle local move?

PR #1208, opened 2018-10-17
core/config: Replace require with fs.readJSON

taratatach:
NodeJS' require maintains a cache of imported modules or, in our case,
  configurations.
  This is an issue in tests where loading an empty config results in a
  complete config object.

  In our situation, we can easily replace require with fs.readJSON
  which does not maintain any cache.

- [x] PR is not too big

### DIFF
--- a/test/unit/app.js
+++ b/test/unit/app.js
@@ -179,8 +179,9 @@ describe('App', function () {
     it('works when app is configured', function () {
       configHelpers.createConfig.call(this)
       configHelpers.registerClient.call(this)
-      const app = new App(this.basePath)
+      this.config.persist() // the config helper does not persist it
 
+      const app = new App(this.basePath)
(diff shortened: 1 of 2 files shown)


test/unit/app.js
```diff
@@ -179,8 +179,9 @@ describe('App', function () {
     it('works when app is configured', function () {
       configHelpers.createConfig.call(this)
       configHelpers.registerClient.call(this)
-      const app = new App(this.basePath)
+      this.config.persist() // the config helper does not persist it
 
+      const app = new App(this.basePath)
```
sebn: I usually separate the *Given* / *When* / *Then* steps in my tests (except for very short ones), hence the `const app` assignment grouped with `configHelpers` calls since both are part of the context (aka *Given*), while the `const app` assignment is the actual subject (aka *When*).

PR #1214, opened 2018-10-19
WIP: core/local: Fix harmless file/dir identical renaming loopback issues

sebn:
The fixes in `core/loca/change` are quite ugly.
This is due to the fact that we have no simple way to turn the deletion into a move because this would require a source doc while we only have the destination one in case of an identical renaming.
Which brings us to the fact that a few things have misleading names (e.g. `old`) and some legitimate cases are currently not possible to represent when getting some events + doc combinations (e.g. identical renaming loopback).

The scenario cases actually pass without the fix even with the new captures, because we don't fail on error, but the *Invalid move* messages are still visible in logs.
So those are only fixed according to unit tests for now.

### DIFF
--- a/core/local/change.js
+++ b/core/local/change.js
@@ -156,6 +156,7 @@ function _fromEvent (e/*: LocalEvent */) /*: LocalChange */ {
 }
 
 function fileMoveFromUnlinkAdd (unlinkChange /*: LocalFileDeletion */, e /*: LocalFileAdded */) /*: * */ {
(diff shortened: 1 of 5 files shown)


sebn:
To be rebased on #1215

aenario:
This was rebased on #1215 and merged into master.

aenario:
This was rebased on #1215 and merged into master.

PR #1220, opened 2018-10-22
Correctly handle offline status at startup

aenario:


aenario:
This appears to work from manual testing. Trying to figure out a way to unit-test it.

core/remote/warning_poller.js
```diff
@@ -80,7 +80,6 @@ class RemoteWarningPoller {
       this.events.emit('remoteWarnings', warnings)
     } catch (err) {
       log.error({err})
-      throw err
```
taratatach: Maybe we should swallow only some types of errors instead of all errors?

core/remote/watcher.js
```diff
@@ -109,6 +110,7 @@ class RemoteWatcher {
         throw userActionRequired.includeJSONintoError(err)
       } else {
         log.error({err})
+        this.events.emit('offline')
```
taratatach: There's a number of other errors we could get besides an unreachable host. What about emitting this event only if the error status is `0`?

core/remote/watcher.js
```diff
@@ -82,6 +82,7 @@ class RemoteWatcher {
   async watch () {
     try {
       const seq = await this.pouch.getRemoteSeqAsync()
+      this.events.emit('online')
```
sebn: I guess this would unpause the GUI but not the Sync. User would see the app as unpaused, but pending Sync changes would still not happen (looks like the behavior I got during validation).

test/unit/remote/offline.js
```diff
@@ -0,0 +1,49 @@
...
+const {
+  deleteAll, createTheCouchdbFolder
+} = require('../../support/helpers/cozy')
+
+/*::
+import type { Metadata } from '../../../core/metadata'
+import type { RemoteDoc, JsonApiDoc } from '../../../core/remote/document'
+*/
+
+describe('Remote', function () {
```
sebn: Not really a unit test.

core/remote/watcher.js
sebn: (not a blocker since Sync should have unpaused itself after 1 minute)

PR #1223, opened 2018-10-24
test: Reproduce 2 move merge issues

sebn:
So we can fix them at some point.


aenario:
With a merge commit in order to keep commits referenced in trello.

PR #1224, opened 2018-10-25
travis: Make built artifacts available in each PR

taratatach:
Each time we build a Pull Request on travis, we build the binary
  artifacts for linux and mac OSX.

  We upload those binaries to bintray.com and add a comment on the PR so
  we can easily find them and test the Pull Request before merging it.

- [x] PR is not too big

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -90,6 +90,10 @@ after_failure: ./dev/ci/after_failure.sh
 branches:
     only:
         - master
+stages:
+    - test
+    - name: notify
+      if: type = pull_request
 jobs:
     fast_finish: true
     include:
@@ -119,3 +123,10 @@ jobs:
           os: linux
           after_failure: skip
           script: ./dev/ci/coverage.sh
+        - stage: notify
+          name: artifacts built
+          services: false
+          cache: false
+          before_install: skip
+          install: skip
+          script: ./dev/ci/notify.sh
--- a/dev/ci/dist.sh
+++ b/dev/ci/dist.sh
@@ -19,3 +19,18 @@ if [ "$TRAVIS_OS_NAME" == "linux" ]; then
 else
   yarn dist:all
 fi
+
+if [ "$TRAVIS_PULL_REQUEST" != "false" ]; then
+    version=$(git rev-parse --short HEAD);
+    bintray_version="PR-$TRAVIS_PULL_REQUEST";
+    dist_files=$(ls -p dist/ | grep -v /);
+
+    IFS=$'\n'
+    for file in $dist_files; do
+      bintray_path="$version-${file/ /%20}";
+      curl -u "$BINTRAY_USER:$BINTRAY_API_TOKEN" \
+           -T "dist/$file" \
+           "https://api.bintray.com/content/$BINTRAY_ORG/$BINTRAY_REPO/$BINTRAY_PACKAGE/$bintray_version/$bintray_path?publish=1&override=1";
+    done
+    unset IFS
+fi
--- a/dev/ci/notify.sh
+++ b/dev/ci/notify.sh
@@ -1,0 +1,17 @@
+#!/bin/bash
+
+if [ "$TRAVIS_PULL_REQUEST" != "false" ]; then
+    version=$(git rev-parse --short HEAD);
+
+    comment_body="Available development artifacts for commit $version:\n";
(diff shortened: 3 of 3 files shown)


taratatach:
test comment

taratatach:
test comment

taratatach:
$comment_body

taratatach:
Dev artifacts for commit 69822b25
[file1](http://example.com)
[file2](http://example.com)

.travis.yml
```diff
@@ -1,18 +1,23 @@
...
-        - os: osx
-          node_js: 8
+node_js: 8
+addons:
+  apt:
+      sources:
+          - ubuntu-toolchain-r-test
+      packages:
+          - gcc-4.8
+          - g++-4.8
```
nono: Maybe we can use a more recent gcc (https://launchpad.net/~ubuntu-toolchain-r/+archive/ubuntu/test?field.series_filter=trusty) to speed thing a lit bit (to be tested)

.travis.yml
```diff
@@ -26,117 +31,112 @@ env:
...
+          os: linux
+        - <<: *dist
+          os: osx
+        - stage: test
+          name: code coverage
+          os: linux
+          before_install: skip
+          install: skip
+          after_failure: skip
+          script: bash <(curl -s https://codecov.io/bash)
```
nono: I think you have cheated a bit: we need to generated the coverage (`yarn test:unit:coverage`) before uploading it to codecov (`bash <(curl -s https://codecov.io/bash)`)

.travis.yml
```diff
@@ -26,117 +31,112 @@ env:
...
+          script: ./script/unit.sh
+          name: unit tests
+          os: linux
+        - <<: *test_unit
+          os: osx
+        - &test_integration
+          stage: test
+          script: ./script/integration.sh
+          name: integration tests
+          os: linux
```
nono: I think that we should keep unit and integration tests together. For those tests, the setup is a lot longer that running the tests, and running both blocks travis workers for other pull requests

script/after_failure.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
```
nono: Maybe we can put those scripts in `dev/ci` instead of `script` as we already have a lot of things at the root of the repository

.travis.yml
```diff
@@ -26,117 +31,112 @@ env:
...
-        sources:
-            - ubuntu-toolchain-r-test
-        packages:
-            - gcc-4.8
-            - g++-4.8
+          #- CXX=$([ "$TRAVIS_OS_NAME" = "linux" ] && echo "g++-4.8" || echo $CXX)
 
-# install cozy stack for integration test
-before_install:
+before_install: | # install cozy stack for integration test
```
nono: We can move all the build_install to a script to be consistent with other steps

.travis.yml
taratatach: OK. No big deal here.

.travis.yml
taratatach: Actually I tried and it wouldn't work because the `travis_retry` function doesn't seem to be available inside external bash scripts :(

nono: OK, it's a good reason.

For the record, https://twitter.com/plexus/status/499194992632811520 says that you can copy the `travis_retry` function.

taratatach: I would definitely prefer to see the setup in another file. What would be best in your opinion? Copying the function into a `travis_helpers.sh` file or copying the entire `header.sh` file from the travis repo?

nono: Just a copy of the function, with a link in a comment from where it's taken

sebn:
@taratatach Is it normal to have 2 *artifacts build*  in *Test* and 1 in *Notify*?

taratatach:
@sebn yes, you can see it as a merge step. We create only one comment with the artifacts of both linux and osx.

sebn:
Oops envvars are not expanded in asset links :wink:

taratatach:
Links are fixed.

sebn:
64bits AppImage is half the size of [3.10.1](https://github.com/cozy-labs/cozy-desktop/releases/tag/v3.10.1) release, not sure why.

sebn:
The 64 bits AppImage seems to work.
Regarding the macOS dmg, the app detects an update on start, not sure why, investigating...

sebn:
This was probably due to the overriding of the previous version. Removing the old macOS app then adding the new one seems to work as expected.

:ship:

PR #1233, opened 2018-10-29
travis: Run tests and other jobs in parallel

taratatach:
Travis offers running different jobs in parallel to speed up builds.
  If we can run more than 2 jobs in parallel (it depends on load) then
  we can get the final results of our builds much more quickly.

  This commit separates unit & integration tests, real life scenarios,
  building the binaries and running codecov.

- [x] PR is not too big

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,18 +1,23 @@
 ---
 dist: trusty
 language: node_js
-branches:
-    only:
-        - master
-matrix:
-    fast_finish: true
-    include:
-        - os: linux
-          node_js: 8
-          env: CXX=g++-4.8
-          # @TODO macOS build is much slower
-        - os: osx
-          node_js: 8
+node_js: 8
+addons:
+  apt:
+      sources:
+          - ubuntu-toolchain-r-test
+      packages:
+          - gcc-8
+          - g++-8
+          - jq
+services:
+  - docker
+cache:
+  yarn: true
+  directories:
+      - $HOME/.cache/electron
+      - $HOME/.cache/electron-builder
+      - $HOME/bin
 env:
     global:
         - COZY_DESKTOP_DIR=/tmp/cozy-desktop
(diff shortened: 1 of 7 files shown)


.travis.yml
```diff
@@ -26,117 +31,101 @@ env:
...
-        - $HOME/.cache/electron-builder
-        - $HOME/bin
-addons:
-    apt:
-        sources:
-            - ubuntu-toolchain-r-test
-        packages:
-            - gcc-4.8
-            - g++-4.8
+          #- CXX=$([ "$TRAVIS_OS_NAME" = "linux" ] && echo "g++-4.8" || echo $CXX)
```
taratatach: I commented this variable definition since it made some tests fail in this parallel situation and actually doesn't seem to be useful but wanted to discuss its removal.

nono:
[Améliorer les builds travis](https://trello.com/c/FFj7JbLu/1120-am%C3%A9liorer-les-builds-travis)

dev/ci/coverage.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
+yarn build
```
sebn: Maybe `yarn build` is not needed for coverage.

dev/ci/tests.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+
+yarn build
```
sebn: Maybe only `yarn build:elm` is needed for tests...

dev/ci/coverage.sh
taratatach: I tried without and it wouldn't work (some command wasn't found if I remember correctly)

dev/ci/scenarios.sh
```diff
@@ -0,0 +1,12 @@
...
+
+export DISPLAY=:99.0
+if [ "${TRAVIS_OS_NAME}" == "osx" ]; then
+  ( sudo Xvfb :99 -ac -screen 0 1024x768x8; echo ok )&
+else
+  sh -e /etc/init.d/xvfb start;
+fi
+sleep 3 # give xvfb some time to start
+
+yarn build
```
sebn: Maybe `yarn build` is not needed for scenarios...

dev/ci/scenarios.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
+export DISPLAY=:99.0
+if [ "${TRAVIS_OS_NAME}" == "osx" ]; then
+  ( sudo Xvfb :99 -ac -screen 0 1024x768x8; echo ok )&
+else
+  sh -e /etc/init.d/xvfb start;
+fi
+sleep 3 # give xvfb some time to start
```
sebn: Maybe we could extract and source the xvfb hack?

dev/ci/coverage.sh
sebn: That's weird, `yarn build` only builds elm + stylus and pulls from transifex...

dev/ci/scenarios.sh
taratatach: Actually I did that before but it seems it needs to run in the same shell as the following commands to work.

taratatach: But maybe that's possible to do with something like `("path/to/script")`

dev/ci/coverage.sh
taratatach: OK. This issue could come from somewhere else. I've tried to remove it.

dev/ci/scenarios.sh
sebn: I thought `source some/script.sh` could work, but I may be wrong...

taratatach: you were not ;-)

.travis.yml
nono: I think you can delete this line

PR #1236, opened 2018-10-30
jq: Improve filters [ci skip]

sebn:
- Include Chokidar in Local filter
- Introduce nouptodate & noproxy filters
- Document path filter caveat
- Find logs before/after some given time


sebn:
@aenario Sorry, I added two more filters: `before` and `after`.

PR #1241, opened 2018-10-31
[Refacto] Rename the local Watcher to ChokidarWatcher

nono:


sebn:
@nono I assume the goal is to introduce a 2nd watcher to be used only on Windows?

nono:
@sebn Yes, the goal is to be able to add new watchers without a giant PR that needs to stop the world for merging it. For the moment, I'm working on a Linux watcher, but it won't probably be stable enough to be used in production before sometimes. It's just a step to add the new Windows watcher.

core/local/index.js
```diff
@@ -13,7 +13,7 @@ const logger = require('../logger')
 const { isUpToDate } = require('../metadata')
 const { hideOnWindows } = require('../utils/fs')
 const sentry = require('../sentry')
-const Watcher = require('./watcher')
+const watcher = require('./watcher')
```
taratatach: Why the lower case here while you haven't changed the case in other files?

nono: With an upper-case, flow complains. I needs to manually import the `Watcher` type: L31 `import type { Watcher } from './watcher'`, and it works only if I don't use `Watcher` here too.

So, it's probably just my flow-fu that is not good enough :p

core/local/watcher.js
```diff
@@ -1,450 +1,26 @@
...
-  emptyDirRetryCount: number,
-  flushed: boolean,
-  resolve: () => void
+export type Watcher = {
+  checksumer: Checksumer,
+  running: Promise<*>,
+  start: () => Promise<*>,
+  stop: (force: ?bool) => Promise<*>,
+  ensureDirSync: () => void,
+  onFlush: (ChokidarEvent[]) => *
```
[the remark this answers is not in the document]
nono: I've started reading about the interface, but I'm a bit lost on covariance and contravariance, and I think I need one of those for things to work here. I'll try again later when I'll be more advanced in flow.

nono:
@taratatach for the tests, I've tried to use the generic `watcher.build` when possible, but most tests depend of the specific behavior of Chokidar and/or some properties of the ChokidarWatcher.

PR #1242, opened 2018-10-31
Run unit tests only one time on Travis/Linux

nono:


sebn:
@taratatach I believe this is because we have to (1) instrument the code before (2) running the unit tests in order to (3) generate the coverage report. Both 1-2-3 steps could theoretically run separately, but since you need to do 1 before 2 in order to get 3... I can't think of another way (but I may be missing something)

PR #1243, opened 2018-11-02
locales: update ES locale

taratatach:

- [x] PR is not too big


aenario:
We should also probably update this part : 
https://github.com/cozy-labs/cozy-desktop/blob/master/gui/js/i18n.js#L9

taratatach:
@aenario you mean activating the es locale?

nono:
@taratatach yes, we should activate the es locale

PR #1245, opened 2018-11-05
Make travis/macOS builds faster

nono:
We can avoid the slow brew update by installing CouchDB via an URL


sebn:
> Shouldn't we fix the version of couchDB we fetch?

+1 in a separate PR (this one doesn't change anything regarding CouchDB version and improves build time).

PR #1246, opened 2018-11-05
fix: Fetch changesFeed with recursive paginated call

enguerran:
When changesFeed `_changes` are very big, the request fails with timeout.
The idea of this PR is to fetch changesFeed by chunks of 10k items.
Implementation is recursive.

What do you think?

### DIFF
--- a/core/remote/cozy.js
+++ b/core/remote/cozy.js
@@ -281,11 +281,14 @@ async function getChangesFeed (since, client) {
     return response
   }
   const nextResponse = await getChangesFeed(last_seq, client)
-  return {
-    ...nextResponse,
-    results: [
-      ...results,
-      ...nextResponse.results
-    ]
-  }
+  return Object.assign(
+    {},
+    nextResponse,
+    {
+      results: [
+        ...results,
+        ...nextResponse.results
+      ]
+    }
+  )
 }
--- a/core/remote/cozy.js
+++ b/core/remote/cozy.js
@@ -125,8 +125,7 @@ class RemoteCozy {
   }
 
   async changes (since/*: string */ = '0') /*: Promise<{last_seq: string, docs: Array<RemoteDoc|RemoteDeletion>}> */ {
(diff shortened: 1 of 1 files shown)


core/remote/cozy.js
```diff
@@ -274,3 +273,19 @@ module.exports = {
   FSCK_PATH,
   RemoteCozy
 }
+
+async function getChangesFeed (since, client) {
```
nono: Can you add flow annotations here?

core/remote/cozy.js
```diff
@@ -274,3 +273,19 @@ module.exports = {
...
+  const response = await client.data.changesFeed(FILES_DOCTYPE, { since, include_docs: true, limit: 10000 })
+  const {last_seq, pending, results} = response
+  if (pending === 0) {
+    return response
+  }
+  const nextResponse = await getChangesFeed(last_seq, client)
+  return {
+    ...nextResponse,
```
nono: Can object spread/rest be used in node 8? I think it is but I'd like to have a confirmation

enguerran: O_o 
```
➜  cozy-desktop (fix/changesfeedtimeout) ✗ yarn start
yarn run v1.12.1
$ env-cmd .env.dev electron .
App threw an error during load
cozy/cozy-desktop/core/remote/cozy.js:285
    ...nextResponse,
    ^^^

SyntaxError: Unexpected token ...
```

enguerran: fixed!

enguerran:
Commits are added as fixup to let me `rebase --autosquash` after the full revision process.

core/remote/cozy.js
sebn: I believe it's available [in 8.9.4 but not in 8.2.1](https://node.green/#ES2018-features-object-rest-spread-properties-object-rest-properties) (which is the one available with our current electron version) without a flag.

PR #1250, opened 2018-11-06
Improve metadata status after moves

aenario:

- [ ] PR is not too big

### DIFF
--- a/core/merge.js
+++ b/core/merge.js
@@ -215,7 +215,15 @@ class Merge {
         if (doc.class == null) { doc.class = file.class }
         if (doc.mime == null) { doc.mime = file.mime }
       } else if (!isUpToDate(side, file)) {
-        return this.resolveConflictAsync(side, doc, file)
+        await this.resolveConflictAsync(side, doc, file)
+        if (side === 'remote') {
+          // we just renamed the remote folder as a conflict
+          // the old file should be dissociated from the remote
+          delete file.remote
+          if (file.sides) delete file.sides.remote
+          await this.pouch.put(file)
+        }
+        return
       }
       if (sameFile(file, doc)) {
         log.info({path}, 'up to date')
--- a/core/merge.js
+++ b/core/merge.js
@@ -1,7 +1,8 @@
 /* @flow */
 
 const autoBind = require('auto-bind')
-const { clone } = require('lodash')
+const _ = require('lodash')
+const { clone } = _
 const { basename, dirname, extname, join } = require('path')
 
 const IdConflict = require('./IdConflict')
@@ -24,7 +25,7 @@ const fsutils = require('./utils/fs')
 /*::
 import type { IdConflictInfo } from './IdConflict'
 import type Local from './local'
-import type { SideName, Metadata } from './metadata'
+import type { SideName, Metadata, RemoteRevisionsByID } from './metadata'
 import type Pouch from './pouch'
 import type Remote from './remote'
 */
@@ -311,7 +312,7 @@ class Merge {
   }
 
   // Rename or move a folder (and every file and folder inside it)
-  async moveFolderAsync (side /*: SideName */, doc /*: Metadata */, was /*: Metadata */, newRevs /*:: ?: { [string]: string} */) {
+  async moveFolderAsync (side /*: SideName */, doc /*: Metadata */, was /*: Metadata */, newRemoteRevs /*: ?RemoteRevisionsByID */) {
     log.debug({path: doc.path, oldpath: was.path}, 'moveFolderAsync')
     if (was.sides && was.sides[side]) {
       const folder /*: ?Metadata */ = await this.pouch.byIdMaybeAsync(doc._id)
@@ -330,14 +331,14 @@ class Merge {
         const dst = await this.resolveConflictAsync(side, doc, folder)
         dst.sides = {}
         dst.sides[side] = 1
-        return this.moveFolderRecursivelyAsync(side, dst, was, newRevs)
+        return this.moveFolderRecursivelyAsync(side, dst, was, newRemoteRevs)
       } else {
         if (folder && doc.overwrite) {
           doc.overwrite = folder
           doc._rev = folder._rev
         }
         await this.ensureParentExistAsync(side, doc)
-        return this.moveFolderRecursivelyAsync(side, doc, was, newRevs)
+        return this.moveFolderRecursivelyAsync(side, doc, was, newRemoteRevs)
       }
     } else { // It can happen after a conflict
       return this.putFolderAsync(side, doc)
@@ -345,7 +346,7 @@ class Merge {
   }
 
   // Move a folder and all the things inside it
-  async moveFolderRecursivelyAsync (side /*: SideName */, folder /*: Metadata */, was /*: Metadata */, newRevs /*:: ?: { [string]: string } */) {
+  async moveFolderRecursivelyAsync (side /*: SideName */, folder /*: Metadata */, was /*: Metadata */, newRemoteRevs /*: ?RemoteRevisionsByID */) {
     const docs = await this.pouch.byRecursivePathAsync(was._id)
     move(was, folder)
     let bulk = [was, folder]
@@ -362,9 +363,8 @@ class Merge {
 
       let existingDstRev = existingDstRevs[dst._id]
       if (existingDstRev && folder.overwrite) dst._rev = existingDstRev
-      if (side === 'remote' && newRevs && newRevs[dst.remote._id]) {
-        dst.remote._rev = newRevs[dst.remote._id]
-      }
+      const newRemoteRev = _.get(newRemoteRevs, _.get(dst, 'remote._id'))
+      if (newRemoteRev) dst.remote._rev = newRemoteRev
 
       markSide(side, dst, src)
       bulk.push(src)
--- a/core/merge.js
+++ b/core/merge.js
@@ -10,6 +10,7 @@ const {
   assignId,
   assignMaxDate,
   detectPlatformIncompatibilities,
+  isAtLeastUpToDate,
   isUpToDate,
   markSide,
   sameBinary,
@@ -214,7 +215,7 @@ class Merge {
         if (doc.size == null) { doc.size = file.size }
         if (doc.class == null) { doc.class = file.class }
         if (doc.mime == null) { doc.mime = file.mime }
-      } else if (!isUpToDate(side, file)) {
+      } else if (!isAtLeastUpToDate(side, file)) {
         await this.resolveConflictAsync(side, doc, file)
         if (side === 'remote') {
           // we just renamed the remote folder as a conflict
--- a/core/merge.js
+++ b/core/merge.js
@@ -219,7 +219,7 @@ class Merge {
       } else if (!isAtLeastUpToDate(side, file)) {
         await this.resolveConflictAsync(side, doc, file)
         if (side === 'remote') {
-          // we just renamed the remote folder as a conflict
+          // we just renamed the remote file as a conflict
           // the old file should be dissociated from the remote
           delete file.remote
           if (file.sides) delete file.sides.remote
--- a/core/merge.js
+++ b/core/merge.js
@@ -302,7 +302,7 @@ class Merge {
   }
 
   // Rename or move a folder (and every file and folder inside it)
-  async moveFolderAsync (side /*: SideName */, doc /*: Metadata */, was /*: Metadata */) {
+  async moveFolderAsync (side /*: SideName */, doc /*: Metadata */, was /*: Metadata */, newRevs /*:: ?: { [string]: string} */) {
     log.debug({path: doc.path, oldpath: was.path}, 'moveFolderAsync')
     if (was.sides && was.sides[side]) {
       const folder /*: ?Metadata */ = await this.pouch.byIdMaybeAsync(doc._id)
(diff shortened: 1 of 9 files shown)


aenario:
To be totally honnest I have no idea why it works \o/

core/remote/change.js
```diff
@@ -40,11 +40,13 @@ export type RemoteChange =
   | RemoteFileUpdate
 
 export type RemoteIgnoredChange = {type: 'RemoteIgnoredChange', doc: Metadata|RemoteDoc|RemoteDeletion, detail: string}
+export type RemoteDescendantChange = {type: 'RemoteDescendantChange', doc: Metadata|RemoteDoc|RemoteDeletion, detail: string}
```
sebn: `RemoteNoise` is not used anymore.
Subtypes should be merged with `RemoteChange` ones.
Which requires using the `sideName` property instead of a `Remote` prefix in `type` (same as `RemoteChange`).
Done in #1262.

core/merge.js
```diff
@@ -109,6 +110,8 @@ class Merge {
     if (base.length > 180) {
       base = base.slice(0, 180)
     }
+    // avoid long chain of -conflict-DATE-conflict-DATE-conflict-DATE
+    base = base.split('-conflict-20')[0]
```
sebn: :+1: for extracting it to #1258

core/merge.js
```diff
@@ -210,8 +213,16 @@ class Merge {
         if (doc.size == null) { doc.size = file.size }
         if (doc.class == null) { doc.class = file.class }
         if (doc.mime == null) { doc.mime = file.mime }
-      } else if (!isUpToDate(side, file)) {
-        return this.resolveConflictAsync(side, doc, file)
+      } else if (!isAtLeastUpToDate(side, file)) {
+        await this.resolveConflictAsync(side, doc, file)
+        if (side === 'remote') {
+          // we just renamed the remote folder as a conflict
```
sebn: `s/folder/file/`? Or did I miss something?

core/merge.js
```diff
@@ -349,6 +360,9 @@ class Merge {
 
       let existingDstRev = existingDstRevs[dst._id]
       if (existingDstRev && folder.overwrite) dst._rev = existingDstRev
+      if (side === 'remote' && newRevs && newRevs[dst.remote._id]) {
```
sebn: Is the `side === 'remote'` check needed?
Wouldn't `newRevs` be null anyway? 
Maybe rename it to `remoteRevs` or `newRemoteRevs` everywhere to make it clear it's remote-specific?

core/merge.js
```diff
@@ -349,6 +360,9 @@ class Merge {
...
+        dst.remote._rev = newRevs[dst.remote._id]
```
sebn: Would it be possible for `dst.remote` to be null in case a child was added locally but not yet synced remotely?
Wouldn't using lodash's `get` work when `newRevs` and/or `dst.remote` are null?

```js
const newRemoteRev = _.get(newRevs, _.get(dst, 'remote._id'))
if (newRemoteRev) dst.remote._rev = newRemoteRev
```

core/metadata.js
```diff
@@ -253,6 +254,12 @@ function isUpToDate (side /*: SideName */, doc /*: Metadata */) {
   return currentRev === lastRev
 }
 
+function isAtLeastUpToDate (side /*: SideName */, doc /*: Metadata */) {
+  let currentRev = doc.sides[side] || 0
+  let lastRev = extractRevNumber(doc)
+  return currentRev >= lastRev
```
sebn: Would it work to just replace `isUpToDate` with it? (I wonder...)

core/remote/watcher.js
```diff
@@ -273,14 +273,17 @@ class RemoteWatcher {
             // move inside move
             change.was.path = remoteChange.applyMoveToPath(previousChange, change.was.path)
             change.needRefetch = true
-            return change
+            return change // FileMove
```
sebn: Maybe we should just rename `change` to `fileMove`?

core/sync.js
```diff
@@ -296,6 +295,9 @@ class Sync {
       } else {
         await this.doMove(side, doc, from)
       }
+      if (from.md5sum !== doc.md5sum) {
```
sebn: Use `sameBinary()`?

sebn:
I'm trying to understand the implications, so I'm not done yet :)

core/merge.js
```diff
@@ -298,7 +309,7 @@ class Merge {
   }
 
   // Rename or move a folder (and every file and folder inside it)
-  async moveFolderAsync (side /*: SideName */, doc /*: Metadata */, was /*: Metadata */) {
+  async moveFolderAsync (side /*: SideName */, doc /*: Metadata */, was /*: Metadata */, newRevs /*:: ?: { [string]: string} */) {
```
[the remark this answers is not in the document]
sebn: My bad, those are actually remote ids. Some hint is definitely needed :)

core/sync.js
```diff
@@ -296,6 +296,9 @@ class Sync {
       } else {
         await this.doMove(side, doc, from)
       }
+      if (!sameBinary(from, doc)) {
+        await side.overwriteFileAsync(doc, from)
+      }
```
sebn: Well spotted!

What happens when calling `sameBinary()` for a moved directory is not obvious, although it actually works.
But the whole `#applyDoc()` and `#do*()` methods need refactoring (out of the scope of this PR).

We're calling `Side#overwriteFileAsync()` directly, which means we save an `if`, but also `transfer-started` won't be emitted... Could this be an issue?

Actually I'm also wondering why we're emitting `*-started` events after the local/remote writes are complete...
Could this be because some old code didn't use `async` / `await` hence the events emitted before writes completion?

aenario: Looks like we just reused the event, but it has a new meaning of "transfer-made"

https://github.com/cozy-labs/cozy-desktop/commit/b3b3e446998d67744e64dfad765847f35b56bc15

core/remote/change.js
```diff
@@ -20,12 +20,13 @@ export type RemoteFileUpdate = {sideName: 'remote', type: 'FileUpdate', doc: Met
 export type RemoteDirAddition = {sideName: 'remote', type: 'DirAddition', doc: Metadata, was: Metadata}
 export type RemoteDirDeletion = {sideName: 'remote', type: 'DirDeletion', doc: Metadata}
 export type RemoteDirDissociation = {sideName: 'remote', type: 'DirDissociation', doc: Metadata, was: Metadata}
-export type RemoteDirMove = {sideName: 'remote', type: 'DirMove', doc: Metadata, was: Metadata, needRefetch?: true}
+export type RemoteDirMove = {sideName: 'remote', type: 'DirMove', doc: Metadata, was: Metadata, needRefetch?: true, descendantDirMoves?: RemoteDescendantChange[] }
```
sebn: Rename to `descendantMoves` since it can include files?
Or eventually to `childMoves` to be consistent with the existing Merge/Sync wording? (although descendant would definitely be more accurate in both cases)

sebn: Are they move-only or can file moves include content changes too?

sebn: > Are they move-only or can file moves include content changes too?

Answering myself: they can include content changes (hence the `previousMoveToSamePath()` change in the `RemoteWatcher`).

core/remote/watcher.js
```diff
@@ -234,7 +234,7 @@ class RemoteWatcher {
       }
       const previousMoveToSamePath = _.find(previousChanges, change =>
         // $FlowFixMe
-        change.type === 'FileMove' && change.doc.path === was.path)
+        (change.type === 'DescendantChange' || change.type === 'FileMove') && change.doc.path === was.path)
```
sebn: :ok_hand: Child file move + update handling

core/remote/change.js
```diff
@@ -20,12 +20,13 @@ export type RemoteFileUpdate = {sideName: 'remote', type: 'FileUpdate', doc: Met
...
 export type RemoteDirRestoration = {sideName: 'remote', type: 'DirRestoration', doc: Metadata, was: Metadata}
 export type RemoteDirTrashing = {sideName: 'remote', type: 'DirTrashing', doc: Metadata, was: Metadata}
 export type RemoteIgnoredChange = {sideName: 'remote', type: 'IgnoredChange', doc: Metadata|RemoteDoc|RemoteDeletion, detail: string}
 export type RemoteInvalidChange = {sideName: 'remote', type: 'InvalidChange', doc: *, error: Error}
 export type RemoteUpToDate = {sideName: 'remote', type: 'UpToDate', doc: Metadata, was: Metadata}
+export type RemoteDescendantChange = {sideName: 'remote', type: 'DescendantChange', doc: Metadata|RemoteDoc|RemoteDeletion, detail: string}
```
sebn: `{doc: Metadata, was: Metadata}`?
I assume it doesn't have to be different from the other move subtypes.
Also one can conclude the `details` from the `type` so they're not needed.

sebn: > Also one can conclude the details from the type so they're not needed.

Actually the type doesn't include the ancestor path, but I wonder whether it's important now that descendant/child moves are pushed into their ancestor. Maybe replace `details` with just `ancestorPath`?

core/remote/watcher.js
```diff
@@ -285,15 +285,18 @@ class RemoteWatcher {
...
+            const descendantChange = {
               sideName,
-              type: 'IgnoredChange',
+              type: 'DescendantChange',
+              update: change.update,
               doc,
               was,
               detail: `File was moved as descendant of ${_.get(previousChange, 'doc.path')}`
             }
+            previousChange.type === 'DirMove' && remoteChange.includeDescendant(previousChange, descendantChange)
```
sebn: Isn't `previousChange.type === 'DirMove'` already checked line 283?
Otherwise, detail above wouldn't make sense.
(I would assume it does i.e. this check is not needed)

core/remote/watcher.js
```diff
@@ -318,19 +321,23 @@ class RemoteWatcher {
             continue
           } else {
             _.assign(previousChange, {
-              type: 'IgnoredChange',
+              type: 'DescendantChange',
               detail: `Folder was moved as descendant of ${change.doc.path}`
             })
+            // $FlowFixMe
+            remoteChange.includeDescendant(change, previousChange)
```
sebn: Didn't we just lose `previousChange.descendantDirMoves`?

sebn: My bad, we didn't, it's just that flow doesn't see them anymore (still an issue though).

So I suppose we could either:

1. progressively flatten `*.descendantDirMoves` (by concatenating them in `includeDescendant()`
2. flatten them all at once in `RemoteWatcher#apply()`

core/remote/watcher.js
```diff
@@ -416,7 +424,22 @@ class RemoteWatcher {
         break
       case 'DirMove':
         log.info({path, oldpath: change.was.path}, 'folder was moved or renamed remotely')
-        await this.prep.moveFolderAsync(sideName, change.doc, change.was)
+        const needUpdates /*: any */ = []
+        const newRemoteRevs /*: RemoteRevisionsByID */ = {}
+        for (let descendant of (change.descendantDirMoves || [])) {
```
sebn: Here I believe we're missing the fact that each item of `descendantDirMoves` can actually have `descendantDirMoves` too (cf [proposals above](https://github.com/cozy-labs/cozy-desktop/pull/1250/files#r233131826)).

Or did I miss something?

core/remote/watcher.js
```diff
@@ -365,6 +372,7 @@ class RemoteWatcher {
       case 'InvalidChange':
         throw change.error
       case 'IgnoredChange':
+      case 'DescendantChange':
         log.debug({path, remoteId: change.doc._id}, change.detail)
```
sebn: Not sure the same info should be logged in both cases.

core/remote/watcher.js
```diff
@@ -318,19 +321,23 @@ class RemoteWatcher {
...
-          return {
+          const descendantChange = {
             sideName,
-            type: 'IgnoredChange',
+            type: 'DescendantChange',
             doc,
             was,
             detail: `Folder was moved as descendant of ${_.get(previousChange, 'doc.path')}`
           }
+          previousChange.type === 'DirMove' && remoteChange.includeDescendant(previousChange, descendantChange)
```
sebn: `if` block?

sebn: I know you love one-liners, but `if` blocks are probably more common in the current codebase ;)

aenario:
I splited this PR into your #1273 & #1274

Will include your suggestions there.

core/remote/watcher.js
aenario: Flow is :sparkles:

core/remote/watcher.js
aenario: Did :one: in #1274 / https://github.com/cozy-labs/cozy-desktop/pull/1274/commits/fc34fc7361795bac4348d7e6759748f82cefd78f

core/remote/change.js
aenario: renamed to `descendantMoves` in #1274 https://github.com/cozy-labs/cozy-desktop/pull/1274/commits/04a2fd8e0449c5179b4e3e17471887954863893b  

IMHO, being consistent with merge/sync would be more confusing.

core/remote/change.js
aenario: Done in #1274

PR #1252, opened 2018-11-07
Handle doctype mismatch on doTrash

nono:
The conflict resolution can't work here, as it will try to move a deleted file/folder.


sebn:
Codecov complains about [lower coverage on removed code](https://codecov.io/gh/cozy-labs/cozy-desktop/pull/1252/diff?src=pr&el=tree#diff-Y29yZS9wcmVwLmpz)... :man_facepalming:

PR #1253, opened 2018-11-07
Add linux atom watcher

nono:


core/local/atom_watcher.js
```diff
@@ -0,0 +1,86 @@
...
+  stop: (force: ?bool) => Promise<*>
+  ensureDirSync: () => void
+  */
+
+  constructor (syncPath /*: string */, prep /*: Prep */, pouch /*: Pouch */, events /*: EventEmitter */) {
+    this.syncPath = syncPath
+    this.prep = prep
+    this.pouch = pouch
+    this.events = events
+    this.checksumer = checksumer.init()
```
sebn: `this.prep`, `this.pouch` and `this.checksumer` don't seem to be used outside of the constructor.

core/local/atom_watcher.js
```diff
@@ -0,0 +1,86 @@
...
+
+const fse = require('fs-extra')
+const Promise = require('bluebird')
+
+const checksumer = require('./checksumer')
+const logger = require('../logger')
+const LinuxSource = require('./layers/linux')
+const Identifier = require('./layers/identifier')
+const ChecksumLayer = require('./layers/checksum')
+const Dispatcher = require('./layers/dispatcher')
```
sebn: `ChecksumLayer` has a suffix, others don't. Standardize?

core/local/atom_watcher.js
```diff
@@ -0,0 +1,86 @@
...
+
+/*::
+import type Pouch from '../pouch'
+import type Prep from '../prep'
+import type EventEmitter from 'events'
+import type { Checksumer } from './checksumer'
+import type { ChokidarEvent } from './chokidar_event'
```
sebn: Does it really need the `ChokidarEvent` type?

core/local/atom_watcher.js
nono: `this.checksumer` is used from the outside: https://github.com/cozy-labs/cozy-desktop/blob/master/core/local/index.js#L236

I've removed the two others

core/local/atom_watcher.js
nono: Well, I prefer to have no suffix. But we already have a lot of stuff called checksum and checksumer and I don't want to add confusion to that.

core/local/layers/sequential.js
```diff
@@ -0,0 +1,49 @@
+/* @flow */
+
+/*::
+import type { Layer, LayerEvent } from './events'
+*/
+
+// Sequential is a base class for layers that must preserve the orders of
+// initial and process calls. The subclasses must call the super constructor
+// with the next layer and overload the doProcess method.
+module.exports = class Sequential {
```
sebn: Shouldn't the Sequential class explicitely implement the Layer interface?

core/local/layers/checksum.js
```diff
@@ -0,0 +1,36 @@
...
+  }
+
+  async doProcess (events /*: LayerEvent[] */) {
+    for (const event of events) {
+      if (['add', 'update'].includes(event.action) && event.doc.docType === 'file') {
+        try {
+          event.doc.md5sum = await this.checksumer.push(event.abspath)
+        } catch (err) {
+          // TODO
+          console.error(err)
```
sebn: Logger?

nono: Well, for the moment, it's a TODO. I will see later what should be done here

core/local/layers/linux.js
```diff
@@ -0,0 +1,180 @@
...
+
+  async buildUpdateEvent (abspath /*: string */) /*: Promise<LayerUpdateEvent> */ {
+    let doc /*: ?Metadata */
+    const fpath = this.relativePath(abspath)
+    const stats = await fse.stat(abspath)
+    if (stats != null && stats.isDirectory()) {
+      doc = buildDir(fpath, stats)
+    } else {
+      doc = buildFile(fpath, stats, '')
+    }
```
sebn: These 8 lines are exactly the same as those in `#buildAddEvent()`

core/local/layers/linux.js
```diff
@@ -0,0 +1,180 @@
...
+  }
+
+  async buildAddEvent (abspath /*: string */) /*: Promise<LayerAddEvent> */ {
+    let doc /*: ?Metadata */
+    const fpath = this.relativePath(abspath)
+    const stats = await fse.stat(abspath)
+    if (stats != null && stats.isDirectory()) {
+      doc = buildDir(fpath, stats)
+    } else {
+      doc = buildFile(fpath, stats, '')
```
sebn: Although not clearly stated, I believe `buildFile()` was supposed to be used with a valid checksum.

core/local/layers/linux.js
```diff
@@ -0,0 +1,180 @@
...
+      src = buildDir(oldpath, new fs.Stats())
+    } else {
+      doc = buildFile(fpath, stats, '')
+      src = buildFile(oldpath, new fs.Stats(), '')
+    }
+    return { action: 'move', abspath, doc, src }
+  }
+
+  relativePath (abspath /*: string */) {
+    return path.relative(this.syncPath, abspath)
```
sebn: Will we have to compute relative paths on other platforms too?

core/local/layers/linux.js
```diff
@@ -0,0 +1,180 @@
...
+          batch.push(eDel)
+          const w = this.watchers.get(eDel.doc.path)
+          if (w) {
+            w.dispose()
+            this.watchers.delete(eDel.doc.path)
+          }
+          break
+        case 'renamed':
+          batch.push(await this.buildMoveEvent(event.path, event.oldPath))
+          break
```
sebn: I assume existing watchers automatically keep watching the new destination?

core/local/layers/events.js
```diff
@@ -0,0 +1,47 @@
+/* @flow */
+
+/*::
+import type { Metadata } from '../../metadata'
+
+export type AtomWatcherEvent = {
+  action: "created" | "modified" | "deleted" | "renamed",
+  kind: "file" | "directory" | "symlink" | "unknown",
+  path: string,
+  oldPath: string
```
sebn: Is `oldPath` mandatory for every AtomWatcherEvent subtype?

core/local/layers/linux.js
```diff
@@ -0,0 +1,180 @@
...
+        default:
+          throw new Error(`Unknown atom/watcher action ${event.action}`)
+      }
+    }
+    this.next.process(batch)
+  }
+
+  async buildAddEvent (abspath /*: string */) /*: Promise<LayerAddEvent> */ {
+    let doc /*: ?Metadata */
+    const fpath = this.relativePath(abspath)
```
sebn: `fpath` sounds a bit like `filepath` while it could actually point to a directory... What about `relpath`?

nono: The `f` from `fpath` could be used for `file` or `folder` ;-)

core/local/layers/events.js
nono: No, but I didn't find a way to make `flow` happy without that

core/local/layers/linux.js
nono: Yes, inotify watchers are attached to a folder and move with it.

core/local/layers/linux.js
nono: I think we will have to. Why?

core/local/layers/linux.js
nono: I believe so, but I'm not convinced that it's a good idea.

core/local/layers/linux.js
```diff
@@ -0,0 +1,180 @@
...
+  }
+
+  async buildMoveEvent (abspath /*: string */, oldpath /*: string */) /*: Promise<LayerMoveEvent> */ {
+    let doc /*: ?Metadata */
+    let src /*: ?Metadata */
+    const fpath = this.relativePath(abspath)
+    const stats = await fse.stat(abspath)
+    if (stats != null && stats.isDirectory()) {
+      doc = buildDir(fpath, stats)
+      src = buildDir(oldpath, new fs.Stats())
```
sebn: I'm not comfortable with the use of empty stats, the dir (or file) ends up with an invalid date, flow doesn't even complain, neither does the `Date` constructor, everything would just crash at some point later when calling whatever Date method...

I assume we end up fixing everything in another layer, but it still feels weird to me...

Also the more I think about it, the more I believe we're currently trying to build docs too early even in the current code, which leads to a lot of error-prone *fixing* afterwards...
Shouldn't we just build a data structure that represents the event identified by the watcher, then derive a doc from it later in merge?

sebn: Also the more I think about it, the more I believe we're currently trying to build docs too early even in the current code, which leads to a lot of error-prone *fixing* afterwards...
Shouldn't we just build a data structure that represents the event identified by the watcher, then derive a doc from it later in merge?

core/local/layers/linux.js
```diff
@@ -0,0 +1,180 @@
...
+
+  async buildRemoveEvent (abspath /*: string */, kind /*: string */) /*: Promise<LayerRemoveEvent> */ {
+    const fpath = this.relativePath(abspath)
+    let doc /*: ?Metadata */
+    if (kind === 'directory') {
+      doc = buildDir(fpath, new fs.Stats())
+    } else {
+      doc = buildFile(fpath, new fs.Stats(), '')
+    }
+    return { action: 'remove', abspath, doc }
```
sebn: Previously we had `action: 'deleted'`, now we have also `action: 'remove'`... Isn't it a bit error-prone?

core/local/layers/linux.js
nono: I feel the pain, but I don't have a good solution for now.

> Shouldn't we just build a data structure that represents the event identified by the watcher, then derive a doc from it later in merge?

I don't see what it will change. This data structure that represents the event will still be build step by step, and we will still juggle to be sure it has the expected properties. It's just moving the issue to another structure, no?

Flow seems too weak for our step by step building of events (or maybe I don't understand it enough).

sebn: We could pass the original atom-watcher event to the various `#build*Event()` methods without having to destructure it to later restructure it into something almost similar.
From what I can see, the next layer basically needs the atom-watcher event + stats.
Everything else can actually be derived from those two.
So I find the current code a bit complicated regarding the purpose...
Or did I miss something?

sebn: TLDR: pass an AtomWatcherEvent to every `#build*Event()` method and pass a batch of `{atomWatcherEvent, relpath, stats}` (without any doc) to the next layer?

sebn: (I'm not sure a `#build*Event()` is actually needed for every action type)

nono: I could use a custom struct for the events in the layer, and transform them to something compatible with Merge in the last layer (Dispatcher), but this struct will still be build step by step. It just moves complexity, not fix it.

core/local/layers/linux.js
sebn: I mean in the original atom-watcher event.

core/local/atom_watcher.js
```diff
@@ -0,0 +1,81 @@
...
+  events: EventEmitter
+  checksumer: Checksumer
+  ensureDirInterval: *
+  running: Promise<void>
+  _runningResolve: ?Function
+  _runningReject: ?Function
+  source: LinuxSource
+  start: () => Promise<*>
+  stop: (force: ?bool) => Promise<*>
+  ensureDirSync: () => void
```
sebn: I believe the way the `Watcher` interface is declared is messing up flow (my bad, I didn't noticed it in the previous PR).
Following [flow interface syntax](https://flow.org/en/docs/types/interfaces/#toc-interface-syntax) for methods/props (aka no `:` after method name, `:` instead of `=>` for method return type, end every line with `;`) should allow removing the 3 lines above.

core/local/atom_watcher.js
```diff
@@ -0,0 +1,81 @@
...
+    this.syncPath = syncPath
+    this.events = events
+    this.checksumer = checksumer.init()
+
+    // TODO detect platform to build the correct chain of layers
+    // TODO do we need a debounce layer (a port of awaitWriteFinish of chokidar)?
+    const dispatcher = new Dispatcher(prep, pouch, events)
+    const checksum = new ChecksumLayer(dispatcher, this.checksumer)
+    const identifier = new Identifier(checksum)
+    this.source = new LinuxSource(syncPath, identifier)
```
sebn: I wonder about the *layer* abstraction...

- None of them actually accept/return exactly the same thing.
- Identifier and ChecksumLayer look like small single functions, while Linux and Dispatcher are huge.
- Do we actually need a full layer to assign ids while none of the next layers actually use it? Even so, shouldn't we just assign ids in `metadata.build*()`?
- Having to build the next Layer-compatible object to test any of them looks a bit complicated.
- All of them are mixing data, computation and side-effects, which doesn't make them easier to understand/test.

nono: I'm not really pleased with the layer abstraction. But I have a hard time to find better, and I still prefer this to the way the old watcher is organized. At least, each layer can be understood without to read other layers (even if it is a mess from a static types point of view).

> while Linux and Dispatcher are huge

They aren't huge, they are a lot smaller that the existing local analysis for moves.

core/local/layers/linux.js
nono: Well, there is a mismatch between the naming from atom/watcher and what we use elsewhere (like names of the method of Prep/Merge). I've chosen to stick with the later, but if you prefer the former, I don't really mind.

nono: Well, I still find that atom/watcher `renamed` that can mean `moved` is not terrible. But if you prefer atom/watcher naming, I can live with that.

core/local/atom_watcher.js
sebn: My apologies, *huge* was excessive.

sebn: Regarding the `Identifier` layer, don't docs already have an id since they're built using `buildFile` / `buildDir` in the `LinuxSource` layer?

nono: Yes, it's the case now. It wasn't true when I wrote this code for `buildDir`, but it was fixed since then to match the behavior of `buildFile`.

core/local/layers/linux.js
sebn: Could checksums be computed in a single async function doing a Promise.reduce of the file add/update event paths over a Map<Path,Md5sum> using the checksumer?
Could stats be computed the same way?
With those 2 maps and the original event, couldn't we build a full doc at once?

core/local/layers/linux.js
sebn: I believe Prep/Merge use `deleteFile` / `deleteFolder`, not *remove* ;)

core/local/layers/checksum.js
sebn: Logger in case we forget to fix the TODO? :)

PR #1254, opened 2018-11-07
core: Fix ignore rules CRLF issue

sebn:
- User-defined rules with CRLF inside were not loaded
- Default rules could not be loaded on Windows development workstation depending on git config


sebn:
Duplication is fixed in #1255

PR #1256, opened 2018-11-07
Try to make appveyor builds faster

nono:

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -1,10 +1,16 @@
+image: Visual Studio 2017
+
+clone_depth: 10
+
 environment:
-  nodejs_version: "8"
   MOCHA_TIMEOUT: "60000"
   NO_BREAKPOINTS: "1"
 
+cache:
+  - node_modules
+  - '%LOCALAPPDATA%/Yarn'
+
 install:
-  - cmd: appveyor-retry powershell Install-Product node $env:nodejs_version
   - cmd: appveyor-retry yarn install:all
   - cmd: appveyor-retry yarn bootstrap:remote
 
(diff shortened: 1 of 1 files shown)


.appveyor.yml
```diff
@@ -1,21 +1,27 @@
+image: Visual Studio 2017
+
+clone_depth: 10
```
sebn: :ok_hand: Didn't know of [shallow clones](https://git-scm.com/docs/git-clone#git-clone---depthltdepthgt), although I'm not sure to understand the implications...

Could this be an issue for a PR including more than 10 commits?

.appveyor.yml
```diff
@@ -1,21 +1,27 @@
...
+
 environment:
-  nodejs_version: "8"
   MOCHA_TIMEOUT: "60000"
   NO_BREAKPOINTS: "1"
 
+cache:
+  - node_modules
```
sebn: Should we specify [cache dependencies](https://www.appveyor.com/docs/build-cache/#cache-dependencies)?

.appveyor.yml
```diff
@@ -1,21 +1,27 @@
+image: Visual Studio 2017
+
+clone_depth: 10
+
 environment:
-  nodejs_version: "8"
```
sebn: :ok_hand: Node 8 is already the [default on all images including VS2017](https://www.appveyor.com/docs/windows-images-software/#node-js) (I'm just documenting my understanding so other reviewers don't have to go through the same process)

.appveyor.yml
nono: > Could this be an issue for a PR including more than 10 commits?

No, I don't think it could be an issue. Even if the PR has more than 10 commits, we are only interested by tests on the last one, and this one will be here.

And I think it's the default option on travis.

.appveyor.yml
nono: I'd say no. Even if `package.json` has evolved, it's faster to start from a slightly outdated cache than from scratch. And yarn knows how to manage this correctly.

PR #1259, opened 2018-11-08
tech: upgrade elm, elm-format and elm-test

taratatach:
Notable changes:
  - use Browser.element instead of Html.programWithFlags
  - use Time.Posix for timestamps (timestamp in milliseconds, UTC)
  - encode File type when using ports (updated field is Time.Posix which
  cannot be passed through)
  - use Main.init method to initialize the Elm application instead of
  the removed Main.embed method**
  - replace package sporto/erl with elm/url
  sporto/erl is not compatible with Elm v0.19.0 and recommands to use the
  elm/url package instead.
  This means rewriting the Onboarding.Address.correctAddress method to
  match the new API.

- [x] PR is not too big


dev/ci/dist.sh
```diff
@@ -2,8 +2,7 @@
 
 set -ex
 
-yarn install:electron
-yarn install:elm
+yarn install:all
```
nono: I think we should avoid `yarn install:all` here, as it executes again a `yarn install` that is already done before by travis: it slows down the build.

dev/ci/tests.sh
```diff
@@ -4,8 +4,7 @@ set -ex
 
 . "./dev/ci/start_xvfb.sh"
 
-yarn install:electron
-yarn install:elm
+yarn install:all
```
nono: Idem here

dev/ci/dist.sh
taratatach: It won't take more than a few seconds to detect there aren't any modules to install while it helps in avoiding to forget some install phase (or to remove one of them even though that shouldn't happen often).

PR #1263, opened 2018-11-09
feat: Check updates in-app

enguerran:
This PR aims to add a check update without stopping/starting the app.

The solution chosen is to check and download the updates when the user interacts with the application (registration to the `browser-window-focus' event); according to the tests, the update is fast and painless.


enguerran:
I cannot understand codecov report:
> core/remote/change.js | 78.68% <0%> (-3.82%) |  
core/remote/watcher.js | 60.35% <0%> (+0.63%) |  

This PR does not change any of these files.

enguerran:
I changed the implemented scenario:
- everytime an update is available, we ask the user if the update should be installed
- yes → download → update → restart
- no → postpone

should update

taratatach:
```json
{"name":"Cozy Desktop","hostname":"erwans-macbook-pro.home","pid":91217,"component":"GUI:autoupdater","level":30,"msg":"Downloading update from CozyDrive-3.10.3-mac.zip, CozyDrive-3.10.3.dmg","time":"2018-11-22T15:35:22.240Z","v":0}
{"name":"Cozy Desktop","hostname":"erwans-macbook-pro.home","pid":91217,"component":"GUI:autoupdater","level":40,"err":{"message":"Could not get code signature for running application","name":"Error","stack":"Error: Could not get code signature for running application\n    at Server.<anonymous> (/Applications/Cozy Drive.app/Contents/Resources/app.asar/node_modules/electron-updater/src/MacUpdater.ts:85:28)\n    at Object.onceWrapper (events.js:314:30)\n    at emitNone (events.js:105:13)\n    at Server.emit (events.js:207:7)\n    at emitListeningNT (net.js:1346:10)\n    at _combinedTickCallback (internal/process/next_tick.js:135:11)\n    at process._tickDomainCallback (internal/process/next_tick.js:218:9)"},"msg":"Could not get code signature for running application","time":"2018-11-22T15:35:22.253Z","v":0}
{"name":"Cozy Desktop","hostname":"erwans-macbook-pro.home","pid":91217,"component":"GUI:autoupdater","level":50,"msg":"Error: Error: Could not get code signature for running application\n    at Server.<anonymous> (/Applications/Cozy Drive.app/Contents/Resources/app.asar/node_modules/electron-updater/src/MacUpdater.ts:85:28)\n    at Object.onceWrapper (events.js:314:30)\n    at emitNone (events.js:105:13)\n    at Server.emit (events.js:207:7)\n    at emitListeningNT (net.js:1346:10)\n    at _combinedTickCallback (internal/process/next_tick.js:135:11)\n    at process._tickDomainCallback (internal/process/next_tick.js:218:9)","time":"2018-11-22T15:35:22.253Z","v":0}
{"name":"Cozy Desktop","hostname":"erwans-macbook-pro.home","pid":91217,"component":"GUI:autoupdater","level":50,"err":{"message":"Could not get code signature for running application","name":"Error","stack":"Error: Could not get code signature for running application\n    at Server.<anonymous> (/Applications/Cozy Drive.app/Contents/Resources/app.asar/node_modules/electron-updater/src/MacUpdater.ts:85:28)\n    at Object.onceWrapper (events.js:314:30)\n    at emitNone (events.js:105:13)\n    at Server.emit (events.js:207:7)\n    at emitListeningNT (net.js:1346:10)\n    at _combinedTickCallback (internal/process/next_tick.js:135:11)\n    at process._tickDomainCallback (internal/process/next_tick.js:218:9)"},"msg":"Error in auto-updater! ","time":"2018-11-22T15:35:22.253Z","v":0}
{"name":"Cozy Desktop","hostname":"erwans-macbook-pro.home","pid":91217,"component":"GUI/UPDATER","level":20,"msg":"hide","time":"2018-11-22T15:35:22.253Z","v":0}
```

gui/js/updater.window.js
```diff
@@ -34,22 +34,21 @@ module.exports = class UpdaterWM extends WindowManager {
...
+          ),
+          type: 'question',
+          buttons: ['Update', 'Cancel'].map(translate)
+        }) === 0
+
         if (shouldUpdate) {
           this.show()
         } else {
-          this.skipped = false
+          this.skipped = true
```
enguerran: @taratatach It is not clear what you want to achieve with this commit:


The previous code will not ask the user to update if `this.skipped` is `false`, and because there is a `setInterval` that call this function, we put `this.skipped` to `true` just before the function displays the message box to the user. If the user says it is ok to update, we show the `electron-updater` update window. If the user don't want to update, we set the flag to `false` because the next time the `setInterval` will call this function we want to display the message box.

taratatach: As I understand it, `this.skipped` should be true when we've skipped an update.
Knowing that I would set it to `true` only when the user has purposefully asked not to update (i.e. the user has skipped the update).

But I understand what the issue would be with the interval. Like I said during the stand-up, if the update process works without my commit, let's revert it.

taratatach: It might be a good thing to rename this variable though, as it can be misleading.

enguerran: As you can see in the current codebase:

We set `this.skipped` to `true` when we are _looking for updates…_:
https://github.com/cozy-labs/cozy-desktop/blob/38bb4ffd777542c3305f22c282c55209196b0177/gui/js/updater.window.js#L79-L92

And its value is used to condition the display of the update window:
https://github.com/cozy-labs/cozy-desktop/blob/38bb4ffd777542c3305f22c282c55209196b0177/gui/js/updater.window.js#L30-L35

So I will drop the commit 43cd97d4c6344d860f36eef30f436eef776f3037.

taratatach: In the first code snippet you pasted above, we skip the update if we reached a timeout when looking for an update. In this situation, we don't want to show the update window since we don't have any update.

This situation is different from the one we're talking about since we're really skipping the update.

When we ask the user if she wants to update or not, we probably don't need to change the value of `this.skipped` since we'll either be:
- not updating and, since its value will be `false` already, we'll show the confirm window again once the check interval is reached
- updating and its value won't matter since we'll stop and relaunch the app entirely

enguerran: `setInterval` will display the message box even if a message box is displayed. That is why we control the display with `this.skipped`.

gui/main.js
```diff
@@ -25,6 +25,8 @@ const {incompatibilitiesErrorMessage} = require('./js/incompatibilitiesmsg')
 const UserActionRequiredDialog = require('./js/components/UserActionRequiredDialog')
 const {app, Menu, Notification, ipcMain, dialog} = require('electron')
 
+const DAILY = 3600 * 24
```
taratatach: `setInterval` expects milliseconds so we should have:
```js
const DAILY = 1000 * 3600 * 24
```

PR #1267, opened 2018-11-12
Fix an intermittent test failure

nono:
`doc` was created before `was` in this test, and when the two dates don't have the same ISO string serialization (which happens rarely as they are very close), the test fails. By creating the documents in the good order, we can prevent that.


nono:

PR #1268, opened 2018-11-12
Fix install of har-validator

nono:
The version 5.1.2 of har-validator was unpublished on npm.
See https://github.com/ahmadnassri/node-har-validator/issues/112


nono:

enguerran:
Shall we update `package.json` with `yarn.lock`?

nono:
@enguerran only the direct dependencies are listed in `package.json`, not the dependencies of dependencies, and har-validator is not a direct dependency, so there is nothing to update in `package.json`.

PR #1269, opened 2018-11-12
Fix missing COZY_STACK_TOKEN on travis

nono:
When CouchDB is too slow to start, the generation of the COZY_STACK_TOKEN fails. With this commit, we will wait longer that CouchDB has started.


nono:

enguerran:
How can we be sure thant `sleep 8` is enough and not `sleep 5`?
Cannot we make multiple request until we get a positive response from couchDB?

nono:
@enguerran yes, it would be better. Feel free to open a PR to do that.

PR #1273, opened 2018-11-13
Fix scenario move_dir_and_update_subfile_remotely

aenario:
This is PR is split from #1250 which was unreadable.

### DIFF
--- a/core/merge.js
+++ b/core/merge.js
@@ -11,6 +11,7 @@ const {
   assignId,
   assignMaxDate,
   detectPlatformIncompatibilities,
+  isAtLeastUpToDate,
   isUpToDate,
   markSide,
   sameBinary,
@@ -216,7 +217,7 @@ class Merge {
         if (doc.size == null) { doc.size = file.size }
         if (doc.class == null) { doc.class = file.class }
         if (doc.mime == null) { doc.mime = file.mime }
-      } else if (!isUpToDate(side, file)) {
+      } else if (!isAtLeastUpToDate(side, file)) {
         await this.resolveConflictAsync(side, doc, file)
         if (side === 'remote') {
           // we just renamed the remote file as a conflict
--- a/core/metadata.js
+++ b/core/metadata.js
@@ -103,6 +103,7 @@ module.exports = {
   ensureValidChecksum,
   extractRevNumber,
   isUpToDate,
+  isAtLeastUpToDate,
   markAsUpToDate,
   sameFolder,
   sameFile,
@@ -294,6 +295,12 @@ function isUpToDate (side /*: SideName */, doc /*: Metadata */) {
   let currentRev = doc.sides[side] || 0
   let lastRev = extractRevNumber(doc)
   return currentRev === lastRev
+}
+
+function isAtLeastUpToDate (side /*: SideName */, doc /*: Metadata */) {
+  let currentRev = doc.sides[side] || 0
+  let lastRev = extractRevNumber(doc)
+  return currentRev >= lastRev
 }
 
 function markAsUpToDate (doc /*: Metadata */) {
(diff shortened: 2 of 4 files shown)


core/metadata.js
```diff
@@ -256,6 +257,12 @@ function isUpToDate (side /*: SideName */, doc /*: Metadata */) {
   return currentRev === lastRev
 }
 
+function isAtLeastUpToDate (side /*: SideName */, doc /*: Metadata */) {
+  let currentRev = doc.sides[side] || 0
+  let lastRev = extractRevNumber(doc)
+  return currentRev >= lastRev
```
sebn: When I change `isUpToDate()` to `return currentRev >= lastRev` on master, tests (unit/integration/scenarios) pass on my laptop. I pushed #1276 to see what happens on CI.

Actually, all the conflict integration tests except one pass on master without any change.

The only test that fails (`local applied first -> conflict`) is effectively fixed by the new dissociation code in Merge, but it doesn't need the `isUpToDate()` change either.

core/sync.js
```diff
@@ -297,6 +297,9 @@ class Sync {
       } else {
         await this.doMove(side, doc, from)
       }
+      if (!sameBinary(from, doc)) {
+        await side.overwriteFileAsync(doc, from)
+      }
```
sebn: It doesn't seem to fix the scenario on CI:
https://travis-ci.org/cozy-labs/cozy-desktop/jobs/454611472#L377

I tried every combo of `isUpToDate` / `isAtLeastUpToDate`, but none of them fix it either.
So I assume something else in #1250 was making it work.

core/metadata.js
aenario: :thinking: Either I got confused, or it has more interactions with the #1274 than i thought. Let's merge that one first and then we will recheck.

core/sync.js
aenario: Looks like it, let's merge #1274 and then we will see.

test/integration/conflict_resolution.js
```diff
@@ -52,6 +55,101 @@ describe('Conflict resolution', () => {
...
+        md5sum: await helpers.local.syncDir.checksum('concurrent-edited')
+      }))
+    }
+
+    const expectedTree = [
+      'concurrent-edited',
+      'concurrent-edited-conflict-...'
+    ]
+
+    it('local applied first -> conflict', async () => {
```
sebn: Couldn't *applied* be confused with `Sync#apply()`?
What about *merged*?

test/integration/conflict_resolution.js
```diff
@@ -52,6 +55,101 @@ describe('Conflict resolution', () => {
...
+      await simulateLocalUpdateEvent()
```
sebn: Idem, *simulateLocalUpdateMerge*?

core/merge.js
```diff
@@ -216,8 +217,16 @@ class Merge {
...
-      } else if (!isUpToDate(side, file)) {
-        return this.resolveConflictAsync(side, doc, file)
+      } else if (!isAtLeastUpToDate(side, file)) {
+        await this.resolveConflictAsync(side, doc, file)
+        if (side === 'remote') {
+          // we just renamed the remote file as a conflict
+          // the old file should be dissociated from the remote
+          delete file.remote
+          if (file.sides) delete file.sides.remote
+          await this.pouch.put(file)
```
sebn: Should we be doing this in other Merge methods too?

sebn:
Should we close this one in favor of #1285?

aenario:
This have one more commit than #1285, let's merge that one first and then rebase this one.

aenario:
Rebased on master after merging #1285 , let's wait for travis

core/sync.js
aenario: I opened a PR with only this fix to see if it's sufficient in https://github.com/cozy-labs/cozy-desktop/pull/1290/files

PR #1274, opened 2018-11-13
Manage descendant moves

aenario:
This PR is a split from #1250 which was getting unreadable.

It needs to include @sebn review from there.

### DIFF
--- a/core/remote/change.js
+++ b/core/remote/change.js
@@ -20,12 +20,13 @@ export type RemoteFileUpdate = {sideName: 'remote', type: 'FileUpdate', doc: Met
 export type RemoteDirAddition = {sideName: 'remote', type: 'DirAddition', doc: Metadata, was: Metadata}
 export type RemoteDirDeletion = {sideName: 'remote', type: 'DirDeletion', doc: Metadata}
 export type RemoteDirDissociation = {sideName: 'remote', type: 'DirDissociation', doc: Metadata, was: Metadata}
-export type RemoteDirMove = {sideName: 'remote', type: 'DirMove', doc: Metadata, was: Metadata, needRefetch?: true}
+export type RemoteDirMove = {sideName: 'remote', type: 'DirMove', doc: Metadata, was: Metadata, needRefetch?: true, descendantMoves?: RemoteDescendantChange[] }
 export type RemoteDirRestoration = {sideName: 'remote', type: 'DirRestoration', doc: Metadata, was: Metadata}
 export type RemoteDirTrashing = {sideName: 'remote', type: 'DirTrashing', doc: Metadata, was: Metadata}
 export type RemoteIgnoredChange = {sideName: 'remote', type: 'IgnoredChange', doc: Metadata|RemoteDoc|RemoteDeletion, detail: string}
 export type RemoteInvalidChange = {sideName: 'remote', type: 'InvalidChange', doc: *, error: Error}
 export type RemoteUpToDate = {sideName: 'remote', type: 'UpToDate', doc: Metadata, was: Metadata}
+export type RemoteDescendantChange = {sideName: 'remote', type: 'DescendantChange', doc: Metadata, was: Metadata, ancestorPath: string, descendantMoves?: RemoteDescendantChange[], update?: true}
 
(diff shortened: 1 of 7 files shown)


core/remote/change.js
```diff
@@ -121,6 +124,11 @@ const isTrash = (a /*: RemoteChange */) /*: boolean %checks */ => a.type === 'Di
 const isRestore = (a /*: RemoteChange */) /*: boolean %checks */ => a.type === 'DirRestoration' || a.type === 'FileRestoration'
 const isDissociate = (a /*: RemoteChange */) /*: boolean %checks */ => a.type === 'DirDissociation' || a.type === 'FileDissociation'
 
+function includeDescendant (parent /*: RemoteDirMove */, e /*: RemoteDescendantChange */) {
+  parent.descendantMoves = parent.descendantMoves || []
+  parent.descendantMoves.push(e, ...(e.descendantMoves || []))
```
sebn: Wouldn't concat be more efficient?
Maybe `delete e.descendantMoves` afterwards to prevent confusion when looking at the data?

core/remote/watcher.js
```diff
@@ -416,7 +427,22 @@ class RemoteWatcher {
...
+            needUpdates.push(descendant)
+          }
+          if (descendant.doc.remote) {
+            // $FlowFixMe
+            newRemoteRevs[descendant.doc.remote._id] = descendant.doc.remote._rev
+          }
+        }
+        await this.prep.moveFolderAsync(sideName, change.doc, change.was, newRemoteRevs)
+        for (let needUpdate of needUpdates) {
+          change.was = await this.pouch.byRemoteIdMaybeAsync(change.was.remote._id)
```
sebn: I'm always uncomfortable with the *needUpdate(s)* logic here an in a few other places...
In this case, can't we just assume that child moves with updates need a refetch in Merge, where we're already iterating on them?
Also this makes me believe even more that we should passe data structures representing the changes to Prep/Merge, not the final doc/was which we end up fixing everywhere...

sebn: Also I believe it's called `needRefetch` in other places, not `needUpdate`.

aenario: Agree, but not part of this PR

core/remote/change.js
aenario: `concat` is less efficient (constructing new arrays) and IMHO, less readable : 

`parent.descendantMoves = (parent.descendantMoves || []).concat(e).concat(e.descendantMoves || [])`

deleting further descendantMoves is loosing debug data, better to have them and not need them.

core/remote/watcher.js
aenario: `needRefetch` is not the same as `update`

sebn: Yep, but isn't `needUpdate` used exactly in the same way as `needRefetch` in `LocalWatcher`, aka reloading `old` / `was` from `Pouch` because some previously merged change of the same batch probably changed it?

core/remote/change.js
sebn: > deleting further descendantMoves is loosing debug data, better to have them and not need them.

Don't we already have `ancestorPath` for this?
Or do we believe it won't help debugging?

aenario: Added `delete e.descendantMoves`

core/remote/watcher.js
aenario: cleaned up

core/remote/watcher.js
```diff
@@ -285,15 +285,18 @@ class RemoteWatcher {
...
               sideName,
-              type: 'IgnoredChange',
+              type: 'DescendantChange',
+              update: change.update,
               doc,
               was,
-              detail: `File was moved as descendant of ${_.get(previousChange, 'doc.path')}`
+              ancestorPath: _.get(previousChange, 'doc.path')
             }
+            if (previousChange.type === 'DirMove') remoteChange.includeDescendant(previousChange, descendantChange)
```
sebn: Isn't the `if` check useless? (since already true from line 283)

aenario: Flow :sparkles:

PR #1275, opened 2018-11-13
Handle unapplied remote move - initialScan - sync

aenario:
This PR includes and fix the test from #1235

### DIFF
--- a/core/local/chokidar_event.js
+++ b/core/local/chokidar_event.js
@@ -19,7 +19,8 @@ export type ChokidarEvent =
 */
 
 module.exports = {
-  build
+  build,
+  pretendUnlinkFromMetadata
 }
 
 function build (type /*: string */, path /*: ?string */, stats /*: ?fs.Stats */) /*: ChokidarEvent */ {
@@ -27,4 +28,11 @@ function build (type /*: string */, path /*: ?string */, stats /*: ?fs.Stats */)
   if (path != null) event.path = path
   if (stats != null) event.stats = stats
   return event
+}
+
+function pretendUnlinkFromMetadata (doc /*: Metadata */) /*: ChokidarUnlink|ChokidarUnlinkDir */ {
+  const type = doc.docType === 'file' ? 'unlink' : 'unlinkDir'
+  const path = doc.path
+  // $FlowFixMe
+  return { type, path, old: doc }
 }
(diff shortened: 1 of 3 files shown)


core/local/chokidar_event.js
```diff
@@ -2,12 +2,13 @@
...
 import type fs from 'fs'
+import type {Metadata} from '../metadata'
 
 export type ChokidarAdd = {type: 'add', path: string, stats: fs.Stats}
 export type ChokidarAddDir = {type: 'addDir', path: string, stats: fs.Stats}
 export type ChokidarChange = {type: 'change', path: string, stats: fs.Stats}
-export type ChokidarUnlink = {type: 'unlink', path: string}
-export type ChokidarUnlinkDir = {type: 'unlinkDir', path: string}
+export type ChokidarUnlink = {type: 'unlink', path: string, old?: Metadata}
+export type ChokidarUnlinkDir = {type: 'unlinkDir', path: string, old?: Metadata}
```
sebn: Isn't this supposed to be what chokidar gives us? How come it haves *old* Metadata?

sebn: (ok answer in code below)

aenario:
Cleanup : rebased & removed unused function `pretendAddFromMetadata`

sebn:
I would have expected this to be handled at the Merge level, no in the watcher: It makes the watcher more complex, with more *made up* events (the real move events were already merged) and it will have to be reimplemented for the new AtomWatcher... Also I wonder whether it could happen on the remote side...

sebn:
As discussed with @nono, applying pending changes on restart before initial scan would be a bad idea (since we have no guarantee the local FS still matches the meta tree), handling it in the Merge would be far more complicated than the current solution (according to the current way the Merge works), so better keep it in the local watcher for now (even if it means reimplementing it for the new atom watcher).

So +1 to merge after release 3.10.3

aenario:
[Si un move est appliqué dans le pouch mais pas sur le FS (application interrompue), le redémmarrage de l'app crée un cas de conflits infinis](https://trello.com/c/g5ltJ01O/1191-si-un-move-est-appliqu%C3%A9-dans-le-pouch-mais-pas-sur-le-fs-application-interrompue-le-red%C3%A9mmarrage-de-lapp-cr%C3%A9e-un-cas-de-conflits)

PR #1276, opened 2018-11-13
[WIP] core/metadata: Assume up to date when current rev >= last one

sebn:
- Doesn't break existing tests
- May prevent unexpected conflicts

Courtesy of @aenario

### DIFF
--- a/core/metadata.js
+++ b/core/metadata.js
@@ -293,7 +293,7 @@ function extractRevNumber (doc /*: Metadata|{_rev: string} */) {
 function isUpToDate (side /*: SideName */, doc /*: Metadata */) {
   let currentRev = doc.sides[side] || 0
   let lastRev = extractRevNumber(doc)
-  return currentRev === lastRev
+  return currentRev >= lastRev
 }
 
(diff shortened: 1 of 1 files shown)


core/metadata.js
```diff
@@ -253,7 +253,7 @@ function extractRevNumber (doc /*: Metadata|{_rev: string} */) {
 function isUpToDate (side /*: SideName */, doc /*: Metadata */) {
   let currentRev = doc.sides[side] || 0
   let lastRev = extractRevNumber(doc)
-  return currentRev === lastRev
+  return currentRev >= lastRev
```
nono: The commit says the opposite of the code: _up to date when last rev >= current one_.

I think the code (`currentRev >= lastRev`) is right, and you should fix the commit message.

sebn: Please note the goal is not necessarily to merge this change, but just to see whether it breaks something (as part of discussion on #1250 and #1273)

sebn:
Not sure it's relevent, not tests showing it is... Closing.

PR #1277, opened 2018-11-14
core/merge: Synchronize childs of moved directory

taratatach:
When we move a directory during a synchronization phase, right after
  adding some new files to it, those files don't get uploaded while they
  remain in the renamed directory on the filesystem.

  We mark the source file as never synchronized and its destination as new
  so we don't try to move the source file and upload the destination
  file instead.

- [x] PR is not too big


taratatach:
The builds on mac OSX could not be pushed to bintray due to API usage limits. All tests passed though.

PR #1280, opened 2018-11-14
Update flow-bin

nono:


core/remote/cozy.js
```diff
@@ -219,7 +219,7 @@ class RemoteCozy {
   }
 
   // Retrieve the path of a remote file doc
-  async _setPath (doc /*: any */) /*: Promise<void> */ {
+  async _setPath (doc /*: * */) /*: Promise<*> */ {
```
taratatach: Can't we use `RemoteDoc` to type `doc`?

core/remote/document.js
```diff
@@ -91,7 +91,7 @@ export type JsonApiDoc = {
 }
 */
 
-function jsonApiToRemoteDoc (json/*: JsonApiDoc */) /*: RemoteDoc */ {
+function jsonApiToRemoteDoc (json/*: JsonApiDoc */) /*: * */ {
```
taratatach: What are we missing to type the return value?

core/remote/cozy.js
nono: Nope, it is called with things that looks like `RemoteDoc` but are not: they don't have a `path`, that is mandatory for `RemoteDoc`

core/app.js
```diff
@@ -140,12 +140,12 @@ class App {
...
-      if (err.code === 'ENOTFOUND') {
-        log.warn(`The DNS resolution for ${parsed.hostname} failed.`)
-        log.warn('Are you sure the domain is OK?')
-      } else if (err === 'Bad credentials') {
+      if (err === 'Bad credentials') {
         log.warn(err)
         log.warn('Are you sure there are no typo on the passphrase?')
+      } else if (err.code === 'ENOTFOUND') {
+        log.warn(`The DNS resolution for ${parsed.hostname} failed.`)
+        log.warn('Are you sure the domain is OK?')
```
sebn: Out of curiosity, why did the order have to change?

core/local/chokidar_event.js
```diff
@@ -2,12 +2,13 @@
...
 import type fs from 'fs'
+import type {Metadata} from '../metadata'
 
 export type ChokidarAdd = {type: 'add', path: string, stats: fs.Stats}
 export type ChokidarAddDir = {type: 'addDir', path: string, stats: fs.Stats}
 export type ChokidarChange = {type: 'change', path: string, stats: fs.Stats}
-export type ChokidarUnlink = {type: 'unlink', path: string}
-export type ChokidarUnlinkDir = {type: 'unlinkDir', path: string}
+export type ChokidarUnlink = {type: 'unlink', path: string, old?: Metadata}
+export type ChokidarUnlinkDir = {type: 'unlinkDir', path: string, old?: Metadata}
```
sebn: It's weird that you had to do the [exact same change](https://github.com/cozy-labs/cozy-desktop/pull/1275#discussion-diff-233448601L9) as @aenario in #1275

sebn: Ok, got it.

Could you add a FIXME explaining that the offline unlink events detection is abusing the ChokidarUnlink* event types which actually never have an old property?

core/app.js
nono: Flow complained that we can try to access `err.code` when `err` can be the string `'Bad credentials'`

core/local/chokidar_event.js
sebn: @nono Forget about it, I'll fix it in a separate PR.

nono: What is really weird is that flow didn't tell us about that before

sebn:
:+1:

core/remote/document.js
taratatach: Arf OK. I suppose this is the same issue as above.

PR #1288, opened 2018-11-16
appveyor: Split builds between 2 Cozy instances

taratatach:
Some tests can fail if they run concurrently on 2 builds and the same
  Cozy instance.
  We've created a second Cozy instance dedicated to the AppVeyor builds
  and we split those builds between the 2 to avoid these conflicts.

- [x] PR is not too big

### DIFF
--- a/dev/remote/generate-test-env.js
+++ b/dev/remote/generate-test-env.js
@@ -5,9 +5,17 @@ const fse = require('fs-extra')
 const pkg = require('../../package.json')
 const automatedRegistration = require('./automated_registration')
 
-const cozyUrl = process.env.COZY_URL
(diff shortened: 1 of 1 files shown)


dev/remote/generate-test-env.js
```diff
@@ -5,10 +5,18 @@ const fse = require('fs-extra')
 const pkg = require('../../package.json')
 const automatedRegistration = require('./automated_registration')
 
-const cozyUrl = process.env.COZY_URL
+const cozyUrl = chooseCozyUrl(process.env.APPVEYOR_BUILD_NUMBER)
```
sebn: This makes the script appveyor-dependent (while it was not necessarily).
But fixing AppVeyor has priority :)

PR #1290, opened 2018-11-19
core/sync: Applied updated file in move

aenario:
Splitted from #1273 to see if it works on its own.


aenario:
OK, this won't work. Closing in favor of #1273

PR #1302, opened 2018-11-23
Fix config copy

nono:
- on our current node version (electron 1.8.3 - node 8.2.1), fs.constants.COPYFILE_FICLONE is not available
- fs-extra.copySync takes a third argument that can be an options object or a callback, but not flags
- on node 8.5.0 and later, we can use a copy-on-write reflink via fs.copyFileSync with flags.

### DIFF
--- a/core/config.js
+++ b/core/config.js
@@ -1,5 +1,5 @@
-// FIXME: const autoBind = require('auto-bind')
-const fs = require('fs-extra')
+const fs = require('fs')
+const fse = require('fs-extra')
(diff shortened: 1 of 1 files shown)


nono:
[2 petits correctifs](https://trello.com/c/1AlmbDVA/1274-2-petits-correctifs)

taratatach:
I'm not sure I understand what we're fixing here. Is it a performance issue, hence the copy-on-write?

PR #1304, opened 2018-11-26
Fix disk-quota for Cozy instances with no quota

nono:


nono:
[2 petits correctifs](https://trello.com/c/1AlmbDVA/1274-2-petits-correctifs)

PR #1311, opened 2018-11-27
core/metadata: Null properties are same as absent

taratatach:
Some documents could be considered as different by the sameFile and
  sameFolder functions if one of their properties was either null or
  undefined on one document and absent (or just a different null value)
  on the other document.

  This could lead to the potential creation of conflicts since we
  mutate document properties and can create a property with a null value
  without changing anything else.

- [x] PR is not too big

### DIFF
--- a/core/metadata.js
+++ b/core/metadata.js
@@ -360,10 +360,14 @@ const makeComparator = (name, interestingFields) => {
       })
     })
   }
+  const canBeIgnoredDiff = (difference) => {
+    const diff = difference.item || difference
+    return _.isNil(diff.lhs) && _.isNil(diff.rhs)
+  }
   return (one, two) => {
     const diff = deepDiff(one, two, filter)
     log.trace({path: two.path, diff}, name)
-    return !diff
+    return !diff || _.every(diff, canBeIgnoredDiff)
   }
 }
 
--- a/test/unit/metadata.js
+++ b/test/unit/metadata.js
@@ -312,7 +312,7 @@ describe('metadata', function () {
     })
   })
 
-  describe('sameFolder', () =>
+  describe('sameFolder', () => {
     it('returns true if the folders are the same', function () {
       let a = {
         _id: 'FOO/BAR',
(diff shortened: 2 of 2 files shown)


test/unit/metadata.js
```diff
@@ -4,6 +4,7 @@ const _ = require('lodash')
 const fs = require('fs-extra')
 const should = require('should')
 const path = require('path')
+const deepDiff = require('deep-diff')
```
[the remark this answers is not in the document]
taratatach: Ah yes. It's not supposed to be there.

core/metadata.js
```diff
@@ -360,10 +360,14 @@ const makeComparator = (name, interestingFields) => {
       })
     })
   }
+  const isRealDiff = (difference) => {
+    const diff = difference.item ? difference.item : difference
```
nono: [nitpick] You prefer the ternary to `const diff = difference.item || difference`?

core/metadata.js
```diff
@@ -360,10 +360,14 @@ const makeComparator = (name, interestingFields) => {
       })
     })
   }
+  const isRealDiff = (difference) => {
```
nono: I'm not pleased by the name `isRealDiff`. As I understand, this function returns `false` when we have a "real" diff. Maybe `canBeIgnoredDiff`?

aenario: :+1: for new name and good to merge afterward

core/metadata.js
taratatach: It's a reflex and I was using `!!difference.item` in the condition until the linter complained.
I didn't think about changing the way I was writing this.

Your solution is shorter and can be implemented.

core/metadata.js
```diff
@@ -360,10 +360,14 @@ const makeComparator = (name, interestingFields) => {
...
+    const diff = difference.item ? difference.item : difference
+    return _.isNil(diff.lhs) && _.isNil(diff.rhs)
+  }
   return (one, two) => {
     const diff = deepDiff(one, two, filter)
     log.trace({path: two.path, diff}, name)
-    return !diff
+    return !diff || _.every(diff, isRealDiff)
```
sebn: I wonder whether replacing sameFile/sameFolder with two fileDiff/folderDiff functions returning either a diff object or nothing would make both the producer/consumer code simpler...

Although I wouldn't be comfortable doing this without more test coverage on Merge.

(also I wonder whether using deep-diff was such a good idea, my bad in case it wasn't, but this one is a smaller issue, out of the scope of this PR)

taratatach:
Waiting for https://github.com/cozy-labs/cozy-desktop/pull/1313.

core/metadata.js
taratatach: I feel like this is pretty much what we're doing already except we return a boolean instead of the diff itself.
As for using deep-diff, I think it's a nice tool even if some of their choices don't match with our needs. Those differences are quite small imho and don't justify moving away from it.

sebn: :ok_hand: out of the scope of this PR

PR #1313, opened 2018-11-27
test/unit/merge: Independent tests

sebn:
Pros:

- A test should not break because of another one
- No crazy filenames to prevent PouchDB conflicts


sebn:
Rebased on 3.11.0-beta.1

PR #1321, opened 2018-11-29
Avoid big call stacks on watchLoop

nono:


nono:
[Éviter de consommer trop de mémoire sur watchLoop](https://trello.com/c/Q9JxLQdu/1301-%C3%A9viter-de-consommer-trop-de-m%C3%A9moire-sur-watchloop)

PR #1323, opened 2018-11-29
Move elm tests to test/elm

nono:


nono:
Travis build failed because of the bintray rate limiting.
Appveyor build failed because of a connection error to the stack.

elm.json
```diff
@@ -1,7 +1,8 @@
 {
     "type": "application",
     "source-directories": [
-        "./gui/elm"
+        "./gui/elm",
+        "./test/elm"
```
sebn: Looks like it won't work without adding `./test/elm` to `source-directories`, even though you upgraded elm-test to 0.19.0-rev3 including [a fix](https://github.com/rtfeldman/node-test-runner/pull/315/commits/a80d718b7d101300808230c5029728d89e316b4a) which I was hoping to stop requiring us to do so, but looks like I misunderstood it.

I'm interested in case you saw something in the doc clearly stating that when using glob with non-standard location, you had to add it to `source-directories` (I find the doc is not very clear regarding non-standard uses).

Thanks anyway, let's merge and have our shell completion back :)

nono: Nope, I didn't find anything about that in the documentation

sebn: :ok_hand:

PR #1326, opened 2018-12-03
jq: Rename conflicts filter to conflict & keep props [ci skip]

sebn:
- Singular, same as most of our filters
- Keeping only path/time is easy to do by hand
- Doing it by default sometimes makes the filter useless


sebn:
(sorry I pushed to the wrong place :man_facepalming:)

PR #1328, opened 2018-12-03
core/change: Detect & mark wip compound moves

taratatach:
Mark move changes as wip when we're adding an add event to them and
  the underlying document does not exist anymore (happens in successive
  moves for example).

  This helps in managing a list of moves like A → B → C → B. The A → C
  change would fail because C does not exist anymore and thus we wouldn't
  aggregate the C → B move and instead treat it as an addition of B.

- [x] PR is not too big

### DIFF
--- a/core/local/change.js
+++ b/core/local/change.js
@@ -274,33 +274,26 @@ export type LocalMove = LocalFileMove|LocalDirMove
 export type LocalMoveEvent = LocalFileAdded|LocalDirAdded
 */
 
-function InvalidLocalMoveEvent (moveChange /*: LocalMove */, event /*: LocalMoveEvent */) {
-  this.name = 'InvalidLocalMoveEvent'
-  this.moveChange = moveChange
-  this.event = event
-  // FIXME: Include event/change details in message
-  this.message = `Cannot include event ${event.type} into change ${moveChange.type}`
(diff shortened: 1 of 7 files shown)


core/local/change.js
```diff
@@ -274,33 +274,26 @@ export type LocalMove = LocalFileMove|LocalDirMove
...
-  ensureValidMoveEvent(moveChange, e)
   moveChange.path = e.path
   moveChange.stats = e.stats
   moveChange.md5sum = e.md5sum
-  delete moveChange.wip
-  log.debug(
-    {path: e.path, oldpath: moveChange.old.path, ino: moveChange.stats.ino},
-    'FileMove + add = FileMove')
+
+  if (e.md5sum) {
```
sebn: Maybe we could later use the exact same `!e.wip` condition as in `includeAddDirEventInDirMove()` below.
The logic looks the same (possibly easier to read by reversing the if/else branches in both cases).
Not a blocker anyway :slightly_smiling_face:

taratatach: Yes I thought about the condition itself but I wanted the previous behavior to remain in the first branch. I'm not sure it's useful though :).

As for using `e.wip`, it would be nice but I wasn't sure it was always set at the time.

PR #1329, opened 2018-12-04
test: Fix Metadata size in all scenarios

sebn:
Hardcoding 0 was plain broken, triggering swallowed corrupt file errors.
The issue possibly didn't happen when file content was explicitely defined.

Also errors should definitely be explicit, but this will require more work
and should be fixed later in a separate commit.


sebn:
Other test types probably use test data builders and both remote/metadata builders compute the size from the given data.

taratatach:
Indeed, a simple search confirmed that.

PR #1332, opened 2018-12-06
Update revisions of renamed subdir of renamed dir

taratatach:
When renaming a sub-directory of a directory renamed in the same
  synchronisation loop, we missed two updates to the source directory:
  - one local update when we merge the renaming of the parent directory
  which triggers changes on its children (i.e. the local side and
  revision were updated and merging the sub-directory change would
  trigger a Pouch conflict)
  - one remote update when we synchronise the renaming of the parent
  directory with the remote Cozy and it does the thing that we do with
  the local Pouch document (we could then not update the remote
  sub-directory as its revision had changed)

  To avoid those issues, we refetch the local document right before
  merging the sub-directory renaming and the remote revision right
  before synchronising it.

- [x] PR is not too big

### DIFF
--- a/core/local/change.js
+++ b/core/local/change.js
@@ -301,7 +301,11 @@ function dirMoveFromUnlinkAdd (sameInodeChange /*: ?LocalChange */, e /*: LocalD
   const unlinkChange /*: ?LocalDirDeletion */ = maybeDeleteFolder(sameInodeChange)
   if (!unlinkChange) return
   if (_.get(unlinkChange, 'old.path') === e.path) return dirAddition(e)
-  log.debug({oldpath: unlinkChange.path, path: e.path}, 'unlinkDir + addDir = DirMove')
+  if (!e.wip) {
+    log.debug({oldpath: unlinkChange.path, path: e.path}, 'unlinkDir + addDir = DirMove')
+  } else {
+    log.debug({oldpath: unlinkChange.path, path: e.path}, 'unlinkDir + addDir wip = DirMove wip')
+  }
   return build('DirMove', e.path, {
     stats: e.stats,
     old: unlinkChange.old,
--- a/core/local/chokidar_watcher.js
+++ b/core/local/chokidar_watcher.js
@@ -307,7 +307,6 @@ module.exports = class LocalWatcher {
         if (c.needRefetch) {
           // $FlowFixMe
           c.old = await this.pouch.db.get(metadata.id(c.old.path))
-          c.old.childMove = false
         }
         switch (c.type) {
           // TODO: Inline old LocalWatcher methods
--- a/core/merge.js
+++ b/core/merge.js
@@ -339,6 +339,7 @@ class Merge {
   async moveFolderRecursivelyAsync (side /*: SideName */, folder /*: Metadata */, was /*: Metadata */, newRemoteRevs /*: ?RemoteRevisionsByID */) {
     log.debug({path: folder.path, oldpath: was.path}, 'moveFolderRecursivelyAsync')
     const docs = await this.pouch.byRecursivePathAsync(was._id)
+
     move(was, folder)
     let bulk = [was, folder]
 
--- a/core/merge.js
+++ b/core/merge.js
@@ -337,6 +337,7 @@ class Merge {
 
   // Move a folder and all the things inside it
   async moveFolderRecursivelyAsync (side /*: SideName */, folder /*: Metadata */, was /*: Metadata */, newRemoteRevs /*: ?RemoteRevisionsByID */) {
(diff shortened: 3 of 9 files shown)


core/local/analysis.js
```diff
@@ -273,8 +273,10 @@ function squashMoves (changes /*: LocalChange[] */) {
         log.debug({oldpath: b.old.path, path: b.path}, 'descendant move')
         a.wip = a.wip || b.wip
         if (b.path.substr(a.path.length) === b.old.path.substr(a.old.path.length)) {
+          log.debug({oldpath: b.old.path, path: b.path}, 'ignoring move')
           changes.splice(j--, 1)
         } else {
+          log.debug({oldpath: b.old.path, path: b.path}, 'move inside move')
           // move inside move
```
sebn: Drop comment?

core/local/analysis.js
```diff
@@ -273,8 +273,10 @@ function squashMoves (changes /*: LocalChange[] */) {
         log.debug({oldpath: b.old.path, path: b.path}, 'descendant move')
         a.wip = a.wip || b.wip
         if (b.path.substr(a.path.length) === b.old.path.substr(a.old.path.length)) {
+          log.debug({oldpath: b.old.path, path: b.path}, 'ignoring move')
```
sebn: Maybe explain why in log message?

core/local/chokidar_watcher.js
```diff
@@ -326,12 +326,14 @@ module.exports = class LocalWatcher {
...
             if (c.needRefetch) {
               c.old = await this.pouch.db.get(metadata.id(c.old.path))
-              c.old.childMove = false
             }
             await this.onMoveFile(c.path, c.stats, c.md5sum, c.old, c.overwrite)
             if (c.update) await this.onChange(c.update.path, c.update.stats, c.update.md5sum)
             break
           case 'DirMove':
+            if (c.needRefetch) {
+              c.old = await this.pouch.db.get(metadata.id(c.old.path))
```
sebn: Oh, the root of the 2nd move didn't had its old metadata refreshed after applying the first one :ok_hand:

core/merge.js
```diff
@@ -358,7 +358,11 @@ class Merge {
 
   // Move a folder and all the things inside it
   async moveFolderRecursivelyAsync (side /*: SideName */, folder /*: Metadata */, was /*: Metadata */, newRemoteRevs /*: ?RemoteRevisionsByID */) {
+    log.debug({path: folder.path, oldpath: was.path}, 'moveFolderRecursivelyAsync')
     const docs = await this.pouch.byRecursivePathAsync(was._id)
+
+    if (was.moveFrom && was.moveFrom.childMove) was.needNewRev = true
```
sebn: Hmm... The fact that you feel the need to depend `moveFrom.childMove` + introduce a new `Metadata` property tells us that we (not you @taratatach :wink:) didn't make the right choice when introducing `#childMove` IMHO.

core/local/change.js
```diff
@@ -170,7 +170,11 @@ function fileMoveFromUnlinkAdd (unlinkChange /*: LocalFileDeletion */, e /*: Loc
 
 function dirMoveFromUnlinkAdd (unlinkChange /*: LocalDirDeletion */, e /*: LocalDirAdded */) /*: * */ {
   if (_.get(unlinkChange, 'old.path') === e.path) return fromEvent(e)
-  log.debug({oldpath: unlinkChange.path, path: e.path}, 'unlinkDir + addDir = DirMove')
+  if (!e.wip) {
+    log.debug({oldpath: unlinkChange.path, path: e.path}, 'unlinkDir + addDir = DirMove')
+  } else {
+    log.debug({oldpath: unlinkChange.path, path: e.path}, 'unlinkDir + addDir wip = DirMove wip')
+  }
```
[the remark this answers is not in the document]
taratatach: I wanted to express in the message that the `wip` attribute comes from the event and not the previous change to make sure I understand correctly the flow of events.
Do you feel it's too much information?

core/local/analysis.js
taratatach: I actually didn't spend time to try and understand why we're ignoring it.
I'll do this work.

core/merge.js
taratatach: That's one of the questions I asked @aenario when we paired on this one. I'm not sure I completely understand the need for the child move and its gain compared to adding a bunch of moves to the list of changes.

core/local/chokidar_watcher.js
taratatach: Yes, since child moves are created when merging the parent dir move, we're passed the prep phase of the second move.

PR #1333, opened 2018-12-07
fix: Use a RegExp to rename conflicting files

enguerran:
The function were extracted to metadata.js,
then covered by tests.

- [x] PR is not too big

### DIFF
--- a/core/merge.js
+++ b/core/merge.js
@@ -3,24 +3,11 @@
 const autoBind = require('auto-bind')
 const _ = require('lodash')
 const { clone } = _
-const { dirname } = require('path')
+const path = require('path')
 
 const IdConflict = require('./IdConflict')
 const logger = require('./logger')
-const {
-  assignMaxDate,
-  detectPlatformIncompatibilities,
-  isAtLeastUpToDate,
-  isUpToDate,
-  markAsNeverSynced,
-  markAsNew,
-  markSide,
-  sameBinary,
-  sameFile,
-  sameFolder,
-  upToDate,
-  createConflictingDoc
-} = require('./metadata')
+const metadata = require('./metadata')
 const move = require('./move')
 const { otherSide } = require('./side')
 
@@ -71,11 +58,11 @@ class Merge {
   // Be sure that the tree structure for the given path exists
   async ensureParentExistAsync (side /*: SideName */, doc /*: * */) {
     log.trace({path: doc.path}, 'ensureParentExistAsync')
-    let parentId = dirname(doc._id)
+    let parentId = path.dirname(doc._id)
     if (parentId === '.') { return }
 
     // BUG on windows with incompatible names like "D:IR"
-    if (dirname(parentId) === parentId) { return }
+    if (path.dirname(parentId) === parentId) { return }
 
     try {
       let folder = await this.pouch.db.get(parentId)
@@ -86,7 +73,7 @@ class Merge {
 
     let parentDoc = {
       _id: parentId,
-      path: dirname(doc.path),
+      path: path.dirname(doc.path),
       docType: 'folder',
       updated_at: new Date()
     }
@@ -104,7 +91,7 @@ class Merge {
   // A suffix composed of -conflict- and the date is added to the path.
   async resolveConflictAsync (side /*: SideName */, doc /*: Metadata */, was /*: ?Metadata */) {
     log.warn({path: doc.path, oldpath: was && was.path, doc, was}, 'resolveConflictAsync')
-    let dst = createConflictingDoc(doc)
+    const dst = metadata.createConflictingDoc(doc)
     try {
       // $FlowFixMe
       await this[side].renameConflictingDocAsync(doc, dst.path)
@@ -122,7 +109,7 @@ class Merge {
     log.debug({path: doc.path}, 'addFileAsync')
     const {path} = doc
     const file /*: ?Metadata */ = await this.pouch.byIdMaybeAsync(doc._id)
-    markSide(side, doc, file)
+    metadata.markSide(side, doc, file)
     const idConflict /*: ?IdConflictInfo */ = IdConflict.detect(side, doc, file)
     if (idConflict) {
       log.warn({idConflict}, IdConflict.description(idConflict))
@@ -131,8 +118,8 @@ class Merge {
     } else if (file && file.docType === 'folder') {
       return this.resolveConflictAsync(side, doc, file)
     }
-    assignMaxDate(doc, file)
-    if (file && sameBinary(file, doc)) {
+    metadata.assignMaxDate(doc, file)
+    if (file && metadata.sameBinary(file, doc)) {
       doc._rev = file._rev
       if (doc.size == null) { doc.size = file.size }
       if (doc.class == null) { doc.class = file.class }
@@ -140,7 +127,7 @@ class Merge {
       if (doc.tags == null) { doc.tags = file.tags || [] }
       if (doc.remote == null) { doc.remote = file.remote }
       if (doc.ino == null) { doc.ino = file.ino }
-      if (sameFile(file, doc)) {
+      if (metadata.sameFile(file, doc)) {
         log.info({path}, 'up to date')
         return null
       } else {
@@ -189,22 +176,22 @@ class Merge {
     log.debug({path: doc.path}, 'updateFileAsync')
     const {path} = doc
     const file /*: ?Metadata */ = await this.pouch.byIdMaybeAsync(doc._id)
-    markSide(side, doc, file)
+    metadata.markSide(side, doc, file)
     if (file && file.docType === 'folder') {
       throw new Error("Can't resolve this conflict!")
     }
-    assignMaxDate(doc, file)
+    metadata.assignMaxDate(doc, file)
     if (file) {
       doc._rev = file._rev
       doc.moveFrom = file.moveFrom
       if (doc.tags == null) { doc.tags = file.tags || [] }
       if (doc.remote == null) { doc.remote = file.remote }
       if (doc.ino == null) { doc.ino = file.ino }
-      if (sameBinary(file, doc)) {
+      if (metadata.sameBinary(file, doc)) {
         if (doc.size == null) { doc.size = file.size }
         if (doc.class == null) { doc.class = file.class }
         if (doc.mime == null) { doc.mime = file.mime }
-      } else if (!isAtLeastUpToDate(side, file)) {
+      } else if (!metadata.isAtLeastUpToDate(side, file)) {
         await this.resolveConflictAsync(side, doc, file)
         if (side === 'remote') {
           // we just renamed the remote file as a conflict
@@ -215,7 +202,7 @@ class Merge {
         }
         return
       }
-      if (sameFile(file, doc)) {
+      if (metadata.sameFile(file, doc)) {
         log.info({path}, 'up to date')
         return null
       } else {
@@ -232,11 +219,11 @@ class Merge {
     log.debug({path: doc.path}, 'putFolderAsync')
     const {path} = doc
     const folder /*: ?Metadata */ = await this.pouch.byIdMaybeAsync(doc._id)
-    markSide(side, doc, folder)
+    metadata.markSide(side, doc, folder)
     if (folder && folder.docType === 'file') {
       return this.resolveConflictAsync(side, doc, folder)
     }
-    assignMaxDate(doc, folder)
+    metadata.assignMaxDate(doc, folder)
     const idConflict /*: ?IdConflictInfo */ = IdConflict.detect(side, doc, folder)
     if (idConflict) {
       log.warn({idConflict}, IdConflict.description(idConflict))
@@ -247,7 +234,7 @@ class Merge {
       if (doc.tags == null) { doc.tags = folder.tags || [] }
       if (doc.remote == null) { doc.remote = folder.remote }
       if (doc.ino == null && folder.ino) { doc.ino = folder.ino }
-      if (sameFolder(folder, doc)) {
+      if (metadata.sameFolder(folder, doc)) {
         log.info({path}, 'up to date')
         return null
       } else {
@@ -264,7 +251,7 @@ class Merge {
     log.debug({path: doc.path, oldpath: was.path}, 'moveFileAsync')
     const {path} = doc
     if (was.sides && !was.sides[otherSide(side)]) {
-      await this.pouch.remove(upToDate(was))
+      await this.pouch.remove(metadata.upToDate(was))
       return this.addFileAsync(side, doc)
     } else if (was.sides && was.sides[side]) {
       const file /*: ?Metadata */ = await this.pouch.byIdMaybeAsync(doc._id)
@@ -274,16 +261,16 @@ class Merge {
         await this.resolveConflictAsync(side, doc, file)
         return
       }
-      markSide(side, doc, file)
-      markSide(side, was, was)
-      assignMaxDate(doc, was)
+      metadata.markSide(side, doc, file)
+      metadata.markSide(side, was, was)
+      metadata.assignMaxDate(doc, was)
       if (doc.size == null) { doc.size = was.size }
       if (doc.class == null) { doc.class = was.class }
       if (doc.mime == null) { doc.mime = was.mime }
       if (doc.tags == null) { doc.tags = was.tags || [] }
       if (doc.ino == null) { doc.ino = was.ino }
       move(was, doc)
-      if (file && sameFile(file, doc)) {
+      if (file && metadata.sameFile(file, doc)) {
         log.info({path}, 'up to date (move)')
         return null
       } else if (file && !doc.overwrite && doc.path === file.path) {
@@ -310,9 +297,9 @@ class Merge {
     }
 
     const folder /*: ?Metadata */ = await this.pouch.byIdMaybeAsync(doc._id)
-    markSide(side, doc, folder)
-    markSide(side, was, was)
-    assignMaxDate(doc, was)
+    metadata.markSide(side, doc, folder)
+    metadata.markSide(side, was, was)
+    metadata.assignMaxDate(doc, was)
     if (doc.tags == null) { doc.tags = was.tags || [] }
     if (doc.ino == null) { doc.ino = was.ino }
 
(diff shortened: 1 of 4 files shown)


test/unit/metadata.js
```diff
@@ -772,4 +773,45 @@ describe('metadata', function () {
...
+      const pathRegExp = RegExp(`(${doc.path})-conflict-\\d{4}(?:-\\d{2}){2}T(?:\\d{2}_?){3}.\\d{3}Z`)
+      should(pathRegExp.exec(newDoc.path)).not.null()
+    })
+    it('should get the correct _id', () => {
+      const doc = {
+        path: 'docname'
+      }
+      const newDoc = createConflictingDoc(doc)
+      const pathRegExp = RegExp(`(${doc.path})-conflict-\\d{4}(?:-\\d{2}){2}T(?:\\d{2}_?){3}.\\d{3}Z`)
+      should(pathRegExp.exec(newDoc._id)).not.null()
```
sebn: Failure message could be more helpful with `should(newDoc._id).match(someRegExp)`

sebn: Although this test will pass on Linux, `newDoc._id` will match `/-CONFLICT-/` on macOS/Windows, not `/-conflict-/`, because ids are uppercase on those platforms, because we assume users are using the corresponding default filesystems (HFS+/APFS/NTFS) which are not case sensitive.

sebn: (see https://github.com/cozy-labs/cozy-desktop/blob/v3.11.1-beta.1/core/metadata.js#L157-L191 )

test/unit/metadata.js
```diff
@@ -772,4 +773,45 @@ describe('metadata', function () {
...
+    })
+    it('should keep the correct extension', () => {
+      const doc = {
+        path: 'docname.pdf'
```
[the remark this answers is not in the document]
enguerran: I'll add a test to cover this, you're right.
But `.tar.gz` is in fact a `.gz` of ` .tar`, so it should not be considered as a multiple parts extension. But we have to be aware that the name of the file is the string before `.tar.gz`.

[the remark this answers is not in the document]
enguerran: BTW the tools used come from `path` (see https://nodejs.org/api/path.html#path_path_extname_path)

enguerran:
This PR is based on #1335 and have 3 commits in common, I'll wait for those commits to be in `master` and rebase this PR on it.

Be my guest to review the remaining commits.

sebn:
@enguerran #1335 was merged, feel free to rebase!
Also flow is unhappy about the way you mix `module.exports.whatever = ...` with existing `module.exports = { ... }` assignment in `core/metadata` :wink:

core/metadata.js
```diff
@@ -456,3 +459,34 @@ function buildFile (filePath /*: string */, stats /*: fs.Stats */, md5sum /*: st
   if ((stats.mode & EXECUTABLE_MASK) !== 0) { doc.executable = true }
   return doc
 }
+
+const CONFLICT_REGEXP = new RegExp('-conflict-\\d{4}(?:-\\d{2}){2}T(?:\\d{2}_?){3}.\\d{3}Z')
+module.exports.CONFLICT_REGEXP = CONFLICT_REGEXP
```
sebn: Flow is unhappy about the way you mix `module.exports.whatever = ...` with existing `module.exports = { ... }` assignment in `core/metadata` :wink:

enguerran: I changed the export at the last minute: export a function rather than a RegExp object.

enguerran:
@sebn @taratatach I don't plan to add other changes to this PR, feel free to review.

core/metadata.js
```diff
@@ -248,11 +251,11 @@ export type PlatformIncompatibility = PathIssue & {docType: string}
 // TODO: return null instead of an empty array when no issue was found?
 function detectPlatformIncompatibilities (metadata /*: Metadata */, syncPath /*: string */) /*: Array<PlatformIncompatibility> */ {
   const {path, docType} = metadata
-  const pathLenghIssue = detectPathLengthIssue(join(syncPath, path), platform)
+  const pathLenghIssue = detectPathLengthIssue(nodepath.join(syncPath, path), platform)
```
sebn: What about dropping the `const {path, docType} = metadata` statement above and use `metadata.path` and `metadata.docType` explicitely? Wouldn't the change be smaller and not requiring a different `nodepath` module alias from other places?

core/metadata.js
```diff
@@ -4,14 +4,15 @@ const _ = require('lodash')
...
-const { join } = path
+const nodepath = require('path')
 
 const logger = require('./logger')
 const { detectPathIssues, detectPathLengthIssue } = require('./path_restrictions')
 const { DIR_TYPE, FILE_TYPE } = require('./remote/constants')
 const { maxDate } = require('./timestamp')
 
+const fsutils = require('./utils/fs')
+const { extname, dirname, basename } = require('path')
```
sebn: Out of curiosity, why not `const { ... } = nodepath`?
Do you prefer reusing `require`'s cache to make both require statements independent?

core/metadata.js
```diff
@@ -248,11 +251,11 @@ export type PlatformIncompatibility = PathIssue & {docType: string}
 // TODO: return null instead of an empty array when no issue was found?
 function detectPlatformIncompatibilities (metadata /*: Metadata */, syncPath /*: string */) /*: Array<PlatformIncompatibility> */ {
   const {path, docType} = metadata
-  const pathLenghIssue = detectPathLengthIssue(join(syncPath, path), platform)
+  const pathLenghIssue = detectPathLengthIssue(nodepath.join(syncPath, path), platform)
   const issues /*: PathIssue[] */ = detectPathIssues(path, docType)
   if (pathLenghIssue) issues.unshift(pathLenghIssue)
   return issues.map(issue => (_.merge({
-    docType: issue.path === path ? docType : 'folder'
+    docType: issue.path === nodepath ? docType : 'folder'
```
sebn: - [ ] Automated refactoring introduced a bug here :see_no_evil:

core/metadata.js
```diff
@@ -456,3 +459,33 @@ function buildFile (filePath /*: string */, stats /*: fs.Stats */, md5sum /*: st
...
+
+function withConflictRegExp (prefix /*: string */) /*: RegExp */ {
+  return new RegExp(`${prefix}${CONFLICT_REGEXP.source}`)
+}
+
+/**
+ *
+ * @param {Metadata} doc
+ *
+ * @returns a clone of `doc` with new `path` and `_id` properties
```
sebn: Maybe explain that it returns a copy of the given doc with a suffix conflict, eventualy shortening the base name so the final result doesn't clash with FS limitations?

core/metadata.js
```diff
@@ -456,3 +459,33 @@ function buildFile (filePath /*: string */, stats /*: fs.Stats */, md5sum /*: st
...
+
+const CONFLICT_REGEXP = new RegExp('-conflict-\\d{4}(?:-\\d{2}){2}T(?:\\d{2}_?){3}.\\d{3}Z')
+
+function withConflictRegExp (prefix /*: string */) /*: RegExp */ {
+  return new RegExp(`${prefix}${CONFLICT_REGEXP.source}`)
+}
+
+/**
+ *
+ * @param {Metadata} doc
```
sebn: Is it necessary?
Do you plan to introduce some tag-based documentation generator? :wink:

test/unit/metadata.js
```diff
@@ -772,4 +774,67 @@ describe('metadata', function () {
       should(upToDate(doc).errors).be.undefined()
     })
   })
+
+  describe('generated a doc with conflict', () => {
```
sebn: Maybe use the function name as the description, same as in other unit tests?

test/unit/metadata.js
```diff
@@ -772,4 +774,67 @@ describe('metadata', function () {
...
+    it('should get the correct path', () => {
+      const doc = {
+        path: 'docname'
```
sebn: Why not use `builders.whatever().path('docname').build()`?

core/metadata.js
```diff
@@ -456,3 +459,33 @@ function buildFile (filePath /*: string */, stats /*: fs.Stats */, md5sum /*: st
   if ((stats.mode & EXECUTABLE_MASK) !== 0) { doc.executable = true }
   return doc
 }
+
+const CONFLICT_REGEXP = new RegExp('-conflict-\\d{4}(?:-\\d{2}){2}T(?:\\d{2}_?){3}.\\d{3}Z')
+
+function withConflictRegExp (prefix /*: string */) /*: RegExp */ {
```
sebn: Looking at the function signature, one may expect the `prefix` arg (with `string` type) to be the *main subject*. But then the type of a prefix *with* a regexp doesn't seem obvious...

What about `prefixedConflictRegExp(prefix)`?

test/unit/metadata.js
```diff
@@ -772,4 +774,67 @@ describe('metadata', function () {
...
+      const newDoc = createConflictingDoc(doc)
+      const newDocBasename = withConflictRegExp('(.*)').exec(path.basename(newDoc.path))[1]
+      should(newDocBasename.length).equal(180)
+    })
+    it('should have an id', () => {
+      const doc = {
+        path: 'docname'
+      }
+      const newDoc = createConflictingDoc(doc)
+      should(newDoc).have.property('_id').not.null()
```
sebn: Isn't it already tested in test: *should get the correct _id*?

test/unit/metadata.js
```diff
@@ -772,4 +774,67 @@ describe('metadata', function () {
       should(upToDate(doc).errors).be.undefined()
     })
   })
+
+  describe('generated a doc with conflict', () => {
+    it('should get the correct path', () => {
```
sebn: Maybe we don't need to start every test description with *should*?

test/unit/metadata.js
```diff
@@ -772,4 +774,67 @@ describe('metadata', function () {
...
+      const newDoc = createConflictingDoc(doc)
+      should(path.extname(newDoc.path)).equal(ext)
+    })
+    it('should keep the correct extension when extension is complex', () => {
+      const ext = '.tar.gz'
+      const doc = {
+        path: `docname${ext}`
+      }
+      const newDoc = createConflictingDoc(doc)
+      // FIXME: must be docname-conflict-:ISODATE:.tar.gz instead of docname.tar-conflict-:ISODATE:.gz
```
sebn: Maybe we could describe the current behavior in the test description, so one doesn't believe it actually works looking at the test results?

test/unit/metadata.js
```diff
@@ -772,4 +774,67 @@ describe('metadata', function () {
...
+      should(path.extname(newDoc.path)).equal('.gz')
+    })
+    it('should not have more than 180 characters', () => {
+      const doc = {
+        path: 'Lorem ipsum dolor sit amet consectetur adipiscing elit Nam a velit at dolor euismod tincidunt sit amet id ante Cras vehicula lectus purus In lobortis risus lectus vitae rhoncus quam porta nullam'
+      }
+      const newDoc = createConflictingDoc(doc)
+      const newDocBasename = withConflictRegExp('(.*)').exec(path.basename(newDoc.path))[1]
```
sebn: Looks a bit like lines 482 to 484 in the implementation?

sebn:
Oops, missed the review description, sorry for this @enguerran 

Main blocking point is [here](https://github.com/cozy-labs/cozy-desktop/pull/1333#discussion_r240506343)

test/unit/metadata.js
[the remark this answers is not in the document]
taratatach: I know about the current behavior. Imho this is not ideal to separate `.tar` from `.gz` with a conflict « suffix » as part of the extension could be removed by the user when trying to fix those conflicts.

Adding a test would explicit the behavior we're looking for and then we could fix this as part of this PR.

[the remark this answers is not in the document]
taratatach: As for the rule, this is imho the core issue here. I don't have a solution to offer but can we list all extensions or figure out the extension from the MIME type?

core/merge.js
```diff
@@ -105,21 +104,7 @@ class Merge {
...
-    if (base.length > 180) {
-      base = base.slice(0, 180)
-    }
-    // avoid long chain of -conflict-DATE-conflict-DATE-conflict-DATE
-    // TODO unit-test and pick behaviour for files like
-    //                "cat-conflict-20181116...-THE_GOOD_ONE.jpg"
-    base = base.split('-conflict-20')[0]
-    dst.path = `${join(dir, base)}-conflict-${date}${ext}`
-    assignId(dst)
+    let dst = createConflictingDoc(doc)
```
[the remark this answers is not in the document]
enguerran: The code were written with `let` and I was not aware of this rule.

core/metadata.js
enguerran: Why not

core/metadata.js
enguerran: I don't have any preferences, I use the same rule as 5 lines above.

core/metadata.js
enguerran: No I don't, sorry

test/unit/metadata.js
enguerran: Why add more coupling? What if the function name changes?

test/unit/metadata.js
enguerran: To do what exactly?

core/metadata.js
enguerran: I didn't understand a thing.

Do you want me to explicit the intention with the name of the function? Like:
```js
const myRegExp = createRegExpWithConflictSuffix('docname');
// /docname-conflict-\d{4}-.../
```

test/unit/metadata.js
enguerran: This is recommended by `mocha`: https://mochajs.org/#getting-started

test/unit/metadata.js
enguerran: Yes, you are right, my mistake.

test/unit/metadata.js
enguerran: I will check it.

test/unit/metadata.js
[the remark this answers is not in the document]
enguerran: Using a prefix `[conflict-19700101_133700000]-docname.tar.gz` would be something more robust.
If we can list common extensions, we cannot list all extension.

core/metadata.js
sebn: I'm just a bit confused by `with*` in the current name.
Maybe it's not necessary? E.g.:

```js
function conflictRegExp (basenamePattern /*: string */ ='(.*)') { ... }
````

File: core/metadata.js [hunk 17]
```diff
@@ -456,3 +457,27 @@ function buildFile (filePath /*: string */, stats /*: fs.Stats */, md5sum /*: st
   if ((stats.mode & EXECUTABLE_MASK) !== 0) { doc.executable = true }
   return doc
 }
+
+const CONFLICT_REGEXP = new RegExp('-conflict-\\d{4}(?:-\\d{2}){2}T(?:\\d{2}_?){3}.\\d{3}Z')
```
sebn:
```js
const CONFLICT_PATTERN = '-conflict-\\d{4}(?:-\\d{2}){2}T(?:\\d{2}_?){3}.\\d{3}Z'
```
Wdyt?
From what I can see the current RegExp object is not exported: we only use it 3 lines below to get its `.source` pattern... So maybe better make it a plain string and rename it to make the type clear?

test/unit/metadata.js
```diff
@@ -790,7 +790,7 @@ describe('metadata', function () {
       }
       const newDoc = createConflictingDoc(doc)
       const pathRegExp = withConflictRegExp(doc.path)
-      should(newDoc._id).match(pathRegExp)
+      should(newDoc._id.toLowerCase()).match(pathRegExp)
```
enguerran: Oops on macOS

> AssertionError: expected 'DOCNAME-CONFLICT-2018-12-11T13_33_43.697Z' to match /docname-conflict-\d{4}(?:-\d{2}){2}T(?:\d{2}_?){3}.\d{3}Z/

core/metadata.js
enguerran: Naming is hard, if you prefer some name rather than others, let's rename it.
`prefixedConflictRegExp` or `withConflictRegExp` or `createRegExpWithConflictSuffix` or `conflictRegExp` are very in common. This function returns a regular expression that matches anything that has the `CONFLICT_PATTERN` at the end of it: `getRegExpWithConflictPattern` or `withConflictPattern`, …

test/unit/metadata.js
sebn: It allows guessing the test filter from the function name without even looking at the test description:

```bash
yarn test:unit -g createConflictingDoc
# Vs
yarn test:unit -g 'generated a doc with conflict'
```

sebn: (if the function name changes, we'll have to change it in each test anyway)

enguerran: If the function name changes, my editor's tools that help me renaming things will rename things, but not the comments.

test/unit/metadata.js
sebn: I would still prefer sticking to the current convention:

```bash
$ git co master 
$ grep -r "it('should" test/ | wc -l
1
$ grep -r "it('[^s]" test/ | wc -l
591
```

test/unit/metadata.js
sebn: Build a valid Metadata object, so we could possibly enable flow?
(i.e. adding `/* @flow */`at the top of this file)

test/unit/metadata.js
sebn: Well, it doesn't matter so much. Please forget about my comment :)

test/unit/metadata.js
sebn: Doesn't matter so much. Please forget about my comment :)

test/unit/metadata.js
enguerran: the behavior of the function do not need a `Metadata` object but a subset of this kind of object with only a `path`, why do we want to restrict our tests to an implementation instead of a interface?

enguerran: this function takes something with a `path` property and returns something with a `path` property that contains a suffix to stamp the path with a conflict suffix. In fact, this function should take only a file name, as a string.

test/unit/metadata.js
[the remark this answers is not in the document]
enguerran: We have no plan to change the stamp by prefixing it. In addition, another tool that sync files uses a suffix https://www.dropbox.com/help/syncing-uploads/conflicted-copy.
I resolve this conversation.

test/unit/metadata.js
sebn: :+1: Agreed!

sebn:
:+1:

test/unit/metadata.js
enguerran: I use the flag `i` for the RegExp (case insensitive)

test/unit/metadata.js
```diff
@@ -772,4 +774,60 @@ describe('metadata', function () {
...
+    })
+    it('should keep the correct extension', () => {
+      const ext = '.pdf'
+      const doc = {
+        path: `docname${ext}`
+      }
+      const newDoc = createConflictingDoc(doc)
+      should(path.extname(newDoc.path)).equal(ext)
+    })
+    it('should but don\'t handle complex extension `.tar.gz`', () => {
```
taratatach: NItpick: either remove `should` or replace `but don\'t` with `not`

PR #1334, opened 2018-12-07
[WIP] Add conflict integration tests & fix an edge case

aenario:


test/integration/conflict_resolution.js
```diff
@@ -150,6 +150,107 @@ describe('Conflict resolution', () => {
...
+    await helpers.remote.pullChanges()
+    await helpers.syncAll()
+    await helpers.local.scan()
+    await helpers.remote.pullChanges()
+    await helpers.syncAll()
+    should(await helpers.trees()).deepEqual({remote: expected, local: expected})
+  }
+
+  describe('create 2 files same name', () => {
+    it('file local & folder remote', async () => {
```
taratatach: Don't we have 2 files in this context?

test/integration/conflict_resolution.js
```diff
@@ -150,6 +150,107 @@ describe('Conflict resolution', () => {
...
   })
 
+  const afterFullSyncBothTreeDeepEqual = async (expected) => {
+    await helpers.local.scan()
+    await helpers.remote.pullChanges()
+    await helpers.syncAll()
+    await helpers.local.scan()
+    await helpers.remote.pullChanges()
+    await helpers.syncAll()
+    should(await helpers.trees()).deepEqual({remote: expected, local: expected})
```
taratatach: Maybe the newly created `mergeSideEffects` helper would be useful here?

test/integration/conflict_resolution.js
```diff
@@ -150,6 +150,107 @@ describe('Conflict resolution', () => {
...
+        'same-name-conflict-...',
+        'same-name/'
+      ]
+
+      await afterFullSyncBothTreeDeepEqual(expectedTree)
+      await afterFullSyncBothTreeDeepEqual(expectedTree)
+    })
+  })
+
+  describe('move folder to conflict', () => {
```
taratatach: I don't really understand what is meant here.

test/support/helpers/context_dir.js
```diff
@@ -123,7 +123,7 @@ class ContextDir {
   }
 
   async move (src /*: string */, dst /*: string */) {
-    return fs.move(this.abspath(src), this.abspath(dst))
+    return fs.rename(this.abspath(src), this.abspath(dst))
```
taratatach: What's the difference between those 2?

taratatach:
We're not just adding an integration test here, we're apparently fixing a bug. I think we should make this clear in the commit & PR titles.

test/support/helpers/context_dir.js
sebn: [`Local#_move()`](https://github.com/cozy-labs/cozy-desktop/blob/v3.11.1-beta.1/core/local/index.js#L342) actually uses [`fs.rename()`](https://github.com/cozy-labs/cozy-desktop/blob/v3.11.1-beta.1/core/local/index.js#L359), so this better matches our own sync actions.

Also `fs.move` actually comes from `fs-extra`, not `fs`, it does some black magic to ensure move succeeds in weird cases (e.g. separate volumes), and may end up changing the inode of watched files.

test/integration/conflict_resolution.js
sebn: This is an integration test, not a Merge unit test :)

test/integration/conflict_resolution.js
```diff
@@ -150,6 +150,145 @@ describe('Conflict resolution', () => {
...
   })
 
+  const afterFullSyncBothTreeDeepEqual = async (expected) => {
+    await helpers.local.scan()
+    await helpers.remote.pullChanges()
+    await helpers.syncAll()
+    await helpers.local.scan()
+    await helpers.remote.pullChanges()
+    await helpers.syncAll()
+    should(await helpers.trees()).deepEqual({remote: expected, local: expected})
```
sebn: Looks like putting should.js assertions into non-test functions makes stacktraces useless... :sob:

sebn: I refactored the test helpers to extract the assertions back to the test functions.

core/merge.js
```diff
@@ -266,6 +266,9 @@ class Merge {
         log.info({path}, 'up to date')
         return null
       } else {
+        if (doc.remote._id !== folder.remote._id) {
+          await this.resolveConflictAsync(side, folder, folder)
```
taratatach: Why not mark the new `doc` as the conlicting document instead?

sebn:
I'll make separate PRs so we can merge tests, test support fix & sync fix separately.

PR #1335, opened 2018-12-07
test: Remove mocha require

enguerran:
Mocha runtime injects mocha functions in the global scope
All we need to do is to tell eslint we are in a mocha environment.

- [ ] PR is not too big


test/support/helpers/platform.js
```diff
@@ -27,11 +26,11 @@ module.exports = {
 //     })
 //
 function onMacOSAtLeast (minRelease /*: MacOSReleaseInfo */, spec /*: Function */) {
-  const describe = macOSRelease.isAtLeast(minRelease)
-    ? mocha.describe
-    : mocha.describe.skip
+  const describeOverwritten = macOSRelease.isAtLeast(minRelease)
```
sebn: `describeOrSkip`?

enguerran: Why not, at least it assumes the bad idea of overwriting a function.

enguerran:
> This one may be missing:
https://github.com/cozy-labs/cozy-desktop/blob/v3.11.0-beta.1/test/world/case_and_encoding.js#L4

I was searching in `./test` folder. How can I be so naive. A change is on its way.

test/support/helpers/platform.js
```diff
@@ -43,24 +42,23 @@ function onMacOSAtLeast (minRelease /*: MacOSReleaseInfo */, spec /*: Function *
...
 
-  describe(`on ${macOSRelease.name(maxRelease)} or lower`, spec)
+  describeOrSkip(`on ${macOSRelease.name(maxRelease)} or lower`, spec)
 }
 
-// $FlowFixMe
-function onPlatforms (...platformsAndSpec) {
-  const spec = platformsAndSpec.pop()
-  const expectedPlatforms = platformsAndSpec
+function onPlatforms (...platformsAndSpec /*: any[] */) {
```
enguerran: I had to cast the parameters array with `any` as it can be a `string[]::Function`.
We use it like this `onPlatforms("win32", "linux", callback)` or `onPlatforms("linux", callback)`. So it may be refactored to express a little better what it is all about. For instance it could take an array of platforms as `string` and a callback parameter as a `Function` instead of `any`.

enguerran:
Regarding 4bb0ee79a5ac8aad3debe72ff94386b9a864a260 : I cannot understand why flow checker is happy with something like `onPlatforms("linux", callback)` but it breaks on `expectedPlatforms.join is not a function`.

See 
screenshot-2018-12-10-16-13-05

The signature of `onPlatforms` is `Array<string>`:

```js
function onPlatforms (expectedPlatforms /*: Array<string> */, spec /*: Function */) { }
```

enguerran:
appveyor reached a timeout…

PR #1340, opened 2018-12-12
feat: Consider same binaries as non-conflicting

enguerran:
This PR aims to consider files with same binary not to be treated
as conflicting files.

- [ ] PR is not too big


sebn:
Empty branch/commit? Git mistake maybe?

enguerran:
I won't work on this now, let someone opens its own PR when the work will begin.

PR #1343, opened 2018-12-14
core: Alias fs-extra as fse everywhere & stop promisifying it

sebn:
So one can easily know whether she's using the standard Node.js `fs` module or `fs-extra` just by looking at the function prefix, e.g. `fs.rename(...)` vs `fse.rename(...)`.

Also we were uselessly using Bluebird to promisify `fs-extra` in a few places although it's already promisified by itself.

### DIFF
--- a/core/app.js
+++ b/core/app.js
@@ -1,7 +1,7 @@
 /* @flow */
 
 const autoBind = require('auto-bind')
-const fs = require('fs-extra')
+const fse = require('fs-extra')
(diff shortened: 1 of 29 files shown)


sebn:
This PR will need to be rebased in case it is merged after #1344

sebn:
Please merge #1344 first :wink:

sebn:
Rebased on #1344 and resolved conflicts. Waiting for CI.

PR #1345, opened 2018-12-14
Remove dead code in Remote (dissociate from remote)

nono:


nono:
[Supprimer du code mort](https://trello.com/c/14TCCbS7/1333-supprimer-du-code-mort)

PR #1348, opened 2018-12-18
core: Setup a dev env for Elm windows

enguerran:
Elm windows do not need the whole app to start when they need works.
Here a fake HTML file is used to load elm windows and css and
it is served by a simple http server, so developers can see
their changes in milliseconds.

Here I added
- one HTML file for dev purpose
- an update in chokidar watch eventlistener
- a npm/yarn script to start the whole thing
- some documentation to explain what to do

- [x] PR is not too big

### DIFF
--- a/dev/elm.html
+++ b/dev/elm.html
@@ -1,0 +1,30 @@
+<!DOCTYPE html>
+<html>
+    <head>
+        <meta charset="UTF-8" />
+        <title>Main</title>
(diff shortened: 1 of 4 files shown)


enguerran:
There is no livereload for now, you have to refresh the browser when the transpilations are done (`stylus` → `css` and `elm` → `js`).
If someone knows how to quickly add it, let me know.

PR #1353, opened 2018-12-18
Use fileIds on windows

nono:


nono:
[Utiliser les fileId sous windows (stat et initial diff)](https://trello.com/c/wkDhdScR/1342-utiliser-les-fileid-sous-windows-stat-et-initial-diff)

core/local/index.js
```diff
@@ -146,14 +153,25 @@ module.exports = class Local /*:: implements Side */ {
   inodeSetter (doc /*: Metadata */) {
```
taratatach: We could try to come up with a different name for this function as it doesn't set the inode only anymore.
Something in the likes of `identifiersSetter`?

core/local/steps/add_infos.js
```diff
@@ -24,10 +30,23 @@ module.exports = function (buffer /*: Buffer */, opts /*: { syncPath: string } *
           event._id = id(event.path)
           if (['created', 'modified', 'renamed'].includes(event.action)) {
             log.debug({path: event.path, action: event.action}, 'stat')
-            event.stats = await fse.stat(path.join(opts.syncPath, event.path))
+            if (winfs) {
+              // XXX It would be better to avoid sync IO operations, but
+              // before node 10.5.0, it's our only choice for reliable fileIDs
```
taratatach: `FIXME: upgrade node to v10.5.0`?

core/local/index.js
```diff
@@ -254,20 +272,31 @@ module.exports = class Local /*:: implements Side */ {
...
+            cbStats(err, {})
+          }
+        } else {
+          fse.stat(tmpFile, cbStats)
+        }
       },
 
       next => fse.ensureDir(parent, () => fse.rename(tmpFile, filePath, next)),
 
-      this.inodeSetter(doc),
```
taratatach: When do we set the inode and fileId? Was this line useless?

core/local/steps/add_infos.js
nono: Well, currently, there is no version of electron with node v10.5.0+

core/local/index.js
nono: At lines 279 & 280. It's better to do it before moving the file from the tmp directory to its final destination: it's one less syscall to stats, and it avoids a possible race condition where the file is deleted between renaming and calling `inodeSetter`.

core/local/index.js
nono: `identifier` makes me think to the `id`, that are built from the `path`, not from the `inode` and `fileId`. Do you see another name that could fit?

nono: I moved everything in the new `stater` module

core/local/index.js
nono: It's now line 261

core/local/steps/add_infos.js
taratatach: The goal is not to forget to do it when we can

core/local/steps/initial_diff.js
```diff
@@ -39,6 +38,9 @@ async function initialDiff (buffer, out, pouch) {
     const batch = []
     for (const event of events) {
       if (['created', 'modified', 'renamed', 'scan'].includes(event.action)) {
+        if (event.stats.fileid) {
+          byInode.delete(event.stats.fileid)
+        }
         byInode.delete(event.stats.ino)
```
taratatach: Why no conditions on the presence of `ino`? Do we have this property on Windows?

core/local/stater.js
```diff
@@ -0,0 +1,77 @@
...
+        // XXX It would be better to avoid sync IO operations, but
+        // before node 10.5.0, it's our only choice for reliable fileIDs.
+        resolve(winfs.lstatSync(filepath))
+      } catch (err) {
+        reject(err)
+      }
+    })
+  },
+
+  withStats: function (filepath /*: string */, callback /*: Callback */) {
```
taratatach: Is there a reason why you're using the `property: function` declaration style instead of the new style used below for `isDirectory` and such?

core/local/steps/initial_diff.js
nono: We always have `ino`, even on windows with the new lib

core/local/stater.js
nono: yes, the goal is to avoid rewriting https://github.com/cozy-labs/cozy-desktop/blob/master/core/local/index.js#L180-L278 which use callbacks (this PR seems big enough without that).

core/local/steps/initial_diff.js
taratatach: OK, I thought `fileId` was the concept of inode on Windows.

nono: It is, but nodejs uses `ino` for inode on linux & macOS, and for fileid on Windows (as a number). The issue is that we can have the same `ino` for two fileIds because the precision of number is not enough on JS. So, to improve things, we will use the `fileid` (as a string) from an external lib in the future, and will still use the `ino` to help the transition.

core/local/stater.js
```diff
@@ -0,0 +1,75 @@
...
+  },
+
+  kind (stats /*: Stats */) {
+    return this.isDirectory(stats) ? 'directory' : 'file'
+  },
+
+  assignInoAndFileId (doc /*: Metadata */, stats /*: Stats */) {
+    doc.ino = stats.ino
+    // $FlowFixMe
+    if (stats.fileid) { doc.fileid = stats.fileid }
```
sebn: Maybe using `typeof` would still be better than another *FlowFixMe*?
E.g.:

```js
if (typeof stats.fileid === 'string') { ... }
```

core/local/steps/add_infos.js
```diff
@@ -21,13 +21,14 @@ module.exports = function (buffer /*: Buffer */, opts /*: { syncPath: string } *
...
+          // TODO if (event.kind === 'symlink') emit an error
           event._id = id(event.path)
           if (['created', 'modified', 'renamed'].includes(event.action)) {
             log.debug({path: event.path, action: event.action}, 'stat')
-            event.stats = await fse.stat(path.join(opts.syncPath, event.path))
+            event.stats = await stater.stat(path.join(opts.syncPath, event.path))
           }
           if (event.stats) { // created, modified, renamed, scan
-            event.docType = event.stats.isDirectory() ? 'directory' : 'file'
+            event.docType = stater.kind(event.stats)
```
sebn: Feels like mixing *doctypes* with *kinds*, although the latter can be more than just `file` or `directory`...
Should work now but may break at some point.

core/metadata.js
```diff
@@ -66,6 +67,7 @@ export type Metadata = {
   trashed?: true,
   incompatibilities?: *,
   ino?: ?number,
+  fileid?: ?string,
```
sebn: I suppose we could expect this prop to disappear after some time using a recent enough Node.js version?

core/metadata.js
```diff
@@ -368,7 +370,15 @@ const makeComparator = (name, interestingFields) => {
   return (one, two) => {
     const diff = deepDiff(one, two, filter)
     log.trace({path: two.path, diff}, name)
-    return !diff || _.every(diff, canBeIgnoredDiff)
+    if (diff && !_.every(diff, canBeIgnoredDiff)) {
+      return false
+    }
+    // XXX The fileid can be missing in some old documents in couchdb.
```
sebn: s/couch/pouch :wink:

core/metadata.js
```diff
@@ -424,8 +434,8 @@ function markSide (side /*: string */, doc /*: Metadata */, prev /*: ?Metadata *
   return doc
 }
 
-function buildDir (fpath /*: string */, stats /*: fs.Stats */, remote /*: ?MetadataRemoteInfo */) /*: Metadata */ {
-  const doc /*: Object */ = {
+function buildDir (fpath /*: string */, stats /*: Stats */, remote /*: ?MetadataRemoteInfo */) /*: Metadata */ {
+  let doc /*: Object */ = {
```
sebn: Why `let` since only `fileid` prop is assigned later, not the var itself?

test/unit/merge.js
```diff
@@ -496,7 +496,7 @@ describe('Merge', function () {
 
       const result = await this.pouch.db.get(doc._id)
       should(result._rev).not.equal(old._rev)
-      should(result).have.properties(_.omit(doc, '_rev'))
+      should(result).have.properties(_.omit(doc, ['_rev', 'fileid']))
```
sebn: Shouldn't we omit some prop only in case we cannot possibly guess its value?

core/metadata.js
nono: I'm not sure. Nodejs will give us the fileId as a Big Int in a new field, it's probably safer to serialize it to a string in the `fileid` field.

PR #1354, opened 2018-12-19
test/unit/local/analysis: Group tests by outputs & inputs

sebn:
To make it easier to see (e.g. using editor folding):

- which output cases are not covered yet
- for the same output, which input cases are not covered yet

Tests with more than 1 change as the output can still end up in their own group.

- [x] PR is not too big


sebn:
Before:

local-analysis-tests-before

After:

local-analysis-tests-after

WDYT?

PR #1360, opened 2018-12-20
[WIP] Fix an invariant

nono:
When a file is added on local, and its parent is moved just after, the file has only the local side and no remote. The document for this file is given to markAsNeverSynced, and then it fails on an invariant with the message: doc has 'sides.remote' but not remote. This commit fix it.


nono:
[Corriger une erreur d'invariant](https://trello.com/c/PlIKZEt9/1345-corriger-une-erreur-dinvariant)

nono:
-> #1364

PR #1361, opened 2018-12-20
Check references in property based testing

nono:
We simulate the references in the property based testing by using the `chmod g+w` bit to files, and setting the inode number in the doc.remote field in pouchdb.


nono:
[Ajouter du property based testing pour les déplacements et renommages](https://trello.com/c/rB0MSGcd/1347-ajouter-du-property-based-testing-pour-les-d%C3%A9placements-et-renommages)

test/property/local_watcher/index.js
```diff
@@ -148,6 +199,24 @@ describe('Local watcher', function () {
 
       // And no conflict should have happened
       should(state.conflicts).be.empty()
+
+      // And the references should have kept in pouchdb
```
taratatach: nit: "been kept"

core/merge.js
```diff
@@ -305,16 +306,18 @@ class Merge {
     metadata.markSide(side, doc, folder)
```
taratatach: Shouldn't we mark the sides after the id conflict detection as well? This is what we do for file moves.

PR #1364, opened 2018-12-21
core/metadata: Don't add side when not syncing doc

taratatach:
Marking a file or directory as unsyncable (previously "never synced")
  should not add a new side to the metadata, especially when not syncing
  a local document as this can lead to errors with documents not having
  a remote object but a remote side.

  We now simply mark the document for deletion, increase its existing
  side and save it to the Pouch.
  The Sync component is now responsible for not synchronising deleted
  documents with only one side.

- [x] PR is not too big


nono:
[Corriger une erreur d'invariant](https://trello.com/c/PlIKZEt9/1345-corriger-une-erreur-dinvariant)

PR #1365, opened 2018-12-26
Detect files/dirs moved on windows

nono:

### DIFF
--- a/core/local/steps/win_detect_move.js
+++ b/core/local/steps/win_detect_move.js
@@ -2,6 +2,10 @@
 
 const { id } = require('../../metadata')
 const Buffer = require('./buffer')
+const logger = require('../../logger')
+const log = logger({
+  component: 'winDetectMove'
+})
 
 // Wait at most this delay (in milliseconds) to see if it's a move.
 const DELAY = 1000
@@ -48,6 +52,8 @@ async function winDetectMove (buffer, out, pouch) {
         try {
           const was = await pouch.db.get(id(event.path))
           deleted.set(was.fileid, event.path)
+        } catch (err) {
+          // Ignore the error
         } finally {
           release()
         }
@@ -75,6 +81,7 @@ async function winDetectMove (buffer, out, pouch) {
               event.oldPath = e.path
               pending[i].deleted.delete(event.stats.fileid)
               pending[i].events.splice(j, 1)
+              break
             }
           }
         }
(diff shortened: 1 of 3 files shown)


nono:
[Détecter les déplacements sous windows](https://trello.com/c/JMLno44m/1349-d%C3%A9tecter-les-d%C3%A9placements-sous-windows)

core/local/steps/win_detect_move.js
```diff
@@ -0,0 +1,101 @@
...
+    // Then, see if a created event matches a deleted event
+    for (const event of events) {
+      if (event.action === 'created') {
+        for (let i = 0; i < pending.length; i++) {
+          const path = pending[i].deleted.get(event.stats.fileid)
+          if (!path || path === event.path) {
+            continue
+          }
+          const l = pending[i].events.length
+          for (let j = 0; j < l; j++) {
```
taratatach: Can we have multiple `deleted` events for a unique `created` event when renaming a file/folder?
If not, aren't `pending[i].deleted.get(fileId)` and `pending[i].events[j]` (which has the same file id) the same object? In this case we could avoid this loop entirely.

taratatach: Or maybe it's necessary for the `splice` just a few lines below.
But I'd rather have the loop used for this purpose only for the sake of clarity.

nono: > Can we have multiple deleted events for a unique created event when renaming a file/folder?
No, I don't we can that.

> If not, aren't pending[i].deleted.get(fileId) and pending[i].events[j] (which has the same file id) the same object?
No, the first one is a string (the path), the other is an object (the event).

> In this case we could avoid this loop entirely.
We need the loop for `splice`.

> But I'd rather have the loop used for this purpose only for the sake of clarity.
Why not, but I'm not sure to see what you want to move and where. Can you make a suggestion?

core/local/steps/win_detect_move.js
```diff
@@ -0,0 +1,93 @@
...
+
+  while (true) {
+    // Wait for a new batch of events
+    const events = await buffer.pop()
+
+    // First, push the new events in the pending queue
+    const deleted = new Map()
+    for (const event of events) {
+      if (event.action === 'deleted') {
+        const release = await pouch.lock('winMoveDetector')
```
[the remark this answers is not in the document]
taratatach: That was my point. I'm not convinced it's of any use here. But it probably doesn't do any harm either so if you feel safer with it let's keep it.

core/local/steps/win_detect_move.js
taratatach: No need for suggestions. Avoiding the loop would have been possible if `pending[i].deleted.get(fileId)` and `pending[i].events[j]` would both have been events. In this case I would have prefered to use the getter to access the event since it clearly shows we can only have one `deleted` event for any single `created` event.

PR #1366, opened 2018-12-27
core/merge: Local change & unapplied remote change

taratatach:
When we have fetched a remote change and failed to apply it, we end up
  with a remote side superior to our local side.
  If in this situation we try to merge a local change, we'll try to
  resolve the conflict remotely and end up renaming the existing file
  with the conflict suffix and potentially lose the local data after
  successfully applying the remote changes.

  We can avoid this situation by doing a local conflict resolution,
  ensuring we have both the existing file and its conflicting copy.
  This has the additionnal advantage of keeping potential sharings of the
  remote file clean.

- [x] PR is not too big

### DIFF
--- a/core/merge.js
+++ b/core/merge.js
@@ -136,8 +136,8 @@ class Merge {
       }
     }
     if (file) {
-      if ((side === 'local') && (file.sides.local != null)) {
-        return this.resolveInitialAddAsync(side, doc, file)
+      if (side === 'local' && file.sides.local != null) {
+        return this.updateFileAsync('local', doc)
       } else {
         return this.resolveConflictAsync(side, doc, file)
       }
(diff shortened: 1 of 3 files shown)


core/merge.js
```diff
@@ -156,6 +156,11 @@ class Merge {
     } else if (file.sides.local && file.sides.local >= file.sides.remote) {
       // The file was updated on local after being synched to remote
       return this.updateFileAsync(side, doc)
+    } else if (file.sides.local && file.sides.local < file.sides.remote) {
+      delete doc.remote
+      delete doc.sides
+      metadata.markSide('local', doc)
```
nono: I don't think we need to `markSide` before calling `resolveConflictAsync`

core/merge.js
```diff
@@ -156,6 +156,11 @@ class Merge {
...
+      return this.resolveConflictAsync('local', doc, doc)
```
nono: Maybe, instead of adding a new `else if` in this function, we can modify https://github.com/cozy-labs/cozy-desktop/pull/1366/files#diff-6d18f9e7a792437c06b6f95186c97b88R176 to always resolve the conflict on local. It won't change anything from your PR if `file.sides.local` exist. If it doesn't exist, it looks safe to me to resolve the conflict on local (and it's probably better).

What do you think?

core/merge.js
taratatach: You're right, we're not saving `doc` into the pouch so it's unnecessary. As are the two `delete` above.

core/merge.js
nono: Yes, I mean the last branch where we resolve the conflict remotely.

PR #1367, opened 2018-12-27
core/merge: Separate files with different inodes

taratatach:
When we try to apply locally an update to a file at a path already
  known in the Pouch DB (i.e. during an initial scan), we override the
  existing document and remote copy if we can (i.e. no changes were
  fetched from the Cozy).

  However, if the previous and the new files have different inodes, it
  means the first one was deleted and then a new file was created at
  the same path. It doesn't mean those files are the same or different
  versions of the same file. We should not override the previous copy in
  this situation and do a conflict resolution instead.

Based on #1366, it should be merged after.

- [x] PR is not too big

### DIFF
--- a/core/merge.js
+++ b/core/merge.js
@@ -158,6 +158,14 @@ class Merge {
     }
     metadata.assignMaxDate(doc, file)
     if (file) {
+      let inConflict = doc.ino && doc.ino !== file.ino
+      if (file.fileid && doc.fileid) {
+        inConflict = file.fileid !== doc.fileid
+      }
+      if (inConflict) {
+        return this.resolveConflictAsync(side, doc, file)
+      }
+
       doc._rev = file._rev
       doc.moveFrom = file.moveFrom
       if (doc.tags == null) { doc.tags = file.tags || [] }
--- a/test/support/builders/metadata/base.js
+++ b/test/support/builders/metadata/base.js
@@ -99,6 +99,11 @@ module.exports = class BaseMetadataBuilder {
     return this
   }
 
+  fileid (fileid /*: string */) /*: this */ {
+    this.doc.fileid = fileid
+    return this
+  }
+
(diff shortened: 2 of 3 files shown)


core/merge.js
```diff
@@ -158,6 +158,18 @@ class Merge {
     }
     metadata.assignMaxDate(doc, file)
     if (file) {
+      if (doc.ino && doc.ino !== file.ino) {
+        await this.resolveConflictAsync(side, doc, file)
+        if (side === 'remote') {
+          // we just renamed the remote file as a conflict
```
nono: I think it is dead code: if `side === 'remote'`, the `doc` was built by the Remote Watcher, and it should have no `ino`. Right?

taratatach: You're probably right.

core/merge.js
```diff
@@ -158,6 +158,10 @@ class Merge {
     }
     metadata.assignMaxDate(doc, file)
     if (file) {
+      if (doc.ino && doc.ino !== file.ino) {
```
nono: On windows, we should use the `fileid` (if available). Maybe something like:
```js
const inConflict = doc.ino && doc.ino !== file.ino
if (file.fileid && doc.fileid) {
  inConflict = file.fileid !== doc.fileid
}
if (inConflict) {
  return this.resolveConflictAsync(side, doc, file)
}
```

taratatach:
The failing test was introduced by https://github.com/cozy-labs/cozy-desktop/commit/2eefaa682 and its goal is clearly to make sure we'll overwrite a file during an update even if its inode has changed.
The greater goal behind this is not obvious but I think we can find examples where overwriting an updated file with a different inode is legitimate (e.g. file overwritten by a temporary file during a save).

This is enough to make me doubt about the legitimacy of this PR and close it.

PR #1369, opened 2018-12-28
Handle the checksum error

nono:


nono:
[Gérer les erreurs sur l'étape du checksum](https://trello.com/c/wE71PvRw/1353-g%C3%A9rer-les-erreurs-sur-l%C3%A9tape-du-checksum)

core/local/steps/incomplete_fixer.js
```diff
@@ -0,0 +1,95 @@
...
+      if (incompletes.length === 0 || event.action !== 'renamed') {
+        continue
+      }
+
+      const limit = +new Date() - DELAY
+      for (let i = 0; i < incompletes.length; i++) {
+        const item = incompletes[i]
+
+        // Remove the expired incomplete events
+        if (i === 0 && item.timestamp < limit) {
```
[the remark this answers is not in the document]
nono: The expired incomplete events are always at the start of the array. The condition is probably useless now. When I put it, the limit was not a variable but something rebuilt on each iteration, and I fear that if the loop was too slow, the limit was no longer the same for all the events, and that will make us remove the wrong events. Since then, I've moved the limit in a variable initialized outside of the loop, so, we can remove this condition (or keep it): it shouldn't matter.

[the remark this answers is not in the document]
taratatach: I'd rather remove it if it's unnecessary. Better avoid raising questions about it.

core/local/steps/incomplete_fixer.js
```diff
@@ -0,0 +1,95 @@
...
+          incompletes.shift()
+          i--
+          continue
+        }
+
+        if (event.oldPath && (item.event.path + '/').startsWith(event.oldPath)) {
```
taratatach: If think the second part of the condition rather be:
```javascript
item.event.path.startsWith(event.oldPath + '/')
```

Otherwise, we're still matching `item.event` with `event` in this situation:
```javascript
event = { oldPath: 'src/file', path: 'dst/file' }
item = { event: { path: 'src/file-conflict-<timestamp>' } }
```

core/local/chokidar_watcher.js
```diff
@@ -254,9 +254,6 @@ module.exports = class LocalWatcher {
     return null
   }
 
-  // @PERFOPTIM ?
-  //   - db.allDocs(keys: events.pick(path))
-  //   - process.exec('md5sum ' + paths.join(' '))
```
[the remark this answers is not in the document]
nono: I don't really think it's a good idea to do so. And even if it was, it seems to be at the bottom of the priorities. On macOS, there are a lot of more important tasks, like removing the need of a period of 10 seconds with no events to process them. And when we will have a clean local watcher for macOS, we should do some tracing/benchmarking to see what should be optimized.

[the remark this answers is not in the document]
nono: For `EISDIR`, I think the comment is no longer relevant.

> (e.g. rm foo && mkdir foo)

With that, we will receive an event of type `addDir`, not `add`, and so we don't enter in the block of code.

core/local/steps/incomplete_fixer.js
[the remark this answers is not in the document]
nono: Well, I have the opposite reaction: the code raises question in my mind when we do not check if `i` is 0 as we are doing a `shift` to remove an item. But, if you prefer without it, let's go for that.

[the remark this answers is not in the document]
taratatach: Ah I see what you mean. I didn't think about that at all :D. In this case, maybe we could replace `incomples.shift()` with `incompletes.splice(i, 1)`. This would probably get optimized by the browser.

core/local/chokidar_watcher.js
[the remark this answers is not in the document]
sebn: I'm almost sure Chokidar may report a `change` event when replacing [a dir with a file](https://github.com/cozy-labs/cozy-desktop/blob/v3.11.2-beta.2/test/scenarios/replace_dir_with_file/local/darwin.json.DISABLED) or [a file with a dir](https://github.com/cozy-labs/cozy-desktop/blob/v3.11.2-beta.2/test/scenarios/replace_file_with_dir/local/darwin.json.DISABLED).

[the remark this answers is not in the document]
nono: OK, let's keep this FIXME

core/local/steps/await_write_finish.js
```diff
@@ -9,95 +8,96 @@ const log = logger({
...
 // This is a port of awaitWriteFinish (aWF) from chokidar. It debounces write
 // events for files, as we can have several of them in a short lapse of time,
 // and computing the checksum several times in a row for the same file is not a
 // good idea.
-async function awaitWriteFinish (buffer, out) {
-  const pending = new Map()
+async function awaitWriteFinish (buffer /*: Buffer */, out /*: Buffer */) {
+  const waiting /*: WaitingItem[] */ = []
 
   while (true) {
```
sebn: How is the loop supposed to stop?
Will something in the `while` block just throw at some point?

sebn: Would it make sense to separate *stream* stuff (`Buffer`, `while (true)`) from the rest?
Would passing a plain callback (e.g. `out.push`) make sense for lower-level functions?
Wouldn't it make testing edge cases (especially regarding `waiting` events) easier without having to deal with infinite loop + Buffers?

core/local/steps/incomplete_fixer.js
```diff
@@ -0,0 +1,95 @@
...
+        const item = incompletes[i]
+
+        // Remove the expired incomplete events
+        if (item.timestamp < limit) {
+          incompletes.splice(i, 1)
+          i--
+          continue
+        }
+
+        if (event.oldPath && item.event.path.startsWith(event.oldPath + '/')) {
```
sebn: Will `/` work on Windows?

core/local/steps/incomplete_fixer.js
```diff
@@ -0,0 +1,95 @@
...
+            batch.push({
+              action: item.event.action,
+              path: p,
+              _id: id(p),
+              kind,
+              stats,
+              md5sum
+            })
+          } catch (err) {
+            // If we have an error, there is probably not much that we can do
```
sebn: Maybe log the error?

core/local/steps/incomplete_fixer.js
```diff
@@ -0,0 +1,95 @@
...
+// Cf test/property/local_watcher/swedish_krona.json
+module.exports = function (buffer /*: Buffer */, opts /*: { syncPath: string , checksumer: Checksumer } */) /*: Buffer */ {
+  const incompletes = []
+  return buffer.asyncMap(async (events) => {
+    // Filter out the incomplete events
+    const batch = []
+    for (const event of events) {
+      if (event.incomplete) {
+        const now = new Date()
+        incompletes.push({ event, timestamp: Number(now) })
```
sebn: `Date.now()`

core/local/steps/incomplete_fixer.js
```diff
@@ -0,0 +1,95 @@
...
+            const p = item.event.path.replace(event.oldPath, event.path)
+            const absPath = path.join(opts.syncPath, p)
+            const stats = await stater.stat(absPath)
+            const kind = stater.kind(stats)
+            let md5sum
+            if (kind === 'file') {
+              md5sum = await opts.checksumer.push(absPath)
+            }
+            batch.push({
+              action: item.event.action,
```
sebn: Shouldn't we check the ino/fileid before pushing our made-up event?
Otherwise, couldn't we end building a completely wrong event in case of a weird race condition?

core/local/steps/incomplete_fixer.js
```diff
@@ -0,0 +1,95 @@
...
+        if (event.oldPath && item.event.path.startsWith(event.oldPath)) {
+          // We have a match, try to rebuild the incomplete event
+          try {
+            const p = item.event.path.replace(event.oldPath, event.path)
+            const absPath = path.join(opts.syncPath, p)
+            const stats = await stater.stat(absPath)
+            const kind = stater.kind(stats)
+            let md5sum
+            if (kind === 'file') {
+              md5sum = await opts.checksumer.push(absPath)
```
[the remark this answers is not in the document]
sebn: Although we need checksums in both cases, I don't think resolving an incomplete event is the same job as checksuming the initial ones, so I'm fine with separating both.

core/local/steps/incomplete_fixer.js
sebn: Also what happens if the file itself was created then renamed?

core/local/steps/incomplete_fixer.js
```diff
@@ -0,0 +1,95 @@
...
+        continue
+      }
+
+      const limit = Number(new Date()) - DELAY
+      for (let i = 0; i < incompletes.length; i++) {
+        const item = incompletes[i]
+
+        // Remove the expired incomplete events
+        if (item.timestamp < limit) {
+          incompletes.splice(i, 1)
```
sebn: Maybe log *spliced* events?

core/local/steps/incomplete_fixer.js
```diff
@@ -0,0 +1,95 @@
...
+      if (incompletes.length === 0 || event.action !== 'renamed') {
+        continue
+      }
+
+      const limit = Number(new Date()) - DELAY
+      for (let i = 0; i < incompletes.length; i++) {
+        const item = incompletes[i]
+
+        // Remove the expired incomplete events
+        if (item.timestamp < limit) {
```
sebn: Wouldn't it be more intuitive with `item.timestamp + DELAY >= now`?
Or maybe just rename `limit` to `lowerLimit`?

core/local/steps/incomplete_fixer.js
```diff
@@ -0,0 +1,95 @@
...
+// a directory that is an ancestor of this file has been moved. With the renamed
+// event, by comparing the path, we can extrapolate the new path and check with
+// fs.stats if we have a file here.
+//
+// Cf test/property/local_watcher/swedish_krona.json
+module.exports = function (buffer /*: Buffer */, opts /*: { syncPath: string , checksumer: Checksumer } */) /*: Buffer */ {
+  const incompletes = []
+  return buffer.asyncMap(async (events) => {
+    // Filter out the incomplete events
+    const batch = []
```
sebn: Maybe extract `async function fixIncompleteEvents(incompletes, events, opts)` to make it easier to test out of `buffer.asyncMap()`?

core/local/steps/incomplete_fixer.js
sebn: Also maybe extract the `try` block (except the `batch.push()` call) as a function? E.g.:

```js
const event = await rebuildIncompleteEvent(item, event, opts)
batch.push(event)
```

core/local/steps/incomplete_fixer.js
```diff
@@ -0,0 +1,95 @@
...
+              path: p,
+              _id: id(p),
+              kind,
+              stats,
+              md5sum
+            })
+          } catch (err) {
+            // If we have an error, there is probably not much that we can do
+          }
+          incompletes.splice(i, 1)
```
sebn: Couldn't this end up dropping an initial incomplete `create_file` event followed by multiple `rename` ones, since the   `stat()` call would fail for the intermediate destination?
Do we expect the last rename to be gracefully handled as a `create_file` because its source is missing?

core/local/steps/await_write_finish.js
```diff
@@ -9,95 +8,96 @@ const log = logger({
...
+  }
+}
+
 // This is a port of awaitWriteFinish (aWF) from chokidar. It debounces write
 // events for files, as we can have several of them in a short lapse of time,
 // and computing the checksum several times in a row for the same file is not a
 // good idea.
-async function awaitWriteFinish (buffer, out) {
-  const pending = new Map()
+async function awaitWriteFinish (buffer /*: Buffer */, out /*: Buffer */) {
```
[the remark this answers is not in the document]
nono: It's a good idea, but we should do that in all the steps for consistency, and so, it should be done in another PR

core/local/steps/await_write_finish.js
nono: Currently, the loop does never stop. I'm not sure if we need to do so or not.

nono: About making code easier to test, it's a good idea, but it should probably done when someone will write the test (in another PR)

core/local/steps/incomplete_fixer.js
nono: I don't know. But we should probably test that in the PR that adds test for `incompleteFixer`

core/local/steps/incomplete_fixer.js
nono: In the checklist for things to do before going in production with the new watcher, there is an item for logs. I think it may be easier to add logs in all the steps in the same PR, as it probably more clear to see what should be logged and what should be ignored with a global view.

core/local/steps/incomplete_fixer.js
nono: > Shouldn't we check the ino/fileid before pushing our made-up event?

We are doing a new `fs.stats` to have the more recent ino/fileid. I'm not sure to see what you want to check.

core/local/steps/incomplete_fixer.js
nono: I don't know, it's something that we should test.

sebn: Maybe add a `TODO` in the code?

core/local/steps/incomplete_fixer.js
sebn: To be logged in an upcoming PR, see https://github.com/cozy-labs/cozy-desktop/pull/1369#discussion_r245348961

core/local/steps/await_write_finish.js
sebn: > Currently, the loop does never stop. I'm not sure if we need to do so or not.

Maybe add it to the watcher checklist?

core/local/steps/incomplete_fixer.js
sebn: To be checked in unit tests, same as https://github.com/cozy-labs/cozy-desktop/pull/1369#discussion_r245353053

core/local/steps/incomplete_fixer.js
sebn: To be done when adding unit tests, same as https://github.com/cozy-labs/cozy-desktop/pull/1369#discussion_r245349730

PR #1370, opened 2018-12-28
Fix side number for file/folder created at a path used in the past

nono:


nono:
[Mauvais numéro pour le side d'un fichier/dossier créé pour un chemin qui a été utilisé dans le passé](https://trello.com/c/w7P9wAkR/1355-mauvais-num%C3%A9ro-pour-le-side-dun-fichier-dossier-cr%C3%A9%C3%A9-pour-un-chemin-qui-a-%C3%A9t%C3%A9-utilis%C3%A9-dans-le-pass%C3%A9)

core/merge.js
```diff
@@ -154,7 +154,9 @@ class Merge {
     }
     doc.sides[side] = ++rev
     doc._rev = info.rev
-    return this.pouch.put(doc)
+    const result = await this.pouch.put(doc)
+    delete doc._rev
```
taratatach: What's the point of removing the doc's revision here? We're not returning `doc` or reusing it so I assume we don't really care what its value is.

nono: Some unit tests failed when `_rev` is added to the doc.

taratatach: Wouldn't that be a problem with how the tests were written rather than the implementation in this case?

taratatach:
We'll wait for the stable release before merging this PR.

core/merge.js
taratatach: OK. Another argument for a good refactoring of Merge :).

PR #1374, opened 2019-01-03
test/scenarios: create_file & update_file actions

sebn:
- `>` becomes `create_file` (same a property-based tests)
- `>>` becomes `update_file` (same a property-based tests)
- `update_file` explicit `content` is now mandatory (makes scenarios easier to understand and remote instrumentation doesn't need to download previous content anymore)

- [x] PR is not too big


sebn:
> No explicit content for the `create_file` action?

The existing scenarios with `create_file` actions don't seem confusing to me.
I believe we can fix it later as needed.

sebn:
Rebased on master.

PR #1375, opened 2019-01-03
test/support: Add AtomWatcherEvent builder

taratatach:
This builder follows the same principles as the metadata and remote
  builders to create AtomWatcherEvent objects.

  Some types and constants used in AtomWatcherEvent were exported for
  consistency.

- [x] PR is not too big

### DIFF
--- a/core/local/steps/event.js
+++ b/core/local/steps/event.js
@@ -3,9 +3,22 @@
 /*::
 import type { Stats } from 'fs'
 
+export type EventAction =
+  | 'created'
+  | 'modified'
+  | 'deleted'
+  | 'renamed'
+  | 'scan'
+  | 'initial-scan-done'
+export type EventKind =
+  | 'file'
+  | 'directory'
+  | 'symlink'
+  | 'unknown'
+
 export type AtomWatcherEvent = {
-  action: "created" | "modified" | "deleted" | "renamed" | "scan" | "initial-scan-done",
-  kind: "file" | "directory" | "symlink" | "unknown",
+  action: EventAction,
+  kind: EventKind,
   path: string,
   oldPath?: string,
   _id?: string,
@@ -15,3 +28,15 @@ export type AtomWatcherEvent = {
 
 export type Batch = AtomWatcherEvent[]
 */
+
+const ACTIONS /*: EventAction[] */ = [
+  'created', 'modified', 'deleted', 'renamed', 'scan', 'initial-scan-done'
+]
+const KINDS /*: EventKind[] */ = [
+  'file', 'directory', 'symlink', 'unknown'
+]
+
+module.exports = {
+  ACTIONS,
+  KINDS
+}
--- a/test/support/builders/atom_watcher_event.js
+++ b/test/support/builders/atom_watcher_event.js
@@ -1,0 +1,70 @@
+/* @flow */
+
+const metadata = require('../../../core/metadata')
+const events = require('../../../core/local/steps/event')
+
+/*::
+import type { Stats } from 'fs'
+import type { AtomWatcherEvent, EventAction, EventKind, Batch } from '../../../core/local/steps/event'
+*/
+
+function randomPick /*:: <T> */ (elements /*: Array<T> */) /*: T */{
+  const l = elements.length
+  const i = Math.floor(Math.random() * l)
+  return elements[i]
+}
+
+module.exports = class AtomWatcherEventBuilder {
+  /*::
+  _event: AtomWatcherEvent
+  */
+
+  constructor () {
+    this._event = {
+      action: randomPick(events.ACTIONS),
+      kind: randomPick(events.KINDS),
+      path: '/',
+      _id: '/'
+    }
+  }
+
+  build () /*: AtomWatcherEvent */ {
+    return this._event
+  }
+
+  action (newAction /*: EventAction */) /*: this */ {
+    this._event.action = newAction
+    return this
+  }
+
+  kind (newKind /*: EventKind */) /*: this */ {
+    this._event.kind = newKind
+    return this
+  }
+
+  path (newPath /*: string */) /*: this */ {
+    this._event.path = newPath
+    this._event._id = metadata.id(newPath)
+    return this
+  }
+
+  oldPath (newPath /*: string */) /*: this */{
+    this._event.oldPath = newPath
+    return this
+  }
+
+  id (newId /*: string */) /*: this */ {
+    this._event._id = newId
+    return this
+  }
+
+  stats (newStats /*: Stats */) /*: this */ {
+    this._event.stats = newStats
+    return this
+  }
+
+  md5sum (newMd5sum /*: string */) /*: this */ {
+    this._event.md5sum = newMd5sum
+    return this
+  }
+}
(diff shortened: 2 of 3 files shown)


test/support/builders/atom_watcher_event.js
```diff
@@ -0,0 +1,70 @@
...
+const events = require('../../../core/local/steps/event')
+
+/*::
+import type { Stats } from 'fs'
+import type { AtomWatcherEvent, Action, Kind, Batch } from '../../../core/local/steps/event'
+*/
+
+function randomPick (elements /*: Array<T> */) /*: T */{
+  const l = elements.length
+  const i = Math.floor(Math.random() * Math.floor(l))
```
nono: Why `Math.floor(l)`?

test/support/builders/atom_watcher_event.js
```diff
@@ -0,0 +1,70 @@
+const metadata = require('../../../core/metadata')
```
nono:
```js
/* @flow */
```

test/support/builders/atom_watcher_event.js
```diff
@@ -0,0 +1,70 @@
...
+
+module.exports = class AtomWatcherEventBuilder {
+  constructor () {
+    this._event /*: AtomWatcherEvent */ = {
+      action: randomPick(events.ACTIONS),
+      kind: randomPick(events.KINDS),
+      path: '/',
+      _id: '/'
+    }
+  }
```
enguerran: What is the purpose for a stochastic constructor?

test/support/builders/atom_watcher_event.js
```diff
@@ -0,0 +1,70 @@
...
+    if (!events.KINDS.includes(newKind)) {
+      throw new Error(`Unknown event kind "${newKind}". Available kinds: ${events.KINDS.join(', ')}.`)
+    }
+    this._event.kind = newKind
+    return this
+  }
+
+  path (newPath /*: string */) /*: this */ {
+    this._event.path = newPath
+    this._event._id = metadata.id(newPath)
```
enguerran: This is generated by `add_infos`, what do we want as a goal?

test/support/builders/atom_watcher_event.js
```diff
@@ -0,0 +1,70 @@
...
+import type { AtomWatcherEvent, Action, Kind, Batch } from '../../../core/local/steps/event'
+*/
+
+function randomPick (elements /*: Array<T> */) /*: T */{
+  const l = elements.length
+  const i = Math.floor(Math.random() * Math.floor(l))
+  return elements[i]
+}
+
+module.exports = class AtomWatcherEventBuilder {
```
enguerran: I don't understand what's gained from writing all this code to fill in a shell of an object typed by Flow

test/support/builders/atom_watcher_event.js
taratatach: To have an integer between `0` and `elements.length`

taratatach: Ah I see what you mean. `l` is already an integer. I'll fix that.

test/support/builders/atom_watcher_event.js
taratatach: I'm not sure I fully understand your question but I wrote this builder to have default values for some properties I don't care about most of the time, have one-liners when creating test events and build the foundations for what could evolve if AtomWatcherEvent were to get more complex.

test/support/builders/atom_watcher_event.js
taratatach: g

I want to make sure we don't rely on the action and kind of an event unless we've specified it.
It's also a way to test more combinations over time.

test/support/builders/atom_watcher_event.js
nono: I think this builder may be used for testing most of the steps, but not for `addInfos`

test/support/builders/atom_watcher_event.js
nono: Humm, I think you should care about the action for testing almost all the steps (but I'm not so sure about where this discussion is going ;-) )

test/support/builders/atom_watcher_event.js
taratatach: Maybe I should make some changes to the builder to make it useful when testing `addInfos`?!

core/local/steps/event.js
```diff
@@ -3,9 +3,22 @@
...
+  | 'file'
+  | 'directory'
+  | 'symlink'
+  | 'unknown'
+
 export type AtomWatcherEvent = {
-  action: "created" | "modified" | "deleted" | "renamed" | "scan" | "initial-scan-done",
-  kind: "file" | "directory" | "symlink" | "unknown",
+  action: Action,
+  kind: Kind,
```
sebn: Maybe prefix those type names?

test/support/builders/atom_watcher_event.js
```diff
@@ -0,0 +1,76 @@
...
+    }
+  }
+
+  build () /*: AtomWatcherEvent */ {
+    return this._event
+  }
+
+  action (newAction /*: Action */) /*: this */ {
+    if (!events.ACTIONS.includes(newAction)) {
+      throw new Error(`Unknown event action "${newAction}". Available actions: ${events.ACTIONS.join(', ')}.`)
```
sebn: Aren't we doing flow's job twice?

test/support/builders/atom_watcher_event.js
```diff
@@ -0,0 +1,76 @@
...
+
+/*::
+import type { Stats } from 'fs'
+import type { AtomWatcherEvent, Action, Kind, Batch } from '../../../core/local/steps/event'
+*/
+
+function randomPick /*:: <T> */ (elements /*: Array<T> */) /*: T */{
+  const l = elements.length
+  const i = Math.floor(Math.random() * l)
+  return elements[i]
```
sebn:
```js
_.first(_.shuffle(elements))
```

test/support/builders/atom_watcher_event.js
sebn: @taratatach Did you use all these methods in your tests? Or are we trying to anticipate future needs?

core/local/steps/event.js
taratatach: Why not. I'll do that.

test/support/builders/atom_watcher_event.js
taratatach: I didn't have Flow's checks in place when I wrote those methods so it might be overkill. I left them because I often run the linter only right before I push the commits (or even after sometimes…) and thought it could be useful to have this check.

I can remove them.

test/support/builders/atom_watcher_event.js
taratatach: I'm not a big fan of shuffling the array just to pick one element.

taratatach:
@sebn I anticipated a bit. I added methods to modify every property of the event.
As of now, I'm using: `kind`, `path` and `build`.

test/support/builders/atom_watcher_event.js
taratatach: You can see some examples in https://github.com/cozy-labs/cozy-desktop/pull/1377

PR #1378, opened 2019-01-03
test: Cover scan_folder in some condition

enguerran:

- [x] PR is not too big

### DIFF
--- a/test/unit/local/steps/scan_folder.js
+++ b/test/unit/local/steps/scan_folder.js
@@ -1,0 +1,113 @@
+/* eslint-env mocha */
+/* @flow */
+
+const should = require('should')
+const sinon = require('sinon')
+const scanFolder = require('../../../../core/local/steps/scan_folder')
+const Buffer = require('../../../../core/local/steps/buffer')
+
+describe('core/local/steps/scan_folder', () => {
+  it('should call the producer scan for action `created`', async () => {
+    const batch = [
+      {
+        action: 'created',
+        kind: 'directory',
+        path: __dirname
+      }
+    ]
+    const buffer = new Buffer()
+    buffer.push(batch)
+    const opts = {
+      producer: {
+        start: () => Promise.resolve(),
+        stop: () => Promise.resolve(),
+        scan: sinon.stub().resolves()
+      }
+    }
+    const enhancedBuffer = scanFolder(buffer, opts)
+    const enhancedBatch = await enhancedBuffer.pop()
+    should(enhancedBatch).eql(batch)
+    should(opts.producer.scan).be.calledOnce()
+  })
+
+  it('should not call the producer scan for action other than `created`', async () => {
+    const batch = [
+      {
+        action: 'scan',
+        kind: 'directory',
+        path: __dirname
+      },
+      {
+        action: 'renamed',
+        kind: 'directory',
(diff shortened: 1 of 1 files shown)


test/unit/local/steps/scan_folder.js
```diff
@@ -0,0 +1,113 @@
...
+        kind: 'directory',
+        path: __dirname
+      },
+      {
+        action: 'initial-scan-done',
+        kind: 'directory',
+        path: __dirname
+      },
+      {
+        action: 'created',
```
taratatach: I don't see why we have this created action here since we're testing "action other than created" and we're testing it in other tests anyway.

enguerran: I should have written somthing clearer on https://github.com/cozy-labs/cozy-desktop/pull/1378/files/9cc9d693ec22df872746cad48de369f1aa1ab5e8#diff-412e8b6fa135229b9b9e59c42a1f769bR77

Like `should(enhancedBatch).be.length(batch.filter(event => event.action !== 'created').length)`.

I will make the change, thanks for the review.

enguerran: Oh, I didn't take enough time to get my thoughts straight.
I should have removed the `'should call the producer scan for action created'` test. I wrote it first, but you are right, the test `'should not call the producer scan for action other than `created`'` covers the same feature.
And if I remove the `created` object, given this test, I could not say if it did work or not since I want the stub not to be called.

enguerran: > I don't see why we have this created action here since we're testing "action other than created"

If you want to test the `add` function that takes two numbers and returns the addition of these two numbers, you have to test `add(3,4) === 7` and another addition. Otherwise, we test that the function always returns some number.

We could introduce [fuzzy tests](https://en.wikipedia.org/wiki/Fuzzing) to test random number and check the addition function returns the result of the addition. Or for our case, to test fuzzy events with random `action` and `kind` values and check the result of the `scan_folder` behavior.

taratatach: My point was that, the way I understand the test description, we're testing that for all actions that are not "created", we call the producer scan.
To test this, we don't need any "created" action.
But maybe I misunderstood the description?

test/unit/local/steps/scan_folder.js
```diff
@@ -0,0 +1,123 @@
...
+        scan: sinon.stub().resolves()
+      }
+    }
+    const enhancedBuffer = scanFolder(buffer, opts)
+    const enhancedBatch = await enhancedBuffer.pop()
+    should(enhancedBatch).be.length(batch.length)
+    should(opts.producer.scan).be.calledThrice()
+  })
+
+  it('should call the producer scan for kind `folder` only', async () => {
```
taratatach: `s/kind folder/kind directory`

test/unit/local/steps/scan_folder.js
taratatach: Nevermind. The description has changed and I get it now.

test/unit/local/steps/scan_folder.js
enguerran: the module and the function is called `scan_folder` ;-)

PR #1379, opened 2019-01-07
core: Indent flow type definitions

sebn:
To make future changes easier to review.
This commit includes only indentation changes, nothing else.

PR #1380, opened 2019-01-07
test/scenarios: Rename move dir + update file scenario 

sebn:
So we can later test locally too.


sebn:
To be squashed and merged after https://github.com/cozy-labs/cozy-desktop/pull/1379

sebn:
To be squashed and merged after https://github.com/cozy-labs/cozy-desktop/pull/1369

PR #1381, opened 2019-01-07
Enable atom/watcher by default for Windows and Linux

nono:


nono:
[Activer par défaut le nouveau watcher](https://trello.com/c/TznL4IxQ/1371-activer-par-d%C3%A9faut-le-nouveau-watcher)

core/local/watcher.js
```diff
@@ -20,20 +19,27 @@ export interface Watcher {
...
-const log = logger({
-  component: 'LocalWatcher'
-})
-
 function build (syncPath /*: string */, prep /*: Prep */, pouch /*: Pouch */, events /*: EventEmitter */, ignore /*: Ignore */) /*: Watcher */ {
+  let watcher = 'atom'
+  if (process.platform === 'darwin') {
+    watcher = 'chokidar'
+  }
   const env = process.env.COZY_FS_WATCHER
```
sebn: Wouldn't using the same name as the environment variable be easier to understand? E.g.:

```js
const { COZY_FS_WATCHER } = process.env
```

core/local/watcher.js
```diff
@@ -20,20 +19,27 @@ export interface Watcher {
...
-    } catch (err) {
-      log.error({err}, 'Cannot use AtomWatcher')
-    }
+    watcher = 'atom'
+  } else if (env === 'chokidar') {
+    watcher = 'chokidar'
+  }
+  // FIXME Integration and scenario tests use ChokidarEvents
+  // and are not yet compatible with atom/watcher
+  if (process.env) {
```
sebn: I'm not sure to understand this condition :upside_down_face:

core/local/watcher.js
```diff
@@ -20,20 +19,27 @@ export interface Watcher {
...
-      return new AtomWatcher(syncPath, prep, pouch, events, ignore)
-    } catch (err) {
-      log.error({err}, 'Cannot use AtomWatcher')
-    }
+    watcher = 'atom'
+  } else if (env === 'chokidar') {
+    watcher = 'chokidar'
+  }
+  // FIXME Integration and scenario tests use ChokidarEvents
+  // and are not yet compatible with atom/watcher
```
sebn: Shouldn't we set `COZY_FS_WATCHER` in `.travis.yml` and `.appveyor.yml`?

core/local/watcher.js
```diff
@@ -20,20 +19,27 @@ export interface Watcher {
...
-  component: 'LocalWatcher'
-})
-
 function build (syncPath /*: string */, prep /*: Prep */, pouch /*: Pouch */, events /*: EventEmitter */, ignore /*: Ignore */) /*: Watcher */ {
+  let watcher = 'atom'
+  if (process.platform === 'darwin') {
+    watcher = 'chokidar'
+  }
   const env = process.env.COZY_FS_WATCHER
   if (['experimental', 'atom'].includes(env)) {
```
sebn: Do we really need 2 names for the same thing? Do we really use them?

Also maybe we could make the available watcher types easier to grasp? E.g.:

```js
type WatcherType = 'atom' | 'chokidar'
```

core/local/watcher.js
nono: We are using `onFlush` for those tests, and the `onFlush` method is defined only on the old watcher, not the new watcher. And those tests create some `ChokidarEvents` (with a doctype, not a kind).

nono:
> Another question: are we making it the default in master right now just to make a beta build and revert it immediately?

At least, we should release a beta version with the new watcher activated by default on windows, so that our PO can test it on windows.

And we should revert it before making a new stable release on the OSes where we are not confident about the new watcher.

PR #1382, opened 2019-01-07
Handle errors in addInfos

nono:

### DIFF
--- a/core/local/steps/add_infos.js
+++ b/core/local/steps/add_infos.js
@@ -19,7 +19,7 @@ module.exports = function (buffer /*: Buffer */, opts /*: { syncPath: string } *
     const batch = []
     for (const event of events) {
       if (event.kind === 'symlink') {
-        log.error({event}, 'Symlink are not supported')
+        log.error({event}, 'Symlinks are not supported')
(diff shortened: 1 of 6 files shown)


nono:
[Gérer les erreurs sur l'étape du addInfos](https://trello.com/c/0PAKC0lH/1372-g%C3%A9rer-les-erreurs-sur-l%C3%A9tape-du-addinfos)

core/local/steps/add_infos.js
```diff
@@ -33,11 +37,11 @@ module.exports = function (buffer /*: Buffer */, opts /*: { syncPath: string } *
             event.kind = event.kind === 'directory' ? 'directory' : 'file'
           }
         }
-        batch.push(event)
       } catch (err) {
         log.info({err, event}, 'Cannot get infos')
-        console.log('stats', err) // TODO error handling
+        event.incomplete = true
```
sebn: Could the `try` / `catch` block wrap only the `stater.stat()` call?

nono: We could, but I prefer to not do that. I don't think it's a good idea to set the `kind` for an event when `stats` has failed.

PR #1383, opened 2019-01-07
core/ignore: Change isIgnored signature

taratatach:
We're starting to use Ignore#isIgnored with different types of
  "documents" (i.e. Metadata and AtomWatcherEvent) while they don't all
  have the expected `_id` and `docType` properties.

  Since those 2 properties are the only information we need to decide
  whether a "document" should be ignored or not, we can request them as
  parameters so their values can be computed correctly by the caller
  depending on the type of "document" it's dealing with.
  Also, since we'll move to the new Atom Watcher, we'll use its
  EventKind type instead of the current DocType type to determine
  whether the given "document" is a directory or not.

To help transition to the new format, we replace all calls to
  isIgnored for Metadata objects with calls to a new method
  isIgnoredDoc which uses isIgnored internally and maps the document's
  docType to an EventKind.

- [x] PR is not too big

### DIFF
--- a/core/ignore.js
+++ b/core/ignore.js
@@ -19,10 +19,6 @@ export type IgnorePattern = {
 }
 */
 
-/* ::
-import type {Metadata} from './metadata.js'
-*/
-
 /** Load both given file rules & default ones */
 function loadSync (rulesFilePath /*: string */) /*: Ignore */ {
   let ignored
@@ -46,7 +42,7 @@ function readLinesSync (filePath /*: string */) /*: string[] */ {
 }
 
 // Parse a line and build the corresponding pattern
-function buildPattern (line) {
+function buildPattern (line /*: string */) /*: IgnorePattern */ {
   let folder = false
   let negate = false
   let noslash = line.indexOf('/') === -1
@@ -94,7 +90,7 @@ function isNotBlankOrComment (line /*: string */) /*: boolean */ {
   return line !== '' && line[0] !== '#'
 }
 
-function match (path, isFolder, pattern /*: IgnorePattern */) {
+function match (path /*: string */, isFolder /*: boolean */, pattern /*: IgnorePattern */) /*: boolean */ {
   if (pattern.basename) {
     if (pattern.match(basename(path))) {
       return true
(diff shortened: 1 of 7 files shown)


core/ignore.js
```diff
@@ -132,23 +132,23 @@ class Ignore {
 
   // Add some rules for things that should be always ignored (temporary
   // files, thumbnails db, trash, etc.)
-  addDefaultRules () {
+  addDefaultRules () /*: this */ {
```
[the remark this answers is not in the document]
taratatach: Yes, I've seen that used in the test builders. It is very handy indeed.

core/prep.js
```diff
@@ -109,8 +109,9 @@ class Prep {
     doc.docType = 'file'
     assignId(doc)
     assignId(was)
-    let docIgnored = this.ignore.isIgnored(doc)
-    let wasIgnored = this.ignore.isIgnored(was)
+    let docIgnored = this.ignore.isIgnored(doc._id, 'file')
+    const wasKind = was.docType === 'folder' ? 'directory' : was.docType
+    let wasIgnored = this.ignore.isIgnored(was._id, wasKind)
```
sebn: The old code looks more simple than the new one...

sebn:
What about adding a separate `Ignore` method for this?

core/ignore.js
```diff
@@ -132,23 +132,23 @@ class Ignore {
...
     const defaultPatterns = buildPatternArray(readLinesSync(defaultRulesPath))
     this.patterns = defaultPatterns.concat(this.patterns)
     return this
   }
 
   // Return true if the given file/folder path should be ignored
-  isIgnored (doc /*: Metadata */) {
+  isIgnored (pathLike /*: string */, kind /*: EventKind */) /*: boolean */ {
```
sebn: What about adding a separate `Ignore` method for this?

core/prep.js
taratatach: I know, this is the problem with transitions :/. I didn't want to go too far but if that's an issue I can change my position on this.

core/ignore.js
taratatach: You'd rather have one method for `Metadata` and one for `AtomWatcherEvent`?

sebn: I just want to make sure we keep a `Metadata` specific one, since we're still using it with docs in ~10 places.
It could eventually be renamed, e.g. to something like `#isIgnoredDoc()`.

sebn: Actually `Ignore` only need to know whether something is a folder (aka *boolean*):

```
Ignore#match({path /*: string */, isFolder /*: boolean */})
```

Both `Metadata` and `AtomWatcherEvent` versions could be plain functions using `Ignore#match()`:

```
metadata.shouldIgnore(doc /*: Metadata */, ignoreRules /*: Ignore */)
shouldIgnore(event /*: AtomWatcherEvent */, ignoreRules /*: Ignore */) // defined in ignore step since it is only used there
```

toutes deux utilisant :

core/local/steps/filter_ignored.js
```diff
@@ -11,8 +12,15 @@ import type { Ignore } from '../../ignore'
 // watchers), but it needs to be put after the AddInfos step as the docType is
 // required to know if the event can be ignored.
 module.exports = function (buffer /*: Buffer */, opts /*: { ignore: Ignore } */) /*: Buffer */ {
+  const notIgnored = shouldKeep(opts.ignore)
+
   return buffer.map((batch) => {
-    // $FlowFixMe
-    return batch.filter(event => !opts.ignore.isIgnored(event))
+    return batch.filter(notIgnored)
```
sebn: *notIgnore* => *shouldKeep* => *isIgnored* :wink:

test/unit/ignore.js
```diff
@@ -44,352 +44,176 @@ describe('Ignore', function () {
   describe('Ignored patterns', () => {
     it("don't ignore file name not matching to the pattern", function () {
       const ignore = new Ignore(['foo'])
-      const doc = {
-        _id: 'bar',
-        docType: 'file'
-      }
-      ignore.isIgnored(doc).should.be.false()
+      ignore.isIgnored('bar', false).should.be.false()
```
sebn: Named params could have helped understand what `false` means (see [example](https://github.com/cozy-labs/cozy-desktop/pull/1383#discussion_r245671355)).

PR #1385, opened 2019-01-07
core/local: Replace change.fromEvent() with dedicated builders

sebn:
Flow typechecking is terribly inaccurate in `local/analysis` & `local/change`.
The `change.fromEvent()` function in particular is messing up flow in unexpected ways.
For example, it prevents using the `?: ...` syntax (vs `: ?...`) in a few type definitions.

Replacing it with dedicated builders is a first necessary step to regain sanity.

The builders currently include some duplication.
But this should be easy to DRY later, once the code is more reliable.

Also the `LocalFileAddition` type is currently lying: `md5sum` is missing when `wip` is `true`.
Fixing it will require quite more work, hence the temporary `delete md5sum` hack when a file
is added then unlinked.

### DIFF
--- a/core/local/analysis.js
+++ b/core/local/analysis.js
@@ -100,7 +100,7 @@ function analyseEvents (events /*: LocalEvent[] */, pendingChanges /*: LocalChan
 
             changeFound(
               localChange.fileMoveIdenticalOffline(e) ||
-              localChange.fromEvent(e)
+              localChange.fileAddition(e)
             )
           }
           break
@@ -123,7 +123,7 @@ function analyseEvents (events /*: LocalEvent[] */, pendingChanges /*: LocalChan
             }
             changeFound(
               localChange.dirMoveIdenticalOffline(e) ||
-              localChange.fromEvent(e)
+              localChange.dirAddition(e)
             )
           }
           break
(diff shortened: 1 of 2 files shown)


sebn:
Sorry for the CI failure. Everything is green now. Ready for last review, then squash & merge.

core/local/change.js
```diff
@@ -193,35 +197,76 @@ function isChildAdd (a /*: LocalChange */, b /*: LocalChange */) { return childO
...
+  log.debug({path: e.path}, 'addDir = DirAddition')
+  const change /*: LocalDirAddition */ = {
+    sideName,
+    type: 'DirAddition',
+    path: e.path,
+    stats: e.stats,
+    ino: e.stats.ino
+  }
+  if (e.old) change.old = e.old
+  if (e.wip) change.wip = e.wip
```
nono: By curiosity, we can have `wip` on directory? In which conditions?

core/local/analysis.js
```diff
@@ -179,6 +179,7 @@ function analyseEvents (events /*: LocalEvent[] */, pendingChanges /*: LocalChan
               // $FlowFixMe
```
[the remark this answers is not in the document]
sebn: Yes, otherwise flow complains because we're changing the `type` tag property.
And I believe flow fails when a `$FlowFixMe` is not needed anymore.

core/local/change.js
sebn: Currently `wip` is set on an `addDir` event in case the directory doesn't exist anymore (although this may not be such a good idea).

PR #1386, opened 2019-01-07
test: Cover linux_producer with tests

enguerran:
This PR aims to cover `core/local/steps/linux_producer.js`.
It uses temporary folder to create folders and files in an asynchronous mode.
I tried to stay clear by reducing variable declaration, but let me know if reviewers think it will be clearer to name variable.

- [x] PR is not too big

### DIFF
--- a/test/unit/local/steps/linux_producer.js
+++ b/test/unit/local/steps/linux_producer.js
@@ -4,6 +4,7 @@
 const should = require('should')
 const os = require('os')
 const fs = require('fs')
+const fse = require('fs-extra')
 const path = require('path')
 const { onPlatform } = require('../../../support/helpers/platform')
 const LinuxProducer = require('../../../../core/local/steps/linux_producer')
@@ -169,13 +170,4 @@ onPlatform('linux', () => {
   })
 })
 
-function mkdirsSync (mypath = '/') {
-  mypath.split('/').reduce((acc, folder) => {
-    try {
-      fs.mkdirSync(path.join(acc, folder))
-    } catch (ex) {
-      // `EEXIST ${path.join(acc, folder)}`
-    }
-    return `${acc}/${folder}`
-  }, '/')
-}
+const mkdirsSync = fse.mkdirsSync
--- a/test/unit/local/steps/linux_producer.js
+++ b/test/unit/local/steps/linux_producer.js
@@ -5,9 +5,10 @@ const should = require('should')
 const os = require('os')
 const fs = require('fs')
 const path = require('path')
+const { onPlatform } = require('../../../support/helpers/platform')
 const LinuxProducer = require('../../../../core/local/steps/linux_producer')
 
-if (os.platform() === 'linux') {
+onPlatform('linux', () => {
   describe('core/local/steps/linux_producer', () => {
     describe('API of the producer', () => {
       let syncPath
@@ -174,7 +175,7 @@ if (os.platform() === 'linux') {
       })
     })
   })
-}
+})
 
 function mkdirSyncRecursive (syncPath = '/', mypath = '/') {
   mypath.split('/').reduce((acc, folder) => {
--- a/test/unit/local/steps/linux_producer.js
+++ b/test/unit/local/steps/linux_producer.js
@@ -7,166 +7,174 @@ const fs = require('fs')
 const path = require('path')
 const LinuxProducer = require('../../../../core/local/steps/linux_producer')
 
-describe.only('core/local/steps/linux_producer', () => {
-  describe('API of the producer', () => {
-    let syncPath
-    let producer
+if (os.platform() === 'linux') {
+  describe('core/local/steps/linux_producer', () => {
+    describe('API of the producer', () => {
+      let syncPath
+      let producer
 
-    beforeEach(() => {
-      syncPath = fs.mkdtempSync(path.join(os.tmpdir(), 'foo-'))
-      producer = new LinuxProducer({
-        syncPath
+      beforeEach(() => {
+        syncPath = fs.mkdtempSync(path.join(os.tmpdir(), 'foo-'))
+        producer = new LinuxProducer({
+          syncPath
+        })
+      })
+
+      it('should register on start and dispose on stop the watcher', async () => {
+        await producer.start()
+        should(producer.watcher).be.not.null()
+        await producer.stop()
+        should(producer.watcher).be.null()
+      })
+
+      it('should change absolute path to relative with process', async () => {
+        const dirname = 'foobar'
+        const batch = [
+          {
+            action: 'created',
+            kind: 'directory',
+            path: path.join(syncPath, dirname)
+          }
+        ]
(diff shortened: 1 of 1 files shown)


test/unit/local/steps/linux_producer.js
```diff
@@ -7,166 +7,173 @@ const fs = require('fs')
 const path = require('path')
 const LinuxProducer = require('../../../../core/local/steps/linux_producer')
 
-describe.only('core/local/steps/linux_producer', () => {
-  describe('API of the producer', () => {
-    let syncPath
-    let producer
+os.platform() === 'linux' &&
+  describe.only('core/local/steps/linux_producer', () => {
```
enguerran: What do you think about this code to run tests only on linux platform?

nono: I'm not a fan of using `&&` with a lot of lines after: it means that I can't jump in my editor from `{` to `}` to see on what it applies. I know it's a matter of taste (as Python and CoffeeScript avoid blocks by using indentation), but still I don't think it's a good idea here.

And s/describe.only/describe/

enguerran: There is only one instruction here `os.platform() === 'linux' && describe('core/local/steps/linux_producer', linuxProducerTests)`.

But okay, I was not sure anyway. I will update to a more classic if statement:

```
if (os.platform() === 'linux') {
    describe('core/local/steps/linux_producer', linuxProducerTests)
}
```

enguerran: I always forget the `.only`!!!
I am an evil person!

taratatach: Did you think about using the `onPlatform` helper?

test/unit/local/steps/linux_producer.js
```diff
@@ -0,0 +1,184 @@
...
+            kind: 'file',
+            path: newname
+          }
+        ])
+      })
+    })
+  })
+}
+
+function mkdirSyncRecursive (syncPath = '/', mypath = '/') {
```
taratatach: `fs-extra` has the `ensureDir` method that does exactly that: https://github.com/jprichardson/node-fs-extra/blob/master/docs/ensureDir.md

```javascript
fse.ensureDir(path.join(syncPath, myPath))
```

test/unit/local/steps/linux_producer.js
```diff
@@ -0,0 +1,184 @@
...
+          (await producer.buffer.pop()).map(({ action, kind, path }) => ({
+            action,
+            kind,
+            path
+          }))
+        ).eql([
+          {
+            action: 'scan',
+            kind: 'unknown',
+            path: `${mypath[0]}/${mypath[1]}`
```
taratatach: We should probably use `path.join` here to make sure paths are correct on Windows as well.

test/unit/local/steps/linux_producer.js
```diff
@@ -0,0 +1,184 @@
...
+          (await producer.buffer.pop()).map(({ action, kind, path }) => ({
+            action,
+            kind,
+            path
+          }))
+        ).eql([
+          {
+            action: 'scan',
+            kind: 'unknown',
+            path: mypath.join('/')
```
taratatach: Same here.

test/unit/local/steps/linux_producer.js
```diff
@@ -0,0 +1,184 @@
...
+          batch.map(evt => Object.assign({}, evt, { path: dirname }))
+        )
+      })
+
+      it('should scan a subfolder tree', async () => {
+        const mypath = ['foo', 'bar', 'baz']
+        mkdirSyncRecursive(syncPath, mypath.join('/'))
+        await producer.start()
+        should(
+          (await producer.buffer.pop()).map(({ action, kind, path }) => ({
```
taratatach: suggestion: pop 3 times and then map and compare only once

test/unit/local/steps/linux_producer.js
enguerran: Starting from nodejs 10.2.0, the `mkdir` and `mkdirSync` functions from `fs` is also recursive. 10.x is a LTS. Do we plan to upgrade from 8.2.1? /poke @nono @sebn

test/unit/local/steps/linux_producer.js
enguerran: I have no plan to test `linux_producer` on Windows, but you are right.

taratatach: You're right, I should have thought about that :man_facepalming:

test/unit/local/steps/linux_producer.js
nono: Our nodejs version is constrained by electron. We should upgrade electron, but it's often a burden, so I don't think we will do that in the next 15 days. After that, it depends of the priorities.

test/unit/local/steps/linux_producer.js
enguerran: done

enguerran:
@taratatach I changed the implementation, you can follow the changes history with the commits.
Let me know if your review have changed.

test/unit/local/steps/linux_producer.js
```diff
@@ -0,0 +1,173 @@
...
+const should = require('should')
+const os = require('os')
+const fs = require('fs')
+const fse = require('fs-extra')
+const path = require('path')
+const { onPlatform } = require('../../../support/helpers/platform')
+const LinuxProducer = require('../../../../core/local/steps/linux_producer')
+
+onPlatform('linux', () => {
+  describe.only('core/local/steps/linux_producer', () => {
```
taratatach: The `only` made its way back ;-)

enguerran: arf! So sorry for the noise.

enguerran:
@sebn Could you check this PR and give it a review

test/unit/local/steps/linux_producer.js
```diff
@@ -0,0 +1,173 @@
...
+          syncPath
+        })
+        await producer.start()
+        await producer.buffer.pop()
+      })
+
+      it('detect events on folder in temp dir', async () => {
+        const dirname = 'foobaz'
+        const newname = 'foobarbaz'
+        fs.mkdirSync(path.join(syncPath, dirname))
```
nono: Here, `mkdirSync` from `fs` is used, but at other places it's `mkdirsSync` from `fs-extra` that is used. Is there a reason for not using the same everywhere?

enguerran: yes, `mkdirSync` is used to create a single subfolder (`mkdirSync('/temp/foo-xxx', 'foobar')`) when `mkdirsSync` is used to create a suite of subfolders given a path (`mkdirsSync('/temp/foo-xxx', 'foo/bar/baz')`).
The recursive `mkdirSync` will be available in nodejs 10.2.x.
I did write a simple function to compensate for the lack, but @taratatach asked me to use `fs-extra.mkdirsSync`.

PR #1390, opened 2019-01-08
Revert "Enable atom/watcher by default for Windows and Linux"

taratatach:
Reverts cozy-labs/cozy-desktop#1381.

We're not ready for a public release of the new watcher so we'll revert its activation prior to the next public release.


taratatach:
I thought it would be simpler the revert the PR so it can be easily reapplied as well later.
I don't have a strong opinion about that though.

sebn:
Not to be merged before https://github.com/cozy-labs/cozy-desktop/pull/1393

sebn:
@enguerran @nono Any thought about this?

PR #1393, opened 2019-01-09
Do not ignore the initial-scan-done in filterIgnored

nono:


nono:
[Ne pas ignorer initial-scan-done lors de l'étape filterIgnored](https://trello.com/c/AAxLMObX/1378-ne-pas-ignorer-initial-scan-done-lors-de-l%C3%A9tape-filterignored)

PR #1395, opened 2019-01-09
core/local/analysis: Intermediate refactoring

sebn:
This refactoring is a first necessary step to simplify the deeply nested conditionals in analysis (in upcoming #1396). The values newly accepted by `changeFound()` feel a bit weird, but this will be improved in a subsequent PR (they will either be replaced by something more meaningful, or dropped because not necessary anymore).

Please see commit messages for the motivation behind each change.

### DIFF
--- a/core/local/analysis.js
+++ b/core/local/analysis.js
@@ -181,21 +181,23 @@ function analyseEvents (events /*: LocalEvent[] */, pendingChanges /*: LocalChan
               break
             }
           }
-          withChangeByPath(e, samePathChange => {
-            const moveChangeSamePath /*: ?LocalFileMove */ = localChange.maybeMoveFile(samePathChange)
-            if (moveChangeSamePath && moveChangeSamePath.md5sum == null) { // FIXME: if change && change.wip?
-              localChange.convertFileMoveToDeletion(moveChangeSamePath)
-              return
-            }
-            const addChangeSamePath /*: ?LocalFileAddition */ = localChange.maybeAddFile(samePathChange)
-            if (addChangeSamePath && addChangeSamePath.wip) {
-              // $FlowFixMe
-              addChangeSamePath.type = 'Ignored'
-              delete addChangeSamePath.wip
-              delete addChangeSamePath.md5sum
-            }
-            // Otherwise, skip unlink event by multiple moves
-          })
+          changeFound(
+            withChangeByPath(e, samePathChange => {
+              const moveChangeSamePath /*: ?LocalFileMove */ = localChange.maybeMoveFile(samePathChange)
+              if (moveChangeSamePath && moveChangeSamePath.md5sum == null) { // FIXME: if change && change.wip?
+                localChange.convertFileMoveToDeletion(moveChangeSamePath)
+                return
+              }
+              const addChangeSamePath /*: ?LocalFileAddition */ = localChange.maybeAddFile(samePathChange)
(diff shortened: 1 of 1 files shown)


core/local/analysis.js
```diff
@@ -53,7 +53,11 @@ function analyseEvents (events /*: LocalEvent[] */, pendingChanges /*: LocalChan
   const getChangeByPath = (e) => {
     return changesByPath.get(e.path)
   }
-  const changeFound = (c /*: LocalChange */) => {
+  const changeFound = (c /*: LocalChange | true */) => {
+    if (c === true) {
+      // A previous change with was transformed. Nothing more to index.
```
taratatach: What's supposed to go between `with` and `was`?

core/local/analysis.js
```diff
@@ -180,20 +180,22 @@ function analyseEvents (events /*: LocalEvent[] */, pendingChanges /*: LocalChan
               changeFound(localChange.fileDeletion(e))
               break
             }
-            const moveChangeSamePath /*: ?LocalFileMove */ = localChange.maybeMoveFile(getChangeByPath(e))
+          }
```
taratatach: Why are we closing the block before calling `withChangeByPath`?

core/local/analysis.js
```diff
@@ -199,21 +199,23 @@ function analyseEvents (events /*: LocalEvent[] */, pendingChanges /*: LocalChan
...
-                addChangeSamePath.type = 'Ignored'
-              }
-
-              const moveChangeSamePath /*: ?LocalDirMove */ = localChange.maybeMoveFolder(getChangeByPath(e))
-              if (moveChangeSamePath && moveChangeSamePath.wip) {
-                localChange.convertDirMoveToDeletion(moveChangeSamePath)
-              }
+              break
+            }
+            const addChangeSamePath /*: ?LocalDirAddition */ = localChange.maybePutFolder(getChangeByPath(e))
```
taratatach: Why not using `withChangeByPath` here?

core/local/analysis.js
sebn: It will make future diffs easier to read :stuck_out_tongue:

core/local/analysis.js
sebn: Oops typo!

core/local/analysis.js
sebn: I believe we're doing it. At least that's what the global diff shows. Maybe I made a mistake in some intermediate commit... Where did you see it?

core/local/analysis.js
sebn: (fixed)

sebn:
Fixed commit message + other typos and rebased on master.

core/local/analysis.js
sebn: Please note that for some unknown reason existing additions & moves are checked in reverse order in the `unlinkDir` (vs the `unlink` one), which can be confusing regarding what is included in the `withChangeByPath` callback.

core/local/analysis.js
taratatach: OK. What was the point of using blocks for each case anyway?

sebn: I believe vars/consts defined in different `case` branch have the same scope in js. A few of them used to have the same names, hence the necessary block statements to provide separate scope per `case` branch and prevent name clashes.

PR #1399, opened 2019-01-10
core/local/steps/initial_diff: Cover basic cases

taratatach:
To avoid processing a `scan` event following the transformation of a
  `created` event into `renamed` event, we wait before passing it along
  to the next step and remove any `scan` event for the same path.

- [x] PR is not too big

### DIFF
--- a/test/unit/local/steps/initial_diff.js
+++ b/test/unit/local/steps/initial_diff.js
@@ -1,0 +1,100 @@
+/* eslint-env mocha */
+/* @flow */
+
+const should = require('should')
+const Builders = require('../../../support/builders')
+const configHelpers = require('../../../support/helpers/config')
+const pouchHelpers = require('../../../support/helpers/pouch')
+
+const Buffer = require('../../../../core/local/steps/buffer')
+const initialDiff = require('../../../../core/local/steps/initial_diff')
+
+describe('local/steps/initialDiff', () => {
+  let builders
+  let buffer
+
+  before('instanciate config', configHelpers.createConfig)
+  beforeEach('instanciate pouch', pouchHelpers.createDatabase)
+  beforeEach('populate pouch with documents', function () {
+    builders = new Builders({pouch: this.pouch})
+    buffer = new Buffer()
+  })
+  afterEach('clean pouch', pouchHelpers.cleanDatabase)
+  after('clean config directory', configHelpers.cleanConfig)
+
+  it('detects documents moved while client was stopped', async function () {
+    await builders.metadir().path('foo').ino(1).create()
+    await builders.metafile().path('fizz').ino(2).create()
+
+    const bar = builders.event().action('scan').kind('directory').path('bar').stats({ ino: 1 }).build()
+    const buzz = builders.event().action('scan').kind('file').path('buzz').stats({ ino: 2 }).build()
+    buffer.push([bar, buzz])
+    buffer = initialDiff(buffer, { pouch: this.pouch })
+
+    const events = await buffer.pop()
+    should(events).deepEqual([
+      builders.event(bar).action('renamed').oldPath('foo').build(),
+      builders.event(buzz).action('renamed').oldPath('fizz').build()
+    ])
+  })
+
+  it('detects documents moved while client is doing initial scan', async function () {
+    await builders.metadir().path('foo').ino(1).create()
+    await builders.metafile().path('foo/baz').ino(2).create()
+    await builders.metadir().path('bar').ino(3).create()
+
+    const foo = builders.event().action('scan').kind('directory').path('foo').stats({ ino: 1 }).build()
+    const barbaz = builders.event().action('created').kind('file').path('bar/baz').stats({ ino: 2 }).build()
+    buffer.push([foo, barbaz])
+    const bar = builders.event().action('scan').kind('directory').path('bar').stats({ ino: 3 }).build()
+    buffer.push([
+      bar,
+      builders.event().action('scan').kind('file').path('bar/baz').stats({ ino: 2 }).build()
+    ])
+    buffer = initialDiff(buffer, { pouch: this.pouch })
+
+    const events = [].concat(
+      await buffer.pop(),
+      await buffer.pop()
+    )
+    should(events).deepEqual([
+      foo,
+      builders.event(barbaz).action('renamed').oldPath('foo/baz').build(),
+      bar
+    ])
+  })
+
+  it('detects documents replaced by another one of a different kind while client was stopped', async function () {
+    await builders.metadir().path('foo').ino(1).create()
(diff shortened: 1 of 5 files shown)


test/unit/local/steps/initial_diff.js
```diff
@@ -0,0 +1,100 @@
...
+    builders.metafile().path('foo/baz').ino(2).create()
+    builders.metadir().path('bar').ino(3).create()
+
+    const foo = builders.event().action('scan').kind('directory').path('foo').stats({ ino: 1 }).build()
+    const barbaz = builders.event().action('created').kind('file').path('bar/baz').stats({ ino: 2 }).build()
+    buffer.push([foo, barbaz])
+    const bar = builders.event().action('scan').kind('directory').path('bar').stats({ ino: 3 }).build()
+    buffer.push([
+      bar,
+      builders.event().action('scan').kind('file').path('bar/baz').stats({ ino: 2 }).build()
```
sebn: nitpicking: some events are inline, others are not.

test/unit/local/steps/initial_diff.js
```diff
@@ -0,0 +1,100 @@
...
+  })
+  afterEach('clean pouch', pouchHelpers.cleanDatabase)
+  after('clean config directory', configHelpers.cleanConfig)
+
+  it('detects documents moved while client was stopped', async function () {
+    builders.metadir().path('foo').ino(1).create()
+    builders.metafile().path('baz').ino(2).create()
+
+    const bar = builders.event().action('scan').kind('directory').path('bar').stats({ ino: 1 }).build()
+    const buz = builders.event().action('scan').kind('file').path('buz').stats({ ino: 2 }).build()
```
sebn: nitpicking: *bar*, *baz*  and *buz* names are very similar (error-prone).

test/unit/local/steps/initial_diff.js
```diff
@@ -0,0 +1,100 @@
...
+    buffer.push([bar, buz])
+    buffer = initialDiff(buffer, { pouch: this.pouch })
+
+    const events = await buffer.pop()
+    should(events).deepEqual([
+      builders.event(bar).action('renamed').oldPath('foo').build(),
+      builders.event(buz).action('renamed').oldPath('baz').build()
```
sebn: nitpicking: Although reuse makes sense for docs since we're often building the next version based on the previous one, I wonder whether it makes sense for an event to build upon another one...

test/unit/local/steps/initial_diff.js
```diff
@@ -0,0 +1,100 @@
...
+  beforeEach('populate pouch with documents', function () {
+    builders = new Builders({pouch: this.pouch})
+    buffer = new Buffer()
+  })
+  afterEach('clean pouch', pouchHelpers.cleanDatabase)
+  after('clean config directory', configHelpers.cleanConfig)
+
+  it('detects documents moved while client was stopped', async function () {
+    builders.metadir().path('foo').ino(1).create()
+    builders.metafile().path('baz').ino(2).create()
```
sebn: blocker: Shouldn't we `await` all those *async* `BaseMetadataBuilder#create()` calls everywhere?

test/unit/local/steps/initial_diff.js
```diff
@@ -0,0 +1,100 @@
...
+
+    const foo = builders.event().action('scan').kind('file').path('foo').stats({ ino: 2 }).build()
+    const buz = builders.event().action('scan').kind('directory').path('buz').stats({ ino: 3 }).build()
+    buffer.push([foo, buz])
+    buffer = initialDiff(buffer, { pouch: this.pouch })
+
+    const events = await buffer.pop()
+    should(events).deepEqual([
+      builders.event(foo).action('renamed').oldPath('bar').build(),
+      builders.event(buz).action('renamed').oldPath('baz').build()
```
sebn: Shouldn't we have `deleted` events for `foo` and `buz` preceding the `renamed` ones?

test/unit/local/steps/initial_diff.js
taratatach: I inline them when I can but start putting them on multiple lines when they go beyond 80 characters (I've been a bit more lenient about this recently and only do it if they go way beyond that limit).

Do you have a different approach?

test/unit/local/steps/initial_diff.js
taratatach: I've done it in some cases where I was testing that an event is replaced by another one with very few attributes changing.
But I haven't done so often.

test/unit/local/steps/initial_diff.js
taratatach: We probably do indeed. I forgot to add `await` since the event builders don't need it.
I'll fix that.

test/unit/local/steps/initial_diff.js
taratatach: No we only have the `renamed` event with the new watcher.

test/unit/local/steps/initial_diff.js
sebn: I either inline everything or nothing, I find it easier to scan since everything looks the same.

test/unit/local/steps/initial_diff.js
sebn: Won't this generate conflits for both [files](https://github.com/cozy-labs/cozy-desktop/blob/master/core/merge.js#L281-L282) and [folders](https://github.com/cozy-labs/cozy-desktop/blob/master/core/merge.js#L319-L323)?

test/unit/local/steps/initial_diff.js
sebn: Not a blocker.

test/unit/local/steps/initial_diff.js
sebn: Possibly just a matter of taste & not a blocker anyway.

test/unit/local/steps/initial_diff.js
sebn: (not a blocker)

test/unit/local/steps/initial_diff.js
taratatach: Possibly. I didn't think much about the implications of the code I was testing on other parts of the application.
Maybe we just need to mark those events as overwriting?

taratatach: Actually, I believe it won't be an issue for [files](https://github.com/cozy-labs/cozy-desktop/blob/9a1d05f50dace7e953b5e74ed2c70fea428bc2eb/core/merge.js#L278-L280) but will for folders as we don't have the equivalent branch in `moveFolderAsync` :/.

sebn: A separate issue was extracted, it will be fixed in a future PR.

PR #1400, opened 2019-01-10
Fix the GUI sync status with the new watcher

nono:


nono:
[[3.12.0alpha2] "Synchronize (1 left)"](https://trello.com/c/zk7FWNxn/1384-3120alpha2-synchronize-1-left)

nono:
@enguerran I've updated the commit.Is it better like this?

core/local/atom_watcher.js
```diff
@@ -78,10 +78,20 @@ module.exports = class AtomWatcher {
...
+    const scanDone = new Promise((resolve) => {
       this.events.on('initial-scan-done', resolve)
     })
+    scanDone.then(async () => {
+      let target = -1
+      try {
+        target = (await this.pouch.db.changes({limit: 1, descending: true})).last_seq
+      } catch (err) { /* ignore err */ }
+      this.events.emit('sync-target', target)
+      this.events.emit('local-end')
```
taratatach: Don't we use that event to notify the end of a scan phase rather than the event of the initial scan?

enguerran:
> @enguerran I've updated the commit.Is it better like this?

Yes, I do understand that when the scan is done, we add an event listener, and after that we do some DB access and emit some events. Thanks.

core/local/atom_watcher.js
nono: It's a part of the code that I don't know very well. What do you call a scan phase?

taratatach: I should have said a Sync phase instead. We emit this event in the `onFlush` method of the chokidar watcher.
So we emit it every time we've had 10 seconds without any event on the filesystem.

But here it looks to me like we'll call it only once, right after the initial scan was done. Won't that affect the UI for subsequent synchronisations?

nono: We don't have this 10+ seconds delay on the new watcher, so I think it is acceptable to do it once.

taratatach: I'm just wondering what the expectations are behind this event and if the UI needs it to correctly display the number of changes to process.
If all is well, I'm completely fine with the change.

nono: As I understand, `local-start` and `local-end` are used to show an icon with a spinning cloud, but it's just one of the 3 conditions for that (the 2 others are we are fetching changes from the remote and the sync class has work to do).

`sync-target` is used to tell the number of _things_ to do by the client.

taratatach: So that might be the line I should have cared about with the same reasoning :D

PR #1401, opened 2019-01-10
Avoid a race condition on pouchdb in initialDiff

nono:


nono:
[Problème au démarrage avec initialDIff](https://trello.com/c/edGV9nqk/1388-probl%C3%A8me-au-d%C3%A9marrage-avec-initialdiff)

nono:
> Can't we have await pouch.byRecursivePathAsync('') wait for the initialization of pouch instead?

I don't see how to wait that the views are created in PouchDB. We can try the request, and if it fails wait and try again, but it looks even more hackish.

taratatach:
> I don't see how to wait that the views are created in PouchDB

How do you know for sure that they're created before the watcher is started then?

nono:
Well, if you call PouchDB to create a view, it tells you when it's done, so we can know when the view is created if we are on the path for creating the views. But currently, the watcher is not on this execution path, and I think it would need a lot of refactoring to do so.

core/local/steps/initial_diff.js
```diff
@@ -93,11 +103,13 @@ async function initialDiff (buffer, out, pouch) {
     }
 
     out.push(batch)
+    events = await buffer.pop()
```
enguerran: `const` are block scoped, using a variable from outside the `while` block may lead to unwanted behavior.

nono: `events` is declared outside of the `while` loop (line 26)

nono:
> But if I read right, we want to be sure that pouch is started and ready to use before running things.

I want to be sure that the views (a view is something like an index) have been created in PouchDB before making queries to it.

> a promise that resolves with db info: https://pouchdb.com/api.html#database_information

Db info doesn't give information about the views from what I see from the documentation.

> events emitted by pouchDB: https://pouchdb.com/api.html#events

I don't see events for views.

core/local/steps/initial_diff.js
enguerran: That's right, that's what I was saying. What I meant was that there is no reason to declare this variable outside the `while' loop.

enguerran: No, I don't. We may be somewhere where the only way to do it is what you did.

PR #1403, opened 2019-01-11
Add some property based tests with 2 cozy-desktop clients

nono:

### DIFF
--- a/gui/main.js
+++ b/gui/main.js
@@ -289,10 +289,12 @@ const startSync = (force, ...args) => {
   })
 }
 
-const shouldExit = app.makeSingleInstance(() => showWindow())
-if (shouldExit) {
-  log.warn('Cozy Drive is already running. Exiting...')
-  app.exit()
+if (!process.env.COZY_DESKTOP_PROPERTY_BASED_TESTING) {
+  const shouldExit = app.makeSingleInstance(() => showWindow())
+  if (shouldExit) {
+    log.warn('Cozy Drive is already running. Exiting...')
+    app.exit()
+  }
 }
 
 const dumbhash = (k) => k.split('').reduce((a, c) => ((a << 5) - a + c.charCodeAt(0)) | 0)
@@ -340,7 +342,12 @@ app.on('ready', () => {
       updaterWindow.hide()
       showWindowStartApp()
     })
-    updaterWindow.checkForUpdates()
+    if (process.env.COZY_DESKTOP_PROPERTY_BASED_TESTING) {
+      updaterWindow.hide()
+      showWindowStartApp()
+    } else {
+      updaterWindow.checkForUpdates()
+    }
 
     // Os X wants all application to have a menu
     Menu.setApplicationMenu(buildAppMenu(app))
--- a/test/generate_property_json.js
+++ b/test/generate_property_json.js
@@ -7,6 +7,22 @@ const nbInitOps = 32
 const nbRunOps = 256
 const specialChars = [':', '-', 'é', ' ', '%', ',', '&', '@', 'É', 'Ç']
 
+const localOps = [
+  [5, createNewDir],
+  [3, createNewFile],
+  [1, recreateDeletedDir],
+  [1, recreateDeletedFile],
+  [1, updateFile],
+  [2, mvToNewPath],
+  [3, mvToDeletedPath],
+  [1, mvToOutside],
+  [1, mvFromOutside],
+  [5, rm],
+  [3, addReference],
+  [1, stopOrRestart],
+  [2, sleep]
+]
+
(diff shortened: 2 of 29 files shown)


nono:
[Ajouter du property based testing avec 2 clients](https://trello.com/c/aleBosy6/1380-ajouter-du-property-based-testing-avec-2-clients)

gui/main.js
```diff
@@ -340,7 +342,12 @@ app.on('ready', () => {
       updaterWindow.hide()
       showWindowStartApp()
     })
-    updaterWindow.checkForUpdates()
+    if (process.env.COZY_DESKTOP_PROPERTY_BASED_TESTING) {
```
enguerran: I did not undestand why this condition has been added.

test/generate_property_json.js
```diff
@@ -122,19 +145,19 @@ function addReference () {
   return { op: 'reference', path: p }
 }
 
-function stopOrRestart () {
+function stopOrRestartWatcher () {
   if (running) {
     running = false
-    return { op: 'stop' }
+    return { op: 'stop_watcher' }
```
enguerran: Shouldn't commands be variables.

test/generate_property_json.js
```diff
@@ -122,19 +145,19 @@ function addReference () {
...
     running = true
-    return { op: 'restart' }
+    return { op: 'restart_watcher' }
   }
 }
 
 function sleep () {
-  const s = 2 ** (1 + faker.random.number(3))
-  return { op: 'sleep', duration: 1000 * s }
+  const s = 2 ** (6 + faker.random.number(8))
```
enguerran: Magic number, hard to maintain and understand why, what and how.

test/generate_property_json.js
```diff
@@ -157,42 +180,60 @@ function init (ops) {
...
+  run(ops, localOps)
+  if (!running) {
+    ops.push({ op: 'restart' })
+  }
   console.log(JSON.stringify(ops))
 }
 
-generate()
+function generateTwoClients () {
+  let result = { desktop: [], laptop: [] }
```
enguerran: When variables are not reassigned, a lot of JS guidelines recommand using `const` instead of `let`: https://eslint.org/docs/rules/prefer-const

test/generate_property_json.js
```diff
@@ -157,42 +180,60 @@ function init (ops) {
...
+  // Flow doesn't want of:
+  // for (let ops of Object.values(result)) {
```
enguerran: Which version of `flow` does not want this `Object.values({ desktop: [], laptop: [] }`?
I checked on the flowjs online editor: https://flow.org/try/#0PQKgBAAgZgNg9gdzCYAoVBjOA7AzgFzDgCMArMAXjAG9UwwBDAJyYEYAuMAbQF0AaOoxYAmTr1QBfdFDhMAFDACmhAJZEoYAPJlFGfADoAbgxgBXRbjklSAShs1UASElA

taratatach: You need to add some calls to Array methods: https://flow.org/try/#0PQKgBAAgZgNg9gdzCYAoVBjOA7AzgFzDgCMArMAXjAG9UwwBDAJyYEYAuMAbQF0AaOoxYAmTr1QBfdFDhMAFDACmhAJZEoYAPJlFGfADoAbgxgBXRbjklSAShs1UASBXYV+OSpuTpp7HpU4YC5uVgAOuPa09Fh4hPhw8ZRgwoJw4fqhprgAFnLx8V4SQA

nono: My comment wasn't clear: you can use `for (let ops of Object.values(result)) {`, but `ops` is seen `mixed` and not as an array by flow.

test/generate_property_json.js
nono: We are in a script that generates some random scenario for property based testing. There are no right or wrong values here: if you change the number, it will change what is tester. With short sleeps, race conditions are more tested, with long sleeps, it is more the general synchronization workflow.

gui/main.js
nono: I didn't find why, but during the tests, the client was sometimes blocked on the update check.

test/generate_property_json.js
nono: I don't understand what you mean here.

nono:
> Very big PR proposed without any decription nor explanation.

Yup, sorry, I was short in time.

test/generate_property_json.js
enguerran:
```js
- const STOP = 'stop';
+ const STOP = 'stop_watcher';

return { op: STOP }
```

nono: OK, but why do you want that?

PR #1407, opened 2019-01-16
test/unit/local/steps/dispatch: Cover all events

taratatach:
Based on #1399 

We make sure that only the expected Prep methods are called for each
  combination of event action and kind and that they receive the
  expected parameters.

Some changes were made to `core/local/steps/dispatch` to help test it in isolation.
By using an `asyncMap` instead of an `asyncForEach` and by returning
  the resulting buffer, we can detect the end of a batch processing from
  the caller of the step.

- [x] PR is not too big

### DIFF
--- a/test/unit/local/steps/dispatch.js
+++ b/test/unit/local/steps/dispatch.js
@@ -1,0 +1,455 @@
+/* eslint-env mocha */
+/* @flow */
+
+const crypto = require('crypto')
+
+/*::
+import type { Stub, Call } from 'sinon'
+
+type DispatchedCalls = {
+  [string]: Array<Array<any>>
+}
+*/
+
+const should = require('should')
+const sinon = require('sinon')
+
+const Builders = require('../../../support/builders')
+const configHelpers = require('../../../support/helpers/config')
+const pouchHelpers = require('../../../support/helpers/pouch')
+
+const SyncState = require('../../../../core/syncstate')
+const Prep = require('../../../../core/prep')
+const Buffer = require('../../../../core/local/steps/buffer')
+const dispatch = require('../../../../core/local/steps/dispatch')
+
+function dispatchedCalls (obj /*: Stub */) /*: DispatchedCalls */ {
+  const methods = Object.getOwnPropertyNames(obj).filter(m => typeof obj[m] === 'function')
+
+  const dispatchedCalls = {}
+  for (const method of methods) {
+    const calls /*: Array<Call> */ = obj[method].getCalls()
+
+    for (const call of calls) {
+      if (!dispatchedCalls[method]) dispatchedCalls[method] = []
+      dispatchedCalls[method].push(call.args)
+    }
+  }
+
+  return dispatchedCalls
+}
+
+describe('dispatch', function () {
+  let builders
+  let buffer
+  let events
+  let prep
+
+  before('instanciate config', configHelpers.createConfig)
+  beforeEach('instanciate pouch', pouchHelpers.createDatabase)
+  beforeEach('populate pouch with documents', function () {
+    builders = new Builders({pouch: this.pouch})
+    buffer = new Buffer()
+
+    events = sinon.createStubInstance(SyncState)
+    prep = sinon.createStubInstance(Prep)
+  })
+  afterEach('clean pouch', pouchHelpers.cleanDatabase)
+  after('clean config directory', configHelpers.cleanConfig)
+
+  context('when buffer contains an initial-scan-done event', () => {
+    beforeEach(() => {
+      buffer.push([
+        builders.event().action('initial-scan-done').build()
+      ])
+    })
+
+    it('emits an initial-scan-done event via the emitter', async function () {
+      const dispatchedBuffer = dispatch(buffer, { events, prep, pouch: this.pouch })
+      await dispatchedBuffer.pop()
+
+      should(dispatchedCalls(events)).deepEqual({
+        emit: [
+          ['initial-scan-done']
+        ]
+      })
+    })
+
+    it('does not call any Prep method', async function () {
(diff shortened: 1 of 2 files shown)


test/unit/local/steps/dispatch.js
```diff
@@ -0,0 +1,389 @@
...
+    })
+  })
+
+  context('when buffer contains a created file event', () => {
+    const filePath = 'foo'
+    const stats = { ino: 1, mtime: new Date(), ctime: new Date() }
+
+    beforeEach(() => {
+      buffer.push([
+        builders.event().action('created').kind('file').path(filePath).stats(stats).build()
```
sebn: Maybe just `.ino(1)` in case we don't care about the dates?

test/unit/local/steps/dispatch.js
```diff
@@ -0,0 +1,389 @@
...
+      ])
+    })
+
+    it('triggers a call to addFileAsync with a file Metadata object', async function () {
+      const metafile = { path: filePath, ino: stats.ino, docType: 'file' }
```
sebn: Maybe use `builders.metafile()`?
Maybe call it `doc` since it's the usual name in the codebase?

test/unit/local/steps/dispatch.js
```diff
@@ -0,0 +1,389 @@
...
+  context('when buffer contains an initial-scan-done event', () => {
+    beforeEach(() => {
+      buffer.push([
+        builders.event().action('initial-scan-done').build()
+      ])
+    })
+
+    it('emits an initial-scan-done event via the emitter', async function () {
+      const dispatchedBuffer = dispatch(buffer, { events, prep, pouch: this.pouch })
+      await dispatchedBuffer.pop()
```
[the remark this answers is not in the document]
taratatach: We're popping the dispatched events. I modified `dispatch` so it would return a buffer with every event dispatched untouched so we could know when they've been dispatched.

test/unit/local/steps/dispatch.js
taratatach: We kinda do care about the dates unfortunately. If they're not present, `metadata.buildFile` will throw an error.

test/unit/local/steps/dispatch.js
taratatach: I've tried to use `builders.metafile` at first to do an exact comparison with the actual parameter given to the `Prep` method but it was hard to get it right.
Since I moved to a partial comparison I thought it would be as convenient to create a partial `Metadata` object instead (it also shows it's not a complete match).

As for the name, I thought I'd have multiple documents in each test at first but since it's not the case I'll make the change.

core/local/steps/dispatch.js
```diff
@@ -19,11 +25,12 @@ let events, target, pouch, actions
 // Dispatch takes a buffer of AtomWatcherEvents batches, and calls Prep for
 // each event. It needs to fetch the old documents from pouchdb in some cases
 // to have all the data expected by prep/merge.
-module.exports = function (buffer /*: Buffer */, opts /*: { events: EventEmitter, prep: Prep, pouch: Pouch } */) {
+module.exports = function (buffer /*: Buffer */, opts /*: DispatchOptions */) /*: Buffer */ {
   events = opts.events
   target = opts.prep
   pouch = opts.pouch
-  buffer.asyncForEach(async (batch) => {
```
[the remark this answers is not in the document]
taratatach: I don't find the `Buffer` `asyncForEach` and `forEach` methods super straight forward and didn't want to remove them in case @nono had a specific use case in mind that I would see.
But I also thought about removing them since they aren't used anywhere (any more).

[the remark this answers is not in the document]
nono: It's OK for me to remove them if they are no longer used

test/unit/local/steps/dispatch.js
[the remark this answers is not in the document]
sebn: I can't stop thinking that a `dispatch()` function extracted from the current main one (which could be renamed to `dispatchLoop()`) would be far easier to `await` in tests.

test/unit/local/steps/dispatch.js
sebn: Sure, but isn't it the builder's job to set them?

test/unit/local/steps/dispatch.js
sebn: Couldn't [a helper](https://github.com/cozy-labs/cozy-desktop/pull/1407#discussion_r248702048) solve the comparison issue?

test/unit/local/steps/dispatch.js
```diff
@@ -0,0 +1,389 @@
...
+
+    it('triggers a call to addFileAsync with a file Metadata object', async function () {
+      const metafile = { path: filePath, ino: stats.ino, docType: 'file' }
+
+      const dispatchedBuffer = dispatch(buffer, { events, prep, pouch: this.pouch })
+      await dispatchedBuffer.pop()
+
+      shouldHaveCalledOn(prep, ['addFileAsync'])
+      should(prep.addFileAsync).have.been.calledOnce()
+      should(prep.addFileAsync).have.been.calledWithMatch('local', metafile)
```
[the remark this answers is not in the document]
sebn: > I'll have to figure out something else.
A [helper](https://github.com/cozy-labs/cozy-desktop/pull/1407#discussion_r248702048)? :wink:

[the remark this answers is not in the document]
taratatach: Yes, I'm working on it ;-)

test/unit/local/steps/dispatch.js
taratatach: Maybe this should be part of the constructor indeed.

PR #1406, opened 2019-01-16
core/local: Single type change builders

sebn:
Some functions that build `LocalChange`s may handle many edge cases and return different change types. This makes it harder to understand them & to add type annotations (flow errors end up being quite hard to understand/debug).

- [x] PR is not too big
- [x] it improves UX & DX in some way
- [x] it includes tests matching the implementation changes (I'm confident there is at least 1 test that exercise each of those cases)
- [ ] it includes relevant documentation

PR #1408, opened 2019-01-16
[CHOKIDAR] core/local: Side-effect free change detection

sebn:
**Needs https://github.com/cozy-labs/cozy-desktop/pull/1406**

In the previous implementation, we were mutating previously detected changes (including their type), which was confusing flow (and maybe developers).

The change detection functions now return either a *new change* or a *change replacement*.

- [x] PR is not too big

### DIFF
--- a/core/local/analysis.js
+++ b/core/local/analysis.js
@@ -17,7 +17,9 @@ import type {
   LocalDirMove,
   LocalFileAddition,
   LocalFileDeletion,
-  LocalFileMove
+  LocalFileMove,
+  NewChange,
+  ChangeReplacement
 } from './change'
 */
 
@@ -121,8 +123,11 @@ function analyseEvents (events /*: LocalEvent[] */, pendingChanges /*: LocalChan
 
       const result = analyseEvent(e, changesFound)
       if (result == null) continue // No change was found. Skip event.
-      if (result === true) continue // A previous change was transformed. Nothing more to do.
-      changesFound.put(result) // A new change was found
+      if (result.replacedChange) {
+        changesFound.replace(result.replacedChange, result.newChange)
+      } else {
+        changesFound.put(result.newChange)
+      }
     } catch (err) {
       const sentry = err.name === 'InvalidLocalMoveEvent'
       log.error({err, path: e.path, sentry})
@@ -137,7 +142,7 @@ function analyseEvents (events /*: LocalEvent[] */, pendingChanges /*: LocalChan
   return changes
 }
 
-function analyseEvent (e /*: LocalEvent */, previousChanges /*: LocalChangeMap */) /*: ?LocalChange|true */ {
+function analyseEvent (e /*: LocalEvent */, previousChanges /*: LocalChangeMap */) /*: ?NewChange|ChangeReplacement */ {
   const sameInodeChange = previousChanges.findByInode(getInode(e))
 
   switch (e.type) {
--- a/core/local/analysis.js
+++ b/core/local/analysis.js
@@ -42,7 +42,6 @@ const panic = (context, description) => {
 
 class LocalChangeMap {
   /*::
-  changes: LocalChange[]
   changesByInode: Map<number, LocalChange>
   changesByPath: Map<string, LocalChange>
   */
@@ -52,7 +51,6 @@ class LocalChangeMap {
   }
 
   _clear () {
-    this.changes = []
     this.changesByInode = new Map()
     this.changesByPath = new Map()
   }
@@ -70,13 +68,11 @@ class LocalChangeMap {
   put (c /*: LocalChange */) {
     this.changesByPath.set(c.path, c)
     if (typeof c.ino === 'number') this.changesByInode.set(c.ino, c)
-    else this.changes.push(c)
   }
 
   remove (c /*: LocalChange */) {
     this.changesByPath.delete(c.path)
     if (typeof c.ino === 'number') this.changesByInode.delete(c.ino)
-    else _.pull(this.changes, c) // OPTIMIZE: Do not scan the whole Array
   }
 
   replace (replaced /*: LocalChange */, replacement /*: LocalChange */) {
@@ -85,16 +81,14 @@ class LocalChangeMap {
   }
 
   flush () /*: LocalChange[] */ {
-    const changes = this.changes
-    for (let a of this.changesByInode.values()) changes.push(a)
+    const changes = this.changesByInode.values()
     this._clear()
-    return changes
+    return Array.from(changes)
   }
 }
 
 function analyseEvents (events /*: LocalEvent[] */, pendingChanges /*: LocalChange[] */) /*: LocalChange[] */ {
   const stopMeasure = measureTime('LocalWatcher#analyseEvents')
-  // OPTIMIZE: new Array(events.length)
   const changesFound = new LocalChangeMap()
 
   if (pendingChanges.length > 0) {
--- a/core/local/analysis.js
+++ b/core/local/analysis.js
@@ -71,6 +71,17 @@ class LocalChangeMap {
     else this.changes.push(c)
   }
 
+  remove (c /*: LocalChange */) {
+    this.changesByPath.delete(c.path)
+    if (typeof c.ino === 'number') this.changesByInode.delete(c.ino)
+    else _.pull(this.changes, c) // OPTIMIZE: Do not scan the whole Array
+  }
+
+  replace (replaced /*: LocalChange */, replacement /*: LocalChange */) {
+    this.remove(replaced)
+    this.put(replacement)
+  }
+
   flush () /*: LocalChange[] */ {
     const changes = this.changes
     for (let a of this.changesByInode.values()) changes.push(a)
--- a/core/local/change.js
+++ b/core/local/change.js
@@ -151,6 +151,17 @@ export type LocalChange =
   | LocalFileMove
   | LocalFileUpdate
   | LocalIgnored
+
+export type NewChange = {|
+  newChange: LocalChange
+|}
+export type ChangeReplacement = {|
+  replacedChange: LocalChange,
+  newChange: LocalChange
+|}
+export type AnalysisResult =
+  | NewChange
+  | ChangeReplacement
 */
 
 const sideName = 'local'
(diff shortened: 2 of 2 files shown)


core/local/analysis.js
```diff
@@ -71,6 +73,17 @@ class LocalChangeMap {
     else this.changes.push(c)
   }
 
+  remove (c /*: LocalChange */) {
+    this.changesByPath.delete(c.path)
+    if (typeof c.ino === 'number') this.changesByInode.delete(c.ino)
+    else _.pull(this.changes, c) // OPTIMIZE: Do not scan the whole Array
```
sebn: I wonder about this line. Maybe the equivalent to the old code would be not to have an `else` branch at all... But it just seems wrong.

Anyway, I believe I already tried to drop the whole `changes` property and all tests were still passing. I'll try again in another PR.

sebn: We probably can only have one change at a time for the same path currently, since all detection functions taking a `samePathChange` param return a `ChangeReplacement`.

taratatach: Don't we need to remove the change from `this.changes` in every case? I thought this array was an aggregation of all detected changes.

core/local/analysis.js
```diff
@@ -126,7 +142,7 @@ function analyseEvents (events /*: LocalEvent[] */, pendingChanges /*: LocalChan
   return changes
 }
 
-function analyseEvent (e /*: LocalEvent */, previousChanges /*: LocalChangeMap */) /*: ?LocalChange|true */ {
+function analyseEvent (e /*: LocalEvent */, previousChanges /*: LocalChangeMap */) /*: ?NewChange|ChangeReplacement */ {
```
taratatach: Do we have a way to avoid having an optional result here?

core/local/change.js
```diff
@@ -208,98 +219,98 @@ function isChildUpdate (a /*: LocalChange */, b /*: LocalChange */) { return chi
...
-  if (e.old) change.old = e.old
-  if (e.wip) change.wip = e.wip
-  return change
+  if (e.old) newChange.old = e.old
+  if (e.wip) newChange.wip = e.wip
+  return {newChange}
 }
 
-function fileMoveFromUnlinkAdd (sameInodeChange /*: ?LocalChange */, e /*: LocalFileAdded */) /*: * */ {
+function fileMoveFromUnlinkAdd (sameInodeChange /*: ?LocalChange */, e /*: LocalFileAdded */) /*: ?NewChange */ {
```
taratatach: Shouldn't we return a `ChangeReplacement` here since we're replacing an `UnlinkChange` with a `MoveChange`?

core/local/change.js
```diff
@@ -208,98 +219,98 @@ function isChildUpdate (a /*: LocalChange */, b /*: LocalChange */) { return chi
...
     {oldpath: unlinkChange.path, path: e.path, ino: unlinkChange.ino},
-    `unlink + add = FileMove${fileMove.update ? '(update)' : ''}`
+    `unlink + add = FileMove${newChange.update ? '(update)' : ''}`
   )
-  return fileMove
+  return {newChange}
 }
 
-function dirMoveFromUnlinkAdd (sameInodeChange /*: ?LocalChange */, e /*: LocalDirAdded */) /*: * */ {
+function dirMoveFromUnlinkAdd (sameInodeChange /*: ?LocalChange */, e /*: LocalDirAdded */) /*: ?NewChange */ {
```
taratatach: Same here.

core/local/change.js
```diff
@@ -308,28 +319,30 @@ function dirMoveFromUnlinkAdd (sameInodeChange /*: ?LocalChange */, e /*: LocalD
...
     stats: e.stats,
     old: unlinkChange.old,
     ino: unlinkChange.ino,
     wip: e.wip
   })
+  return {newChange}
 }
 
-function fileMoveFromAddUnlink (sameInodeChange /*: ?LocalChange */, e /*: LocalFileUnlinked */) /*: * */ {
+function fileMoveFromAddUnlink (sameInodeChange /*: ?LocalChange */, e /*: LocalFileUnlinked */) /*: ?NewChange */ {
```
taratatach: Same here.

core/local/change.js
```diff
@@ -308,28 +319,30 @@ function dirMoveFromUnlinkAdd (sameInodeChange /*: ?LocalChange */, e /*: LocalD
...
     md5sum: addChange.md5sum,
     old: e.old,
     ino: addChange.ino,
     wip: addChange.wip
   })
+  return {newChange}
 }
 
-function fileMoveFromFileDeletionChange (sameInodeChange /*: ?LocalChange */, e /*: LocalFileUpdated */) {
+function fileMoveFromFileDeletionChange (sameInodeChange /*: ?LocalChange */, e /*: LocalFileUpdated */) /*: ?NewChange */ {
```
taratatach: Same here.

core/local/change.js
```diff
@@ -339,48 +352,49 @@ function fileMoveFromFileDeletionChange (sameInodeChange /*: ?LocalChange */, e
...
     ino: newDst.stats.ino,
     wip: e.wip
   })
-
-  return fileMove
+  return {newChange}
 }
 
-function dirMoveFromAddUnlink (sameInodeChange /*: ?LocalChange */, e /*: LocalDirUnlinked */) /*: * */ {
+function dirMoveFromAddUnlink (sameInodeChange /*: ?LocalChange */, e /*: LocalDirUnlinked */) /*: ?NewChange */ {
```
taratatach: Same here.

core/local/change.js
```diff
@@ -339,48 +352,49 @@ function fileMoveFromFileDeletionChange (sameInodeChange /*: ?LocalChange */, e
...
     stats: addChange.stats,
     old: e.old,
     ino: addChange.ino,
     wip: addChange.wip
   })
+  return {newChange}
 }
 
-function fileMoveIdentical (sameInodeChange /*: ?LocalChange */, e /*: LocalFileUpdated */) /*: * */ {
+function fileMoveIdentical (sameInodeChange /*: ?LocalChange */, e /*: LocalFileUpdated */) /*: ?NewChange */ {
```
taratatach: Same here.

core/local/change.js
```diff
@@ -389,69 +403,87 @@ function fileMoveIdenticalOffline (dstEvent /*: LocalFileAdded */) /*: ?LocalFil
     old: srcDoc,
     ino: dstEvent.stats.ino
   } /*: LocalFileMove */)
+  return {newChange}
 }
 
-function dirRenamingCaseOnlyFromAddAdd (sameInodeChange /*: ?LocalChange */, e /*: LocalDirAdded */) /*: * */ {
+function dirRenamingCaseOnlyFromAddAdd (sameInodeChange /*: ?LocalChange */, e /*: LocalDirAdded */) /*: ?NewChange */ {
```
taratatach: Same here.

core/local/change.js
```diff
@@ -389,69 +403,87 @@ function fileMoveIdenticalOffline (dstEvent /*: LocalFileAdded */) /*: ?LocalFil
...
-  moveChange.path = e.path
-  moveChange.stats = e.stats
-  moveChange.md5sum = e.md5sum
+
+  const newChange /*: LocalFileMove */ = {
+    sideName,
+    type: 'FileMove',
+    path: e.path,
+    old: moveChange.old,
+    ino: moveChange.ino,
```
taratatach: What's the difference between `ino` and `stats.ino`?

taratatach: Shouldn't we use `e.stats.ino` here instead?

core/local/change.js
```diff
@@ -389,69 +403,87 @@ function fileMoveIdenticalOffline (dstEvent /*: LocalFileAdded */) /*: ?LocalFil
...
+    stats: e.stats,
+    md5sum: e.md5sum,
+    wip: moveChange.wip
+  }
+
+  const {update, overwrite} = moveChange
+  if (update) newChange.update = update
+  if (overwrite) newChange.overwrite = overwrite
 
   if (e.md5sum) {
```
taratatach: `if (newChange.md5sum) {` ?

core/local/change.js
```diff
@@ -461,104 +493,162 @@ function dirMoveOverwriteOnMacAPFS (sameInodeChange /*: ?LocalChange */, e /*: L
...
   const moveChange /*: ?LocalDirMove */ = maybeMoveFolder(sameInodeChange)
   if (!moveChange) return
-  moveChange.path = e.path
-  moveChange.stats = e.stats
+  const newChange /*: LocalDirMove */ = {
+    sideName,
+    type: 'DirMove',
+    path: e.path,
+    old: moveChange.old,
+    ino: moveChange.ino,
```
taratatach: `e.stats.ino` ?

core/local/change.js
```diff
@@ -461,104 +493,162 @@ function dirMoveOverwriteOnMacAPFS (sameInodeChange /*: ?LocalChange */, e /*: L
...
-  return true
-}
-
-function convertFileMoveToDeletion (samePathChange /*: ?LocalChange */) {
+  const newChange /*: LocalFileMove */ = {
+    sideName,
+    type: 'FileMove',
+    path: moveChange.path,
+    old: moveChange.old,
+    ino: moveChange.ino,
```
taratatach: `e.stats.ino` ?

core/local/change.js
```diff
@@ -461,104 +493,162 @@ function dirMoveOverwriteOnMacAPFS (sameInodeChange /*: ?LocalChange */, e /*: L
...
+    stats: moveChange.stats, // FIXME: e.stats?
```
taratatach: Why not switching to `e.stats` directly?

core/local/change.js
sebn: I thought exactly the same at first, but later discovered it actually wouldn't work.
So the current implementation is currently working *thanks to some weird behavior*.
But I believe it was worth refactoring since this is now easier to spot.

taratatach: Why wouldn't it work? Don't we replace the change in the list anyway?

core/local/analysis.js
taratatach: Arf, OK. This seems more and more useless since we have `this.changesByPath`.

core/local/change.js
taratatach: Seems fair.

core/local/analysis.js
sebn: But the difference is that, at the end, `changes` are concatenated to the `changesByInode` values (not sure why), while `changesByPath` are dropped. Which is the main point that makes me relunctant to merge my own code: either we use this `_.pull` assuming it ensure the behavior is the same (possibly with no test at all asserting this), or we dropped the `changes` stuff without even knowing if something breaks...

Maybe we should keep this open and add a subsequent PR to check whether the `changes` stuff removal actually breaks any test?

sebn: I pushed a new commit removing all the `changes` stuff. Let's see what happens on CI.

core/local/change.js
sebn: Another option may be to return a `ChangeReplacement` everywhere relevant, but stop removing the replaced change from `LocalChangeMap#changesByPath` with a comment.

sebn:
Marking back as `[WIP]` so we focus on the new watcher instead.

core/local/analysis.js
sebn: Looks like it works :wink:

sebn:
We'll get back to it later.

PR #1412, opened 2019-01-17
core/local: Fix dir move + subfile update

sebn:

- [x] PR is not too big

### DIFF
--- a/core/local/change.js
+++ b/core/local/change.js
@@ -31,6 +31,7 @@ module.exports = {
   maybeDeleteFolder,
   find,
   isChildMove,
+  isChildUpdate,
   addPath,
   delPath,
   childOf,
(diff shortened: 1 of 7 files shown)


core/local/change.js
```diff
@@ -478,6 +489,7 @@ function includeChangeEventIntoFileMove (sameInodeChange /*: ?LocalChange */, e
   log.debug({path: e.path}, 'FileMove + change')
   moveChange.md5sum = moveChange.old.md5sum || moveChange.md5sum
   moveChange.update = _.defaults({
+    md5sum: e.md5sum,
```
taratatach: Don't we already get `md5sum` from `e` via the call to `_.defaults`?

test/unit/local/analysis.js
```diff
@@ -601,6 +602,87 @@ describe('core/local/analysis', function () {
     })
   })
 
+  describe('DirMove(src/ → dst/) + FileUpdate(dst/file)', () => {
+    describe('unlinkDir(src/) + addDir (dst/) + unlink(src/file) + add(dst/file) + change(dst/file)', () => {
+      it('happens when client is running', () => {
```
taratatach: nitpick: the descriptions of the example and the describe around it should be reversed imho, giving something like this:
```javascript
context('when client is running', () => {
  it('translates to unlinkDir(src/) + addDir(dst/) + unlink(src/file) + add(dst/file) + change(dst/file)', () => {
    …
  })
})
```

And the second test would give:
```javascript
context('when client was stopped', () => {
  it('translates to unlinkDir(src/) + addDir(dst/) + unlink(src/file) + add(dst/file, new md5sum) with made up unlink* events', () => {
    …
  })
})

File: core/local/chokidar_watcher.js [hunk 3]
```diff
@@ -304,6 +304,11 @@ module.exports = class LocalWatcher {
     const errors /*: Error[] */ = []
     for (let c of changes) {
       try {
+        if (c.needRefetch) {
+          // $FlowFixMe
```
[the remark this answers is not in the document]
sebn: Because fixing it probably requires a lot of changes in the current type definitions.

core/local/change.js
sebn: Probably :man_facepalming:

test/unit/local/analysis.js
sebn: The descriptions match the whole file layout, introduced in https://github.com/cozy-labs/cozy-desktop/pull/1354

taratatach: OK, it makes sense.

core/local/chokidar_watcher.js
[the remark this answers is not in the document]
taratatach: OK :ok_hand:

core/local/change.js
sebn: Fixed

sebn:
> Should we add the scenarios for Windows and MacOS as part of this PR?

The macOS capture is similar to the GNU/Linux one.

On Windows, I couldn't get a capture where the final `change` was not swallowed.
But for some unknown reason it seems to [pass on AppVeyor](https://ci.appveyor.com/project/cozy/cozy-desktop/builds/21695903/job/v6vfhspb4pljfmjm#L184)... :confused:

sebn:
I believe it works because we still compute the md5sum and since it is different the file ends up being updated. I fixed up the commit [accordingly](https://github.com/cozy-labs/cozy-desktop/pull/1412/commits/22d820f5982c50552b2e55bd360fa6a08b415098).

sebn:
Travis is running the windows case (and failing) on Linux/macOS :man_facepalming:

PR #1427, opened 2019-01-31
Release v3.12.0-beta.3

taratatach:
Improvements for all users:

- We've modified the way we handle child documents updates so everything gets
  synchronised, even if a child update occurs quickly after its parent directory
  has been renamed.
- We found out that creating or moving a file or a directory to a path that has
  already been used in the path then deleted could lead to unrelevant version
  numbers for the new document and in the end to unnecessary conflicts.
  We now correct our version numbers when this situation happens so future
  updates to the document are handled correctly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,20 @@
 # Cozy Drive for Desktop: Changelog
 
+## 3.12.0-beta.3 - 2019-01-31
+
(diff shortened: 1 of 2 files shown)


taratatach:
> > creating or moving
> 
> Wasn't creation already fixed?

You're right. I'll change the wording.

PR #1431, opened 2019-02-04
test/support/builders/metadata: Match implementation

taratatach:
In the metadata module, when building a file or a directory, we use an
  optional `remote` parameter to fill in the `remote` attribute of the
  `Metadata` object being built. The way it is done, if the parameter is
  not passed, we end up with a `remote` attribute with an `undefined`
  value instead of not having the parameter at all.

  Also, when building a directory, we set the `sides` attribute to an
  empty object instead of not having an attribute at all like we do when
  building a file.

The metadata's `buildFile` method creates documents with a class and
  mime attributes which are missing from the metadata test builders thus
  preventing us from comparing them with documents from test builders.

- [x] PR is not too big

### DIFF
--- a/test/support/builders/metadata/base.js
+++ b/test/support/builders/metadata/base.js
@@ -69,7 +69,7 @@ module.exports = class BaseMetadataBuilder {
   }
 
   unmerged (sideName /*: SideName */) /*: this */ {
-    delete this.sides
+    if (this.doc.docType === 'file') delete this.doc.sides
     if (sideName === 'local') this.noRemote()
     return this.noRev()
   }
@@ -85,7 +85,11 @@ module.exports = class BaseMetadataBuilder {
   }
 
   noRemote () /*: this */ {
-    delete this.doc.remote
+    /*
+     * $FlowFixMe Flow is lying to us when allowing metadata's buildFile and
+     * buildDir to set remote to undefined
+     */
+    this.doc.remote = undefined
     return this
   }
 
--- a/test/unit/merge.js
+++ b/test/unit/merge.js
@@ -270,9 +270,11 @@ describe('Merge', function () {
         should(sideEffects).deepEqual({
           savedDocs: [{
             _id: initialFile._id,
+            class: 'application',
             docType: 'file',
             ino: initialFile.ino,
             md5sum: offUpdate.md5sum,
+            mime: 'application/octet-stream',
             path: initialFile.path,
             sides: {local: 2},
             size: offUpdate.size,
@@ -297,9 +299,11 @@ describe('Merge', function () {
           savedDocs: [
             {
               _id: initial._id,
+              class: 'application',
               docType: initial.docType,
               ino: initial.ino,
               md5sum: secondUpdate.md5sum,
+              mime: 'application/octet-stream',
               path: initial.path,
               remote: synced.remote,
               sides: {local: 4, remote: 2},
@@ -456,9 +460,11 @@ describe('Merge', function () {
         savedDocs: [
           {
             _id: initial._id,
+            class: 'application',
             docType: initial.docType,
             ino: initial.ino,
             md5sum: mergedLocalUpdate.md5sum,
+            mime: 'application/octet-stream',
             path: initial.path,
             sides: {local: 3},
             size: mergedLocalUpdate.size,
@@ -768,8 +774,10 @@ describe('Merge', function () {
         const src = {
           _deleted: true,
           _id: was._id,
+          class: 'application',
           docType: was.docType,
           md5sum: was.md5sum,
+          mime: 'application/octet-stream',
           moveTo: dstId,
           path: was.path,
           remote: was.remote,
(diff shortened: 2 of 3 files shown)


test/support/builders/metadata/base.js
```diff
@@ -69,7 +69,7 @@ module.exports = class BaseMetadataBuilder {
   }
 
   unmerged (sideName /*: SideName */) /*: this */ {
-    delete this.sides
+    if (this.doc.docType === 'file') delete this.doc.sides
```
sebn: Why not for a `folder`?

taratatach: To match `buildDir`'s implementation:
> Also, when building a directory, we set the sides attribute to an
empty object instead of not having an attribute at all like we do when
building a file.

If we remove the `sides` attribute entirely, we won't have something equivalent to a directory Metadata object that has been built but not merged yet.

test/unit/merge.js
```diff
@@ -270,9 +270,11 @@ describe('Merge', function () {
         should(sideEffects).deepEqual({
           savedDocs: [{
             _id: initialFile._id,
+            class: 'application',
```
sebn: Do we really want to hardcode mime/class default values everywhere?

Why not use `initialFile.class` here?

test/unit/merge.js
```diff
@@ -270,9 +270,11 @@ describe('Merge', function () {
...
             docType: 'file',
             ino: initialFile.ino,
             md5sum: offUpdate.md5sum,
+            mime: 'application/octet-stream',
```
sebn: `initialFile.mime`?

test/unit/merge.js
```diff
@@ -297,9 +299,11 @@ describe('Merge', function () {
           savedDocs: [
             {
               _id: initial._id,
+              class: 'application',
```
sebn: `initial.class`?

test/unit/merge.js
```diff
@@ -297,9 +299,11 @@ describe('Merge', function () {
...
               docType: initial.docType,
               ino: initial.ino,
               md5sum: secondUpdate.md5sum,
+              mime: 'application/octet-stream',
```
sebn: `initial.mime`?

test/unit/merge.js
```diff
@@ -456,9 +460,11 @@ describe('Merge', function () {
         savedDocs: [
           {
             _id: initial._id,
+            class: 'application',
```
sebn: `initial.class`?

test/unit/merge.js
```diff
@@ -456,9 +460,11 @@ describe('Merge', function () {
...
             docType: initial.docType,
             ino: initial.ino,
             md5sum: mergedLocalUpdate.md5sum,
+            mime: 'application/octet-stream',
```
sebn: `initial.mime`?

test/unit/merge.js
```diff
@@ -768,8 +774,10 @@ describe('Merge', function () {
         const src = {
           _deleted: true,
           _id: was._id,
+          class: 'application',
```
sebn: `was.class`?

test/unit/merge.js
```diff
@@ -768,8 +774,10 @@ describe('Merge', function () {
...
           docType: was.docType,
           md5sum: was.md5sum,
+          mime: 'application/octet-stream',
```
sebn: `was.mime`?

test/unit/merge.js
```diff
@@ -780,8 +788,10 @@ describe('Merge', function () {
         }
         const dst = {
           _id: dstId,
+          class: 'application',
```
sebn: `doc.class`?

test/unit/merge.js
```diff
@@ -780,8 +788,10 @@ describe('Merge', function () {
...
           docType: doc.docType,
           md5sum: doc.md5sum,
+          mime: 'application/octet-stream',
```
sebn: `doc.mime`?

test/unit/merge.js
```diff
@@ -956,8 +966,10 @@ describe('Merge', function () {
         const src = {
           _deleted: true,
           _id: baz._id,
+          class: 'application',
```
sebn: `baz.class`?

test/unit/merge.js
```diff
@@ -956,8 +966,10 @@ describe('Merge', function () {
...
           docType: baz.docType,
           md5sum: baz.md5sum,
+          mime: 'application/octet-stream',
```
sebn: `baz.mime`?

test/unit/merge.js
```diff
@@ -968,8 +980,10 @@ describe('Merge', function () {
         }
         const dst = {
           _id: qux._id,
+          class: 'application',
```
sebn: etc...

test/unit/merge.js
```diff
@@ -968,8 +980,10 @@ describe('Merge', function () {
...
           docType: qux.docType,
           md5sum: baz.md5sum,
+          mime: 'application/octet-stream',
```
sebn: etc...

test/unit/merge.js
sebn: Do we really want to hardcode the default class/mime values everywhere?

sebn: 👌

PR #1433, opened 2019-02-04
core/local/watcher: Use Atom watcher by default

taratatach:
For testing purposes we enable the new Atom watcher by default again
  on Linux and Windows.


taratatach:
Closing in favor of https://github.com/cozy-labs/cozy-desktop/pull/1434

PR #1434, opened 2019-02-04
core/local/watcher: Get watcher type from Config

taratatach:
To simplify testing the new local watcher using Atom Watcher behind
  the scenes, we move the choice of watcher to the Config.
  If a `watcherType` property is defined in `config.json`, we use it,
  otherwise we set with the 2 mechanisms that already existed (i.e. user
  defined watcher via env variable then platform default watcher).

- [x] PR is not too big


test/unit/config.js
```diff
@@ -204,6 +205,50 @@ describe('Config', function () {
     })
   })
 
+  describe('WatcherType', function () {
```
sebn: lowercase `w`?

PR #1436, opened 2019-02-05
[WIP] test/unit/local/steps/buffer: Introduce tests

sebn:

- [x] PR is not too big


sebn:
e209ac37fb598c5220fa9bc5433951c10ced1342 tests more case with less code and makes it easier to see tested scenarios, so I would tend to keep it. WDYT?

test/unit/local/steps/buffer.js
```diff
@@ -0,0 +1,372 @@
...
+    describe('#push()', () => {
+      it('inserts the given Batch into the Buffer so it may be retrieved with #pop()', async () => {
+        const batch = nonEmptyBatch()
+
+        buffer.push(batch)
+
+        await should(buffer.pop()).be.fulfilledWith(batch)
+      })
+
+      it('inserts the same batch multiple times', async () => {
```
taratatach: The way I understand this description is a call to `push()` will insert the given batch multiple times.
What do you think of "does not squash equivalent batches" ?

test/unit/local/steps/buffer.js
```diff
@@ -0,0 +1,372 @@
...
+        const batch = nonEmptyBatch()
+
+        buffer.push(batch)
+        buffer.push(batch)
+
+        await should(buffer.pop()).be.fulfilledWith(batch)
+        await should(buffer.pop()).be.fulfilledWith(batch)
+      })
+
+      it('inserts an empty batch', async () => {
```
taratatach: "accepts empty batches" ?

test/unit/local/steps/buffer.js
```diff
@@ -0,0 +1,372 @@
...
+        await should(buffer.pop()).be.fulfilledWith(batch2)
+      })
+
+      describe('when all Batches were already popped', () => {
+        beforeEach('push and pop', async () => {
+          buffer.push(nonEmptyBatch(1))
+          await buffer.pop()
+        })
+
+        it('awaits for the next Batch', async () => {
```
taratatach: Nicely done!

test/unit/local/steps/buffer.js
```diff
@@ -0,0 +1,372 @@
...
+          const nextBatchPromise = buffer.pop()
+          buffer.push(batch1)
+          buffer.push(batch2)
+
+          await should(nextBatchPromise).be.fulfilledWith(batch1)
+        })
+      })
+
+      describe('when called more than once on an empty Buffer (DO NOT DO THIS)', () => {
+        it('only resolves the last Promise with the next Batch', async () => {
```
taratatach: Isn't this a bug?

test/unit/local/steps/buffer.js
```diff
@@ -0,0 +1,372 @@
...
+        'map pop push push pop',
+        'push map pop',
+        'push map pop push pop',
+        'push map push pop pop',
+        'push push map pop pop'
+      ]
+
+      /** MOCHA SETUP */
+
+      for (let scenario of scenarios) {
```
taratatach: Use `const` instead of `let`?

test/unit/local/steps/buffer.js
```diff
@@ -0,0 +1,372 @@
...
+          beforeEach('init scenarioState', () => {
+            scenarioState = {
+              callback: sinon.spy(transform),
+              inputBatches: [],
+              inputBuffer: new Buffer(),
+              outputBatchesPromise: Promise.resolve([])
+            }
+          })
+
+          for (let step of scenarioSteps(scenario)) {
```
taratatach: Use `const` instead of `let`?

test/unit/local/steps/buffer.js
```diff
@@ -0,0 +1,372 @@
...
+          beforeEach('init scenarioState', () => {
+            scenarioState = {
+              callback: sinon.spy(asyncTransform),
+              inputBatches: [],
+              inputBuffer: new Buffer(),
+              outputBatchesPromise: Promise.resolve([])
+            }
+          })
+
+          for (let step of scenarioSteps(scenario)) {
```
taratatach: Use `const` instead of `let`?

test/unit/local/steps/buffer.js
```diff
@@ -0,0 +1,372 @@
...
+        'asyncMap pop push push pop',
+        'push asyncMap pop',
+        'push asyncMap pop push pop',
+        'push asyncMap push pop pop',
+        'push push asyncMap pop pop'
+      ]
+
+      /** MOCHA SETUP */
+
+      for (let scenario of scenarios) {
```
taratatach: Use `const` instead of `let`?

test/unit/local/steps/buffer.js
```diff
@@ -0,0 +1,372 @@
...
+
+          for (let step of scenarioSteps(scenario)) {
+            beforeEach(step, () => {
+              // Make sure we don't return a Promise to mocha so scenarios may
+              // reproduce async edge cases.
+              scenarioStepFn(step)(scenarioState)
+            })
+          }
+
+          beforeEach('await output', () => scenarioState.outputBatchesPromise)
```
taratatach: I don't quite understand what we're waiting here and why.

sebn: Although `Buffer#map()` applies a synchronous transform, `#pop()` is still async.
And the `pop` steps are non-blocking, so we may test non-waiting push/pop combos.

So at the end, unless we await for all the `#pop()` to complete (using this promise), all the callback assertions will fail.
The new output Buffer assertion doesn't care and the output batches one will work since it's explicitely awaiting for the same promise to get the batches.

It seemed logical to me to await before assertions, otherwise the scenario run is not yet complete in some way...

test/unit/local/steps/buffer.js
sebn: Well, I wonder :)
Ideally we should either prevent it in the implementation, or support it.

sebn:
Fixed and rebased according to feedback.

@taratatach WDYT?

test/unit/local/steps/buffer.js
sebn: Also this may not be our most important issue to fix since we don't have the need to do this for now.

test/unit/local/steps/buffer.js
sebn: Would an improved hook description help? E.g. `wait for output batches to be available`?

taratatach: No it's alright. I just missed the fact that all the actions are run in a `beforeEach` callback prior to this one.

I get it now :).

PR #1437, opened 2019-02-05
[WIP] test/unit/merge: Use builders for expected values

taratatach:
It can get pretty hard to read and maintain expectations when checking
  the results of side effects.
  Using test metadata builders to create the expectations objects can
  help with that.

- [x] PR is not too big

### DIFF
--- a/test/support/builders/metadata/base.js
+++ b/test/support/builders/metadata/base.js
@@ -61,6 +61,12 @@ module.exports = class BaseMetadataBuilder {
     return this
   }
 
+  moveTo (id /*: string */) /*: this */ {
+    this.doc.moveTo = id
+    this.doc._deleted = true
+    return this
+  }
+
   /** Make sure the doc is not the same as before. */
   whateverChange () /*: this */ {
     this.doc.tags = this.doc.tags || []
@@ -156,6 +162,11 @@ module.exports = class BaseMetadataBuilder {
       _id,
       _rev: dbBuilders.rev()
     }
+    return this
+  }
+
+  remote (doc /*: RemoteDoc */) /*: this */ {
+    this.doc.remote = doc
     return this
   }
 
--- a/test/unit/merge.js
+++ b/test/unit/merge.js
@@ -267,20 +267,17 @@ describe('Merge', function () {
         this.merge.addFileAsync('local', _.cloneDeep(offUpdate))
         )
 
+        const doc = builders
+          .metafile(initialFile)
+          .data('off update')
+          .updatedAt(offUpdate.updated_at)
+          .sides({ local: 2 })
+          .build()
+        delete doc._rev // Side effects don't include doc's _rev
+        delete doc.remote // FIXME: stop mixing no remote attribute and undefined remote attribute
+
         should(sideEffects).deepEqual({
-          savedDocs: [{
-            _id: initialFile._id,
-            class: 'application',
-            docType: 'file',
-            ino: initialFile.ino,
-            md5sum: offUpdate.md5sum,
-            mime: 'application/octet-stream',
-            path: initialFile.path,
-            sides: {local: 2},
-            size: offUpdate.size,
-            tags: initialFile.tags, // tags can't be updated on the local side
-            updated_at: offUpdate.updated_at
-          }],
+          savedDocs: [doc],
           resolvedConflicts: []
         })
       })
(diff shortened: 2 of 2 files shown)


test/support/builders/metadata/base.js
```diff
@@ -159,6 +165,11 @@ module.exports = class BaseMetadataBuilder {
     return this
   }
 
+  remote (doc /*: RemoteDoc */) /*: this */ {
```
sebn: Should the type be `MetadataRemoteInfo`, not `RemoteDoc`?
Also what about renaming `doc` to `remoteInfos`? (`remote` would sound a bit too much like the `Remote` side to me)

test/unit/merge.js
```diff
@@ -963,37 +912,8 @@ describe('Merge', function () {
...
-            updated_at: baz.updated_at // FIXME: Stop mixing dates & strings
-          }, src),
-          path: qux.path,
-          remote: qux.remote,
-          sides: {local: 1, remote: 2},
-          size: qux.size,
-          tags: [],
-          updated_at: qux.updated_at
-        }
+        const src = builders.metafile(baz).sides({ local: 1, remote: 2 }).moveTo(qux.path).build()
```
sebn: Should be `moveTo(qux._id)`. Or even better make `moveTo()` take a path.

test/support/builders/metadata/base.js
```diff
@@ -61,6 +61,12 @@ module.exports = class BaseMetadataBuilder {
     return this
   }
 
+  moveTo (id /*: string */) /*: this */ {
```
sebn: Wouldn't a `path` be more intuitive for the caller?

I would think so since even the author of this new builder method mistakenly used `.moveTo(qux.path)` in [`test/unit/merge.js:915`](https://github.com/cozy-labs/cozy-desktop/pull/1437/files#diff-d5f621b4689d0b6f0b95586c98caee11R915) :wink: 

Also, since `BaseMetadataBuilder#moveFrom()` takes a full doc, not a path nor an id, won't this new `#moveTo` method be a bit confusing? Although I'm not sure what would be the best solution... `moveToPath()`?

test/support/builders/metadata/base.js
```diff
@@ -61,6 +61,12 @@ module.exports = class BaseMetadataBuilder {
...
+    this.doc.moveTo = id
+    this.doc._deleted = true
```
sebn: Aren't we duplicating part of `core/move`'s logic with those two lines?

test/unit/merge.js
```diff
@@ -267,20 +267,17 @@ describe('Merge', function () {
         this.merge.addFileAsync('local', _.cloneDeep(offUpdate))
         )
 
+        const doc = builders
```
sebn: Since `doc` is the de-facto name in almost all `Merge` methods, but this `doc` isn't actually the one passed to `Merge`, won't this be a bit confusing?

What about `expectedDoc`? Or `mergedDoc`? Or maybe inline it in the expectation?

test/unit/merge.js
```diff
@@ -267,20 +267,17 @@ describe('Merge', function () {
...
+          .metafile(initialFile)
+          .data('off update')
+          .updatedAt(offUpdate.updated_at)
+          .sides({ local: 2 })
+          .build()
+        delete doc._rev // Side effects don't include doc's _rev
```
sebn: Use `BaseMetadatabuilder#noRev()` instead?

test/unit/merge.js
```diff
@@ -267,20 +267,17 @@ describe('Merge', function () {
...
+        delete doc.remote // FIXME: stop mixing no remote attribute and undefined remote attribute
```
sebn: I suppose we can't use `BaseMetadataBuilders#noRemote()` for this, since it sets `doc.remote` to `undefined`... :sob:

test/unit/merge.js
```diff
@@ -449,30 +440,18 @@ describe('Merge', function () {
...
       const synced = await builders.metafile(initial).sides({local: 2, remote: 2}).create()
-      const mergedLocalUpdate = await builders.metafile(synced).sides({local: 3, remote: 2}).data('local update').create()
+      await builders.metafile(synced).sides({local: 3, remote: 2}).data('local update').create()
       const newRemoteUpdate = builders.metafile(synced).unmerged('remote').data('remote update').build()
 
       const sideEffects = await mergeSideEffects(this, () =>
         this.merge.updateFileAsync('remote', newRemoteUpdate)
       )
 
+      const doc = builders.metafile(initial).data('local update').sides({ local: 3 }).build()
```
sebn: Duplicates the `'local update'` string?

sebn:
I would favor fixing the hardcoded class/mime props and focus on the new watcher.

test/support/builders/metadata/base.js
sebn: Other solutions are not so nice, better duplicate until we find a better way (still better than duplicating in many tests).

test/unit/merge.js
sebn: As discussed, maybe build `doc` from `mergedLocalUpdate`?

taratatach:
Closed in favor of #1439

PR #1440, opened 2019-02-12
core/merge: Reset sides on every move

taratatach:
When moving a file or directory, we delete the source document and
  create a new one in Pouch.
  This new document should only have one side, the one from which the
  document was created, and its version should be 1 since this is a new
  document (n.b. it can end up being higher if the path is being reused
  and we're fixing the sides later).

  We were doing this reset in some situations only. Now, we do it as
  part of the `move` call so it's done in every move.

- [x] PR is not too big

### DIFF
--- a/core/merge.js
+++ b/core/merge.js
@@ -258,7 +258,7 @@ class Merge {
       if (doc.ino == null) { doc.ino = was.ino }
       if (doc.fileid == null) { doc.fileid = was.fileid }
       if (doc.remote == null) { doc.remote = was.remote }
-      move(was, doc)
+      move(side, was, doc)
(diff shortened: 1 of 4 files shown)


taratatach:
It works because we completely remove `sides` in `move()` now so previous calls to `markSide` don't matter.
We should take look at it though and see if we can clean that up.

sebn:
Yeah, in the end I was confident that the behavior was ok and the current PR was definitely an improvement, but maybe I merged a bit quickly, my bad.

sebn:
I'll fix my mess in https://github.com/cozy-labs/cozy-desktop/pull/1441

PR #1442, opened 2019-02-12
test/support/builders/atom_watcher_event: No stats

taratatach:
Passing `fs.Stats` objects to build or modify a test object can be
  cumbersome as it requires a lot of attributes to be set to match the
  objects that will really be built by the production code.

  Most of the time, when building test AtomWatcherEvents, we only want
  to set the `ino` property so we remove the `stats` method and add a
  new `ino` method on the test builder. The `stats` property of the
  event is managed by the builder.

- [x] PR is not too big

### DIFF
--- a/test/support/builders/atom_watcher_event.js
+++ b/test/support/builders/atom_watcher_event.js
@@ -12,10 +12,33 @@ import type { Stats } from 'fs'
 import type { AtomWatcherEvent, EventAction, EventKind, Batch } from '../../../core/local/steps/event'
 */
 
+const NON_EXECUTABLE_MASK = 0 << 6
+
 function randomPick /*:: <T> */ (elements /*: Array<T> */) /*: T */{
   const l = elements.length
   const i = Math.floor(Math.random() * l)
   return elements[i]
+}
+
+function buildStats (kind /*: EventKind */) /*: fs.Stats */ {
+  let baseStats /*: fs.Stats */
+  if (kind === 'file') {
+    baseStats = fs.statSync(__filename)
+  } else {
+    baseStats = fs.statSync(__dirname)
+  }
+
+  return _.defaults(
+    {
+      atime: new Date(),
+      mtime: new Date(),
+      ctime: new Date(),
+      birthtime: new Date(),
+      mode: baseStats.mode & NON_EXECUTABLE_MASK,
+      size: 0
+    },
+    _.omit(baseStats, ['executable'])
+  )
 }
 
 module.exports = class AtomWatcherEventBuilder {
@@ -27,11 +50,14 @@ module.exports = class AtomWatcherEventBuilder {
     if (old) {
       this._event = _.cloneDeep(old)
     } else {
+      const kind = randomPick(events.KINDS)
+      const stats = buildStats(kind)
       this._event = {
         action: randomPick(events.ACTIONS),
-        kind: randomPick(events.KINDS),
+        kind,
         path: '/',
-        _id: '/'
+        _id: '/',
+        stats
       }
     }
   }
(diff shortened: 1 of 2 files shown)


test/unit/local/steps/initial_diff.js
```diff
@@ -92,8 +92,8 @@ describe('local/steps/initialDiff', () => {
 
     const events = await buffer.pop()
     should(events).deepEqual([
-      builders.event().action('deleted').kind('file').path('bar').build(),
-      builders.event().action('deleted').kind('directory').path('foo').build(),
+      builders.event().action('deleted').kind('file').path('bar').noStats().build(),
+      builders.event().action('deleted').kind('directory').path('foo').noStats().build(),
```
sebn: Shouldn't `deleted` events always have no stats?

test/support/builders/atom_watcher_event.js
```diff
@@ -66,11 +75,14 @@ module.exports = class AtomWatcherEventBuilder {
     return this
   }
 
-  stats (newStats /*: { ino?: number } */) /*: this */ {
-    const stats /*: Stats */ = fs.statSync(__filename)
-    Object.assign(stats, newStats)
+  ino (newIno /*: number */) /*: this */ {
```
sebn: :+1:

sebn:
Also, regarding realistic stats, maybe we could later use ``fs.statSync(__dirname)` depending on the event `kind`.

test/support/builders/atom_watcher_event.js
```diff
@@ -27,11 +27,20 @@ module.exports = class AtomWatcherEventBuilder {
     if (old) {
       this._event = _.cloneDeep(old)
     } else {
+      const stats = _.merge(new fs.Stats(), {
+        ino: 1,
```
sebn: I'm not sure to remember how inodes are used in the new watcher...
Should we ensure 2 generated events don't end up with the same inode by default?
(maybe it doesn't matter)

taratatach: Hum this could be a good practice.

sebn: (not a blocker for this PR since previous implementation was always using the same inode too)

test/support/builders/atom_watcher_event.js
```diff
@@ -27,11 +27,20 @@ module.exports = class AtomWatcherEventBuilder {
...
+        atime: new Date(),
+        mtime: new Date(),
+        ctime: new Date(),
+        birthtime: new Date(),
+        size: 0
+      })
```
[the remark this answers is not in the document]
taratatach: I remember trying to use it at some point and having issues but I can't remember why. I'll try again.

test/unit/local/steps/initial_diff.js
taratatach: Maybe?? I can't really tell. But this seems plausible.

test/support/builders/atom_watcher_event.js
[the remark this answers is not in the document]
taratatach: I remember trying to use it at some point and having issues but I can't remember why. I'll try again.

test/unit/local/steps/initial_diff.js
sebn: Please note I actually don't know the answer :smile:

nono: Look at https://raw.githubusercontent.com/cozy-labs/cozy-desktop/master/doc/developer/win_watcher.png & https://raw.githubusercontent.com/cozy-labs/cozy-desktop/master/doc/developer/linux_watcher.png : there are no stats on the deleted lines.

PR #1444, opened 2019-02-12
core/merge: Fix side after remote dissociation

taratatach:
When trying to merge a remote file update which results in a conflict
  resolution, we dissociate the untouched local document from the remote
  version.
  When doing so, we're actually saving a new version of the document in
  Pouch which will result in a `_rev` increase and should therefore
  result in a local side increase as well.

- [x] PR is not too big


sebn:
Waiting for AppVeyor scenarios job.

PR #1450, opened 2019-02-16
test/scenarios: AtomWatcher support

sebn:
Needs #1448 and #1449

- [x] PR is not too big


test/support/helpers/local.js
```diff
@@ -99,6 +110,25 @@ class LocalTestHelpers {
     return this.side.watcher.onFlush(events)
   }
 
+  dispatchAtomEvents (batch /*: Batch */) {
+    const { _resolveMerge } = this
+    if (_resolveMerge && _.isEqual(batch, mergeDoneBatch)) {
+      _resolveMerge()
+    } else {
```
taratatach: We need a `return []` here.

PR #1456, opened 2019-02-19
test/unit/scheduling: Enable flow & fix minor issues 

sebn:
- `Promise.delay()` requires bluebird
- `Sync#start()` takes a mode (although `undefined` happens to work the same as `full`)

---

- [x] PR is not too big


sebn:
To be squashed & merged after test scenarios.

PR #1465, opened 2019-02-22
[CHOKIDAR] core/local: Introduce chokidar step modules

sebn:
First step to make chokidar steps easier to understand/refactor.
This mostly implies moving/extracting stuff to separate files & injecting dependencies that used to be available directly as `ChokidarWatcher` properties.

---

- [x] PR is not too big

### DIFF
--- a/core/local/chokidar_steps/initial_diff.js
+++ b/core/local/chokidar_steps/initial_diff.js
@@ -1,0 +1,89 @@
+/* @flow */
+
+const chokidarEvent = require('../chokidar_event')
+const logger = require('../../logger')
+const metadata = require('../../metadata')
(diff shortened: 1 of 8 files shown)


sebn:
We should probably revert the last (*analysis*) commit to prevent conflicts with #1408.

sebn:
Many other things can be improved later:

- Rename `InitialScan` to `Scan`.
- Bring back reverted `analysis` module step
- Extract more steps from `analysis` (although it may require changing unit tests since analysis is currently tested as a whole)

But I believe the current PR is good to merge.

sebn:
We'll get back to it later.

PR #1467, opened 2019-02-25
core/merge: Compare remotes only when not null

taratatach:
When merging an update or the creation of a directory at a path where
  a directory already exist, we check if their remote ids are equal to
  determine if we should update the existing document or resolve a
  conflict.
  However, in the situation where the both the existing document and new
  document are local only, we don't have a `remote` attribute and trying
  to compare their ids will throw an error for trying to access an
  attribute on `undefined`.

  We just make sure that a `remote` attribute exists on the new
  document before comparing the ids because the document's `remote`
  attribute will be assigned the existing's document's `remote`
  attribute if it exists.

In order to test this, we needed to have coherent `remote` attribute values in test documents.

Indeed, local documents which have not been synchronised should not have a
  `remote` attribute. On the other hand, remote documents and
  synchronised documents should always have one.

  We make sure the test document builders don't assign a `remote` value
  by default and that tests assign one when necessary.

- [x] PR is not too big

### DIFF
--- a/test/unit/remote/index.js
+++ b/test/unit/remote/index.js
@@ -217,9 +217,8 @@ describe('remote.Remote', function () {
         }
       }
       await this.remote.addFileAsync(doc)
-      should.exist(doc.remote._id)
-      should.exist(doc.remote._rev)
-      should.exist(doc._deleted)
+      should(doc._deleted).be.true()
+      should(doc).not.have.property('remote')
     })
   })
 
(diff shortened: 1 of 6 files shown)


test/unit/remote/index.js
```diff
@@ -217,9 +217,8 @@ describe('remote.Remote', function () {
         }
       }
       await this.remote.addFileAsync(doc)
-      should.exist(doc.remote._id)
-      should.exist(doc.remote._rev)
-      should.exist(doc._deleted)
+      should(doc._deleted).be.true()
+      should(doc).not.have.property('remote')
```
sebn: Oh my... File could not be uploaded so `doc` shouldn't have a `remote` afterwards.
:+1: good catch.

sebn:
Closing since the problematic code was reverted in #1482

PR #1472, opened 2019-02-25
core/local/steps/buffer: Drop empty batches

sebn:
No need to pass them around and run useless code.

- [x] PR is not too big


sebn:
- [ ] Rebase on master

PR #1473, opened 2019-02-25
core/config: Default to atom watcher 

sebn:

- [x] PR is not too big


sebn:
- [ ] Rebase on #1474 and make sure the `'renamed'` event kind check isn't required to make tests pass

sebn:
@taratatach We should make sure scenarios don't trigger errors event when passing before merging.

PR #1475, opened 2019-02-26
test/support/helpers: Load default ignore rules

sebn:
Otherwise tests will fail with AtomWatcher since it doesn't have
an `ignored` option like chokidar.

This doesn't prevent us to keep using the chokidar option.

Courtesy of @taratatach.

**Needs #1476**

---

- [x] PR is not too big


sebn:
- [ ] Fix infinite loop on `Ignore#match()` in `test/integration/platform_incompatibilities.js` (introduced by previous commit, possibly a bug in production)

PR #1477, opened 2019-02-28
core/local/atom_watcher: Initialise steps' states

taratatach:
Before starting the producer and looping over events, we initialize
  the state of the steps that need one so this can be done whenever we
  restart the watcher.

  This is useful for the initialDiff step for example since it needs to
  build a map of the existing documents to detect deletions and moves
  that can have occurred while the watcher was stopped.

- [x] PR is not too big

### DIFF
--- a/core/local/atom_watcher.js
+++ b/core/local/atom_watcher.js
@@ -74,12 +74,23 @@ const producer = opts => {
   }
 }
 
+const stepsInitialState = opts =>
+  Promise.reduce(
+    async (state, step) =>
+      step.initialState
+        ? Object.assign(state, await step.initialState(opts))
+        : state,
+    steps,
+    {}
+  )
+
 module.exports = class AtomWatcher {
   /*::
   pouch: Pouch
   events: EventEmitter
   checksumer: Checksumer
   producer: Producer
+  state: Object
   running: Promise<void>
   _runningResolve: ?Function
   _runningReject: ?Function
@@ -90,11 +101,13 @@ module.exports = class AtomWatcher {
     this.events = opts.events
     this.checksumer = checksumer.init()
     this.producer = producer(opts)
+    this.state = {}
 
     const stepOptions = Object.assign(
       ({
         checksumer: this.checksumer,
-        scan: this.producer.scan
+        scan: this.producer.scan,
+        state: this.state
       } /*: Object */),
       opts
     )
@@ -106,13 +119,14 @@ module.exports = class AtomWatcher {
     dispatch.loop(buffer, stepOptions)
   }
 
-  start () {
+  async start () {
     log.debug('starting...')
     this.running = new Promise((resolve, reject) => {
       this._runningResolve = resolve
       this._runningReject = reject
     })
     this.events.emit('local-start')
+    this.state = await stepsInitialState(this)
     this.producer.start()
     const scanDone = new Promise((resolve) => {
       this.events.on('initial-scan-done', resolve)
--- a/core/local/steps/initial_diff.js
+++ b/core/local/steps/initial_diff.js
@@ -8,6 +8,11 @@ import type Pouch from '../../pouch'
 import type { AtomWatcherEvent, Batch, EventKind } from './event'
 import type { Metadata } from '../../metadata'
 
+type InitialDiffState = {
+  waiting: WaitingItem[],
+  byInode: Map<number|string, WatchedPath>,
+}
+
 type WatchedPath = {
   path: string,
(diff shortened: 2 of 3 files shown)


core/local/atom_watcher.js
```diff
@@ -106,13 +109,16 @@ module.exports = class AtomWatcher {
...
+  async start () {
     log.debug('starting...')
     this.running = new Promise((resolve, reject) => {
       this._runningResolve = resolve
       this._runningReject = reject
     })
     this.events.emit('local-start')
+    await Promise.all(steps.map(async step => {
+      if (step.initialState) _.assign(this.state, await step.initialState(this))
+    }))
```
sebn: Maybe extract function? E.g.:

```js
const stepsInitialState = opts =>
  Promise.reduce(
    async (state, step) =>
      step.initialState
        ? Object.assign(state, await step.initialState(opts))
        : state
    steps,
    {}
  )
```

Then in `AtomWatcher` constructor:

```js
this.state = await stepsInitialState(this)
```

core/local/atom_watcher.js
```diff
@@ -80,6 +80,7 @@ module.exports = class AtomWatcher {
   events: EventEmitter
   checksumer: Checksumer
   producer: Producer
+  state: Object
```
sebn: `state` looks good from the steps point of view, but `stepsState` would look better from the watcher point of view...

core/local/steps/initial_diff.js
```diff
@@ -25,35 +29,29 @@ type WaitingItem = {
...
-async function initialDiff (buffer /*: Buffer */, out /*: Buffer */, pouch /*: Pouch */) /*: Promise<void> */ {
-  const waiting /*: WaitingItem[] */ = []
-
-  // XXX we wait to receive the first batch of events before initializing this
-  // component, as pouchdb may not be initialized when initialDiff is created
-  // (its views are added later, but before the local watcher is started, thus
-  // before the first batch of events)
-  let events = await buffer.pop()
-
+async function initialState (opts /*: { pouch: Pouch } */) /*: Promise<{ [typeof STEP_NAME]: InitialDiffState }> */ {
```
sebn: Flow ninja!

core/local/steps/initial_diff.js
```diff
@@ -62,7 +60,18 @@ async function initialDiff (buffer /*: Buffer */, out /*: Buffer */, pouch /*: P
     }
   }
 
+  return {
+    [STEP_NAME]: { byInode }
+  }
+}
+
+async function initialDiff (buffer /*: Buffer */, out /*: Buffer */, pouch /*: Pouch */, state /*: Object */) /*: Promise<void> */ {
+  const waiting /*: WaitingItem[] */ = []
```
sebn: It would really make sense to include it in the step state imho.

test/unit/local/steps/initial_diff.js
```diff
@@ -26,10 +26,12 @@ describe('local/steps/initial_diff.loop()', () => {
     await builders.metadir().path('foo').ino(1).create()
     await builders.metafile().path('fizz').ino(2).create()
 
+    const state = await initialDiff.initialState({ pouch: this.pouch })
```
sebn: DRY? (`beforeEach` hook)

sebn:
I'm wondering... Could the step still use references on some part or the whole state even after *resetting* it?

test/unit/local/steps/initial_diff.js
taratatach: It needs to happen after we've created the metatada documents. So unless we can move those create in a beforeEach hook (which I believe we can't) then we need to have that call in the test itself.

core/local/atom_watcher.js
taratatach: The step initialisation needs to happen in `start()` otherwise we won't reset the state after restarting the watcher. That's the whole point of this PR ;-).

core/local/atom_watcher.js
taratatach: Having a different name complexifies the different function calls. So either we use `stepsState` everywhere or we stick to `state`.
I'd rather do the later but I guess it's a matter of taste.

core/local/steps/initial_diff.js
sebn: (can be improved later)

core/local/atom_watcher.js
sebn: `s/constructor/start` (my bad :stuck_out_tongue:)

PR #1479, opened 2019-02-28
core/local/steps/initial_diff: Detect path reuse

taratatach:
Generate a `deleted` event for a document whose associated file is
  missing only if there is no other document at the same path.
  Otherwise we won't create a document for the new file.

- [x] PR is not too big

### DIFF
--- a/test/unit/local/steps/initial_diff.js
+++ b/test/unit/local/steps/initial_diff.js
@@ -9,100 +9,106 @@ const pouchHelpers = require('../../../support/helpers/pouch')
 const Buffer = require('../../../../core/local/steps/buffer')
 const initialDiff = require('../../../../core/local/steps/initial_diff')
 
-describe('local/steps/initial_diff.loop()', () => {
+describe('local/steps/initial_diff', () => {
   let builders
-  let buffer
 
   before('instanciate config', configHelpers.createConfig)
   beforeEach('instanciate pouch', pouchHelpers.createDatabase)
   beforeEach('populate pouch with documents', function () {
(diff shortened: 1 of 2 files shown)


sebn:
As discussed offline, let's just add unit tests and fix the other issues in separate PR.

test/unit/local/steps/initial_diff.js
```diff
@@ -93,6 +112,24 @@ describe('local/steps/initial_diff', () => {
...
+
+      const state = await initialDiff.initialState({ pouch: this.pouch })
+
+      const foo = builders.event().action('scan').kind('directory').path('foo').ino(3).build()
+      const bar = builders.event().action('scan').kind('file').path('bar').ino(4).build()
+      buffer.push([foo, bar])
+      buffer = initialDiff.loop(buffer, { pouch: this.pouch, state })
+
+      const events = await buffer.pop()
+      should(events).deepEqual([
```
sebn: Unit test doesn't fail without the fix.

test/unit/local/steps/initial_diff.js
```diff
@@ -35,13 +37,14 @@ describe('local/steps/initial_diff', () => {
 
       const bar = builders.event().action('scan').kind('directory').path('bar').ino(1).build()
       const buzz = builders.event().action('scan').kind('file').path('buzz').ino(2).build()
-      buffer.push([bar, buzz])
+      buffer.push([bar, buzz, initialScanDone])
```
sebn: :+1: Good catch!

PR #1483, opened 2019-03-01
core/local: Improve logging in atom watcher & steps

sebn:
- Prefix loggers with `atom/`
- Ensure no error is swallowed
- Handle .event|.path in jq's path() filter
- Fix *experimental* wording
- Log expected/ignorable errors at the debug level
- Log non-critical but unexpected ones at the warn level
- Make sure event and other useful data are logged when available
- At least log path whenever possible if no event is available
- Use logger instead of console.log()
- Remove useless $FlowFixMe notice
- Don't return log.error() result in callback

- [x] PR is not too big

### DIFF
--- a/.jq
+++ b/.jq
@@ -103,8 +103,8 @@ def issues: clean | select(is_issue);
 def path(pattern):
   clean | select(
     (
-      (.path // (.doc | .path) // (.change | .doc | .path) // (.idConflict | .newDoc | .path)),
-      (.oldpath // (.was | .path) // (.idConflict | .existingDoc | .path)),
+      (.path // (.doc | .path) // (.change | .doc | .path) // (.idConflict | .newDoc | .path) // (.event | .path)),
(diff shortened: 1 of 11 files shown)


.jq
```diff
@@ -103,7 +103,7 @@ def issues: clean | select(is_issue);
 def path(pattern):
   clean | select(
     (
-      (.path // (.doc | .path) // (.change | .doc | .path) // (.idConflict | .newDoc | .path)),
+      (.path // (.doc | .path) // (.change | .doc | .path) // (.idConflict | .newDoc | .path) // (.event | .path)),
       (.oldpath // (.was | .path) // (.idConflict | .existingDoc | .path)),
```
sebn: Oops, missing `(.event | .oldPath)`

sebn:
Rebased on master & handled `.event|.oldPath` in jq.

PR #1484, opened 2019-03-01
test: Reclassify ./unit/scheduling as ./integration/scheduling

sebn:
It is not a unit test since it tests many components together.
Also fixed test name to match file & to prevent confusion with Sync unit tests.

- [x] PR is not too big


sebn:
At least it won't prevent integration tests to run anymore in case of random failure :thinking:

PR #1485, opened 2019-03-01
test/support/helpers/local: Fix atom steps init in #simulateAtomEvents()

sebn:
Fixes `TypeError: Cannot match against 'undefined' or 'null'` everywhere
when running scenarios with atom watcher/captures.

Thanks @taratatach for identifying the issue!

Requires checking the watcher type to make flow happy.
The `$FlowFixMe` notice is still needed because of the magic batch.

- [x] PR is not too big

### DIFF
--- a/test/support/helpers/local.js
+++ b/test/support/helpers/local.js
@@ -11,6 +11,7 @@ const conflictHelpers = require('./conflict')
 const { ContextDir } = require('./context_dir')
 
 const Local = require('../../../core/local')
+const atomWatcher = require('../../../core/local/atom_watcher')
 const { TMP_DIR_NAME } = require('../../../core/local/constants')
 const dispatch = require('../../../core/local/steps/dispatch')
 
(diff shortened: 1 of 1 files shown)


test/support/helpers/local.js
```diff
@@ -142,12 +143,19 @@ class LocalTestHelpers {
     }
   }
 
-  simulateAtomEvents (batches /*: Batch[] */) {
+  async simulateAtomEvents (batches /*: Batch[] */) {
+    const { watcher } = this.side
+    if (!(watcher instanceof atomWatcher.AtomWatcher)) {
+      throw new Error(
+        'Cannot only use Local#simulateAtomEvents() with AtomWatcher'
```
taratatach: Nitpick: "Can only"

sebn: Also `s/Local/LocalTestHelpers/`

PR #1487, opened 2019-03-01
test/unit/local/steps/incomplete_fixer: Add tests

taratatach:
Please make sure the following boxes are checked:

- [ ] PR is not too big
- [ ] it improves UX & DX in some way
- [ ] it includes tests matching the implementation changes
- [ ] it includes relevant documentation

PR #1488, opened 2019-03-01
Run scenarios captures on their creation platform only

taratatach:
Based on #1487 

The new atom watcher doesn't generate/handle events the same way on
Windows and GNU/Linux, so better not mix captures & platforms.

- [ ] PR is not too big


sebn:
Reopened as https://github.com/cozy-labs/cozy-desktop/pull/1489 so it doesn't depend on #1487.

sebn:
#1489 was merged

PR #1491, opened 2019-03-04
[WIP] Stopped client atom scenarios

sebn:

- [ ] PR is not too big


sebn:
At least we know which cases are failing.
Enabling stopped client scenarios in CI would require support for disabling failing stopped client cases.
This may be done in a future PR.

PR #1494, opened 2019-03-04
core/local/steps/incomplete_fixer: Handle combined events

sebn:
Please make sure the following boxes are checked:

- [x] PR is not too big
- [x] it improves UX & DX in some way
- [ ] it includes tests matching the implementation changes
- [ ] it includes relevant documentation

PR #1495, opened 2019-03-04
core/local/steps/incomplete_fixer: Extract .step()

sebn:
Minimal builders change + `.loop()` test (nominal case) so we can extract `.step()` and test it more thoroughly without having to deal with buffers.

The code was carefully refactored in order not to conflict with upcoming #1494.
Also, although it will conflict, rebasing #1487 should be relatively easy too.

- [x] PR is not too big


sebn:
Rebased & squashed

PR #1498, opened 2019-03-06
core/local/steps/incomplete_fixer: Overwrite event

taratatach:
When rebuilding an incomplete event, if the completing event and the
  rebuilt one share the same path, we should completely replace the
  completing event since it's now obsolete.

- [x] PR is not too big

### DIFF
--- a/test/unit/local/steps/incomplete_fixer.js
+++ b/test/unit/local/steps/incomplete_fixer.js
@@ -29,7 +29,7 @@ const completionChanges = events => events.map(completedEvent)
 
 describe('core/local/steps/incomplete_fixer', () => {
   describe('.loop()', () => {
-    it('pushes a fixed event into the output buffer', async () => {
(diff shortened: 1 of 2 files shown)


sebn:
Also does it allow some scenario that didn't work before? Does it prevent harmless errors in logs?

taratatach:
Rewriting the `loop` test fixes an error that was raised when running it but the issue fixed in `incomplete_fixer` itself showed up while working on the `loop` test (i.e. I found out that we had 2 events instead of one).

test/unit/local/steps/incomplete_fixer.js
```diff
@@ -131,6 +129,35 @@ describe('core/local/steps/incomplete_fixer', () => {
           ]
         ])
       })
+
+      it.only('replaces the completing event if its path is the same as the rebuilt one', async function () {
```
sebn: `only` :see_no_evil:

PR #1501, opened 2019-03-07
core/local/steps/incomplete_fixer: Avoid rebuilt event errors

taratatach:
- When rebuilding a `renamed` event in incomplete_fixer, we can end up
  with a `renamed` event with equal path and oldPath which is invalid
  and will raise an error at the Merge level.
  We can avoid processing it altogether by dropping it.
- When rebuilding an event, if the destination `path` does not exist
  anymore, trying to get its stats will raise an ENOENT error.
  This path is also the path of the completing event.

  To avoid raising useless errors (i.e. if we don't have a file or
  directory at the given path, we don't need a rebuilt event), we catch
  that error when rebuilding the event and don't return any.
  Also, to avoid keeping events about files or directories that don't
  exist anymore, if we don't have a rebuilt event we drop the completing
  event.

- [x] PR is not too big

### DIFF
--- a/core/local/steps/incomplete_fixer.js
+++ b/core/local/steps/incomplete_fixer.js
@@ -17,8 +17,6 @@ const log = logger({
 //      good value, it is not something that was computed)
 const DELAY = 3000
 
-// TODO add unit tests and logs
-
 /*::
 import type Buffer from './buffer'
 import type { AtomWatcherEvent, Batch } from './event'
@@ -33,6 +31,10 @@ type IncompleteFixerOptions = {
   syncPath: string,
   checksumer: Checksumer
 }
+
+type Completion =
+  | { ignored: false, rebuilt: AtomWatcherEvent }
+  | { ignored: true }
 */
 
 module.exports = {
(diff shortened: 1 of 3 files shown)


test/unit/local/steps/incomplete_fixer.js
```diff
@@ -14,7 +16,7 @@ const incompleteFixer = require('../../../../core/local/steps/incomplete_fixer')
 const Builders = require('../../../support/builders')
 const { ContextDir } = require('../../../support/helpers/context_dir')
 
-const syncPath = __dirname
+const syncPath = fs.mkdtempSync(path.join(os.tmpdir(), 'foo-'))
```
sebn:
```js
before(configHelpers.createConfig)
// Gives you this.syncPath
```

core/local/steps/incomplete_fixer.js
```diff
@@ -74,6 +74,10 @@ async function rebuildIncompleteEvent (item /*: IncompleteItem */, event /*: Ato
       : item.event.oldPath.replace(event.oldPath, event.path)
   }
 
+  if (p === oldPath) {
+    return null
```
sebn: Couldn't we avoir computing the checksum by returning earlier?

core/local/steps/incomplete_fixer.js
```diff
@@ -150,25 +154,30 @@ function step (incompletes /*: IncompleteItem[] */, opts /*: IncompleteFixerOpti
...
-            } else {
-              batch.push(rebuilt)
-            }
+            rebuilt = await rebuildIncompleteEvent(item, event, opts)
           } else if (itemDestinationWasDeleted(item, event)) {
             // We have a match, try to replace the incomplete event
-            const rebuilt = buildDeletedFromRenamed(item, event)
-            log.debug({path: rebuilt.path, action: rebuilt.action}, 'rebuilt event')
-            batch.push(rebuilt)
+            rebuilt = buildDeletedFromRenamed(item, event)
```
sebn:
```js
const detectCompletion = async (item, event, opts) => {
  if (wasRenamedSuccessively(item, event)) {
    // We have a match, try to rebuild the incomplete event
    return rebuildIncompleteEvent(item, event, opts)
  } else if (itemDestinationWasDeleted(item, event)) {
    // We have a match, try to replace the incomplete event
    return buildDeletedFromRenamed(item, event)
  }
}
```

Then:

```js
// Look ma! No let ;)
const completion = await detectCompletion(item, event, opts)
if (!completion) continue
```

Then one could use `completion.rebuilt` to get the event, or maybe some `completion.ignored` prop?

*(also maybe we should find a way at some pointto make the code clear enough so we don't need all those comments, but this is out of the scope of this PR)*

test/unit/local/steps/incomplete_fixer.js
```diff
@@ -263,5 +263,40 @@ describe('core/local/steps/incomplete_fixer', () => {
...
+          .kind('file')
+          .action('renamed')
+          .oldPath(dst)
+          .path(src)
+          .build()
+        const incompletes = []
+        const outputBatches = []
+
+        // Create actual `src` file
+        await syncDir.makeTree(['src'])
```
sebn: I believe `makeTree` is nice when we're mixing dirs/subdirs/files, but here it's easy to wonder whether `src` should actually be `src/`.

Maybe introduce `syncDir.ensureFile('src')`?
Which would probably make the comment useless :wink:

PR #1502, opened 2019-03-07
core/local/steps/win_detect_move: Test & refactor

sebn:
**Needs #1505**
   
It would be interesting to isolate the various side-effects (existing
metadata lookup, timeouts...) and identify their possible combinations
with the correponding race conditions. But this may require quite a lot
of refactoring. So this PR only test event combos, not timing. Still
better than nothing.

Extracted a few functions from implementation.
Also fixed a circular reference in debug data attached to events.

- [x] PR is not too big


sebn:
As discussed: refactored tests & implementation both in the same PR. 🚢?

PR #1505, opened 2019-03-11
test/support/builders: Use realistic test stats

sebn:
**Needs #1504**

Introduces new `StatsBuilder`.
Which builds `WinStats` & generates fileids from test inode numbers.
So we can reuse it in `AtomwatcherEventBuilder` & `BaseMetadataBuilder`.
And make sure our test data actually match our production data.

Since fileids have more precision than inode numbers, it could make sense to
generate inode numbers from fileids instead of the opposite.
But the current way prevent breaking many tests.
And the other one would mostly be relevant to test precision issues.
Which we don't test yet anyway.

- [x] PR is not too big

### DIFF
--- a/test/support/builders/atom_watcher_event.js
+++ b/test/support/builders/atom_watcher_event.js
@@ -1,18 +1,18 @@
 /* @flow */
 
-const fs = require('fs')
 const _ = require('lodash')
 const path = require('path')
 
 const metadata = require('../../../core/metadata')
 const events = require('../../../core/local/steps/event')
 
+const statsBuilder = require('./stats')
+
 /*::
 import type { Stats } from 'fs'
 import type { AtomWatcherEvent, EventAction, EventKind, Batch } from '../../../core/local/steps/event'
+import type { StatsBuilder } from './stats'
 */
-
-const NON_EXECUTABLE_MASK = 0 << 6
 
 function randomPick /*:: <T> */ (elements /*: Array<T> */) /*: T */{
   const l = elements.length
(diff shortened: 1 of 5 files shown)


sebn:
Before merging, I'd like to make sure Metadata#fileid are actually necessary in #1502

test/support/builders/atom_watcher_event.js
```diff
@@ -96,13 +87,12 @@ module.exports = class AtomWatcherEventBuilder {
...
 
   ino (newIno /*: number */) /*: this */ {
-    if (this._event.stats == null) this._event.stats = buildStats(this._event.kind)
-    this._event.stats.ino = newIno
+    this._ensureStatsBuilder().ino(newIno)
     return this
   }
 
   noStats () /*: this */ {
-    delete this._event.stats
```
taratatach: We should keep this line. If we build an event atop another one that has stats, the `stats` attribute will be set.

test/support/builders/metadata/base.js
```diff
@@ -111,6 +112,9 @@ module.exports = class BaseMetadataBuilder {
   }
 
   ino (ino /*: number */) /*: this */ {
+    if (process.platform === 'win32') {
+      this.doc.fileid = statsBuilder.fileIdFromNumber(ino)
```
taratatach: Should we remove the `fileid()` method?

taratatach: Should we use the `StatsBuilder` when calling the `stats()` method as well?

test/support/builders/stats.js
```diff
@@ -0,0 +1,122 @@
...
+        fileid: fileIdFromNumber(commonDefaults().ino),
+        directory: false,
+        symbolicLink: false
+      } /*: Object */),
+      commonDefaults()
+    )
+  }
+
+  ino (newIno /*: number */) /*: this */ {
+    if (process.platform === 'win32') {
```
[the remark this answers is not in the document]
sebn: Not sure... I'll check. Thanks!

test/support/builders/metadata/base.js
sebn: I believe I still had to use it somewhere, but I may be wrong. Will check. Thanks!

sebn: Probably :)

sebn: I was wrong. Good catch!

test/support/builders/stats.js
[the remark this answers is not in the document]
sebn: It was not necessary. Good catch!

sebn:
Fixed code according to review.
To be squashed as soon as approved & CI is green.

sebn:
@taratatach Are you fine with the new changes?

test/support/builders/metadata/base.js
taratatach: OK. It's a confusing name then. But it's out of scope.

PR #1508, opened 2019-03-14
core: Drop Sync#reuploadContentMismatchFiles()

sebn:
It was used in production for quite some time.
Files not yet fixed probably won't be.
    
Also dropped correponding untested RemoteCozy method.
But kept file size check since it may help prevent future issues.
    
This almost reverts the corresponding commit: e668a41

---

- [x] PR is not too big


sebn:
Kept file size check + added corresponding unit tests. Should be ready now.

PR #1511, opened 2019-03-15
Avoid recomputing checksums

taratatach:
When an event matches an existing document's ino, path and last update
  time, we can safely assume the checksum hasn't changed and set it on
  the event.

When an event already has a checksum, it's not necessary to recompute
  it in this step.
  This will speed up the initial scan by avoiding lots of checksum
  computations.

- [x] PR is not too big

### DIFF
--- a/core/local/steps/initial_diff.js
+++ b/core/local/steps/initial_diff.js
@@ -20,6 +20,8 @@ type InitialDiffState = {
 type WatchedPath = {
   path: string,
   kind: EventKind,
+  md5sum?: string,
+  updated_at: string
 }
 
 type WaitingItem = {
@@ -69,7 +71,13 @@ async function initialState (opts /*: { pouch: Pouch } */) /*: Promise<{ [typeof
     if (doc.ino != null) {
       // Process only files/dirs that were created locally or synchronized
       const kind = doc.docType === 'file' ? 'file' : 'directory'
-      byInode.set(doc.fileid || doc.ino, { path: doc.path, kind: kind })
+      const was /*: WatchedPath */ = {
+        kind,
+        path: doc.path,
+        updated_at: doc.updated_at
+      }
+      if (kind === 'file') was.md5sum = doc.md5sum
+      byInode.set(doc.fileid || doc.ino, was)
     }
   }
 
@@ -122,6 +130,9 @@ async function initialDiff (buffer /*: Buffer */, out /*: Buffer */, pouch /*: P
               path: was.path
             })
           }
+        } else if (foundUntouchedFile(event, was)) {
+          _.set(event, [STEP_NAME, 'md5sumReusedFrom'], was.path)
+          event.md5sum = was.md5sum
         }
       }
 
@@ -198,5 +209,17 @@ function debounce (waiting /*: WaitingItem[] */, events /*: AtomWatcherEvent[] *
         }
       }
     }
+  }
+}
+
+function foundUntouchedFile (event, was) {
+  if (was && event.kind === 'file') {
+    const { ctime, mtime } = event.stats
+    const eventUpdateTime = Math.max(ctime.getTime(), mtime.getTime())
+    const docUpdateTime = (new Date(was.updated_at)).getTime()
+
+    return eventUpdateTime === docUpdateTime
+  } else {
+    return false
   }
 }
--- a/test/support/builders/atom_watcher_event.js
+++ b/test/support/builders/atom_watcher_event.js
@@ -91,6 +91,16 @@ module.exports = class AtomWatcherEventBuilder {
     return this
   }
 
+  mtime (newMtime /*: Date */) /*: this */ {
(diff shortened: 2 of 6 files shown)


test/support/builders/atom_watcher_event.js
```diff
@@ -91,6 +91,11 @@ module.exports = class AtomWatcherEventBuilder {
     return this
   }
 
+  updatedAt (mtime /*: Date */) /*: this */ {
```
sebn: Should we rename this method to `mtime`?

`AtomWatcherEvent` type [doesn't have an `updatedAt` property](https://github.com/cozy-labs/cozy-desktop/blob/v3.13.0-alpha.1/core/local/steps/event.js#L19-L29).
And `Metadata#updatedAt` can be set to [either mtime or ctime](https://github.com/cozy-labs/cozy-desktop/blob/v3.13.0-alpha.1/core/metadata.js#L473) (depending on whether the file metadata or content were modified).

So I believe `#updatedAt` may be confusing here.

test/support/builders/stats.js
```diff
@@ -100,6 +106,11 @@ class WinStatsBuilder {
     return this
   }
 
+  mtime (newMtime /*: Date */) /*: this */ {
+    this.winStats.mtime = new Date(newMtime)
+    return this
```
sebn: Having 2 classes implementing `StatsBuilder` actually favors duplication. My bad.
Out of the scope of this PR anyway.

core/local/steps/initial_diff.js
```diff
@@ -122,6 +131,9 @@ async function initialDiff (buffer /*: Buffer */, out /*: Buffer */, pouch /*: P
               path: was.path
             })
           }
+        } else if (was && event.kind === 'file' && sameUpdateTime(was.updated_at, event.stats.mtime)) {
```
sebn: `sameUpdateTime()` feels a bit weird, names says *update*, but takes a string and a date...
What about extracting the whole condition?

taratatach: It really says _update time_ ;-).
But I'll find something better that encompasses the whole condition.

test/unit/local/steps/initial_diff.js
```diff
@@ -151,19 +161,130 @@ describe('local/steps/initial_diff', () => {
...
+        .ino(fizz.ino)
+        .ctime(new Date(fizz.updated_at))
+        .build()
+      const updatedMetadataScan = builders.event()
+        .action('scan')
+        .kind(kind(updatedMetadata))
+        .path(updatedMetadata.path)
+        // $FlowFixMe: updatedMetadata does have an ino
+        .ino(updatedMetadata.ino)
+        .mtime(new Date((new Date(updatedMetadata.updated_at)).getTime() + 1000))
```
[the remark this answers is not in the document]
taratatach: I'm not sure we'd have a big usage of such functions but why not.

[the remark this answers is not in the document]
sebn: Makes sense. Then maybe just a local helper in the test?

core/local/steps/initial_diff.js
```diff
@@ -200,3 +211,15 @@ function debounce (waiting /*: WaitingItem[] */, events /*: AtomWatcherEvent[] *
     }
   }
 }
+
+function foundUntouchedFile (event, was) {
+  if (was && event.kind === 'file') {
+    const { ctime, mtime } = event.stats
+    const eventUpdateTime = Math.max(ctime.getTime(), mtime.getTime())
```
sebn: Shouldn't we check `mtime` only regarding content?

core/local/steps/initial_diff.js
```diff
@@ -122,6 +130,9 @@ async function initialDiff (buffer /*: Buffer */, out /*: Buffer */, pouch /*: P
               path: was.path
             })
           }
+        } else if (foundUntouchedFile(event, was)) {
+          _.set(event, [STEP_NAME, 'md5sumReusedFrom'], was.path)
```
sebn: Still not sure why a boolean is not sufficient.

core/local/steps/initial_diff.js
sebn: From what I understand, it may be safer actually, since some third-party code could modify content then revert mtime (but not ctime), but still improve performances.

core/local/steps/initial_diff.js
sebn: My bad. Since we're actually looking for corresponding metadata in multiple steps, there's no guarantee we'll get the same doc everytime. So maybe you were right and we should actually attach the doc `_id` and `_rev`... WDYT?

sebn: `path` is actually more specific than `_id` so better keep it.

taratatach: True we could get a different doc. Which means `was` should probably be a complete `Metadata` and we should attach it.

PR #1515, opened 2019-03-19
chore(deps): Upgrade to electron 2.0

sebn:
So at least we get security updates.

---

- [x] PR is not too big


sebn:
Works like a charm on GNU/Linux. I still need to test it on Windows/macOS.

PR #1520, opened 2019-03-21
test: Fix scenarios initial diff state issue

sebn:
So initial diff doesn't end up with state before initial scan while
we're testing live captures in scenarios.

---

- [x] PR is not too big


sebn:
- Extracted `LocalTestHelpers#simulateAtomStart()`
- Reused it in test scenarios runner
- Reused it to fix `test/integration/mtime-update.js`

CI should be green now.

sebn:
@taratatach :ship: ?

PR #1545, opened 2019-04-16
core/local/steps/await_write_finish: Aggregate in batch

taratatach:
The step would only work if the first event on which we'll aggregate
  others was in a previous batch.
  If a lot of writes happen quickly we'll see batches of modified
  events for the same file / directory and we want to aggregate those
  to limit the number of changes to merge and sync.

  To do that, we go over the events of the current batch and aggregate
  them before trying to aggregate them with waiting events.

- [x] PR is not too big


taratatach:
@sebn You should read this PR even after it's merged since `aggregateBatch` has been re-written.

PR #1555, opened 2019-04-24
core/local/steps/initial_diff: Fix renamed events #1555

taratatach:
When transforming `scan` events into `renamed` events during the
  initial scan, we don't know if some of the moved documents are
  children of moved directories.
  This is a problem because when dispatching them, we won't find any
  document corresponding to their oldPath in Pouch since they would have
  been moved just before.

  To make sure all stopped client moves are correctly detected and
  merged, we keep track of `renamed` events that are issued by the
  initialDiff step or the watcher during the initial scan and use them
  to correct any of their children's movements.

Please make sure the following boxes are checked:

- [x] PR is not too big
- [x] it improves UX & DX in some way
- [x] it includes tests matching the implementation changes
- [x] it includes relevant documentation

PR #1556, opened 2019-04-25
test/integration/scheduling: Drop test

sebn:
- It is slow
- It is hard to understand / maintain
- It generates false positives
- There is already a unit test for this


sebn:
To be squahed & merged after #1555

PR #1557, opened 2019-04-25
release: 3.13.0-beta.8

sebn:
Improvements for Windows and GNU/Linux users:

- We fixed the last known issue regarding files and directories renamed while
  the client is stopped using the new watcher. The next release should be a
  stable one! 🚀

See also [known issues](https://github.com/cozy-labs/cozy-desktop/blob/master/KNOWN_ISSUES.md).

Happy syncing!


sebn:
To be squashed & merged after #1555 and #1556

PR #1562, opened 2019-04-30
core/local/steps/initial_diff: Fix path of issued deleted events

taratatach:
When we don't receive any events for an existing document during the
  initial scan, we assume it has been deleted and issue a `deleted`
  event for it.
  However, if its parent has been moved or renamed, its path won't be
  correct anymore and we won't be able to merge it.
  To issue a correct event, we fix its path attribute based on its
  parents movements.

Based on #1562

- [x] PR is not too big


sebn:
(rebased)

PR #1532, opened 2019-04-03
yarn: Upgrade dependencies which don't require changes

sebn:
The following will be upgraded separately:
    
- cozy-ui probably requires GUI changes
- latest cheerio is a RC (better keep the current stable)
- electron 3 & 4
- electron-builder (needs to be upgraded with electron-updater)
- electron-updater 4 (needs electron 3)
- eslint* (will be upgraded next while switching to cozy-app)
- mocha, electron-mocha, should, sinon... (many major bumps)
- mime (API change, easy)
- opn (easy, package was just renamed to *open*)

---

- [x] PR is not too big
- [x] it improves UX & DX in some way
- [x] it includes tests matching the implementation changes
- [x] it includes relevant documentation

PR #1567, opened 2019-05-02
[WIP] test/integration/gui/setupProxy: Basic networking

sebn:
This test should fail with a2b3e94b / #1532.
So we can upgrade dependencies without silently breaking everything.

---

- [x] PR is not too big


sebn:
There are 2 underlying issues actually. This test only catches one of them. Not ready to merge yet.

PR #1571, opened 2019-05-06
core/remote/change: Fix deleted-created sort

taratatach:
To re-order changes coming from the remote changesFeed, we use the
  javascript `sort()` function which compares 2 elements at a time,
  elements `a` and `b`.
  In our custom sorter, to make sure changes which delete (i.e. in Pouch
  terms) a specific id come before changes which create the same id, we
  have 2 comparisons following each other :
  - first we check if change `a` deletes the id created by change `b`
  - second, if the first comparison was falsy, we check if change `b`
  deletes the id created by change `a`

  Due to the way to compute the id deleted or created by a change, if
  `b` deletes the id and `a` creates it (i.e. we should be in the second
  comparison case), the first comparison we return a truthy value and
  the creation will come first, leading to conflicts in Merge later.

  To make sure our comparisons are valid, we need to make sure we
  compare ids which are not null (e.g. we should not try to compare the
  created id of a change if that change actually deletes it).

  Expected changes in test/unit/remote/watcher.js were re-ordered to
  reflect the new expected order (deletions first).

- [x] PR is not too big

### DIFF
--- a/test/unit/remote/change.js
+++ b/test/unit/remote/change.js
@@ -18,4 +18,28 @@ describe('remote change sort', () => {
     remoteChange.sort(a)
     a.should.deepEqual([parent, child])
   })
+
+  describe('sorts deleted before created for the same path', () => {
+    const deleted = {
+      doc: { path: 'parent/file' },
+      type: 'FileDeletion'
+    }
+
+    const created = {
+      doc: { path: 'parent/file' },
+      type: 'FileAddition'
+    }
+
+    it('when deleted comes before created', () => {
+      const changes = [deleted, created]
+      remoteChange.sort(changes)
+      changes.should.deepEqual([deleted, created])
+    })
+
+    it('when created comes before deleted', () => {
+      const changes = [created, deleted]
+      remoteChange.sort(changes)
+      changes.should.deepEqual([deleted, created])
+    })
+  })
 })
--- a/core/remote/change.js
+++ b/core/remote/change.js
@@ -269,8 +269,10 @@ const sorter = (a, b) => {
   if (areParentChild(deletedId(b), deletedId(a))) return aFirst
   if (areParentChild(deletedId(a), deletedId(b))) return bFirst
 
-  if (deletedId(a) === createdId(b)) return aFirst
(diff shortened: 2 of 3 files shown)


test/unit/remote/change.js
```diff
@@ -18,4 +18,30 @@ describe('remote change sort', () => {
     remoteChange.sort(a)
     a.should.deepEqual([parent, child])
   })
+
+  describe('sorts deleted before created for the same path', () => {
+    const deleted = {
+      doc: { path: 'parent/file' },
+      type: 'FileDeletion',
+      was: { path: 'parent/file' }
```
sebn: `RemoteFileDeletion` is not supposed to have a `was` (and the test still pass without it).

test/unit/remote/change.js
```diff
@@ -18,4 +18,30 @@ describe('remote change sort', () => {
...
+    }
+
+    const created = {
+      doc: { path: 'parent/file' },
+      type: 'FileAddition',
+      was: null
```
sebn: Maybe just omit `was`?

test/unit/remote/change.js
taratatach: I thought I'd seen it in logs :thinking:. I just wanted to reproduce that situation.
I'll remove the `was` then.

test/unit/remote/change.js
taratatach: If that's how are formed `FileAddition` changes, yes.

sebn: You're right, it would actually be set to `null` by RemoteWatcher, my bad.

sebn:
@taratatach `RemoteWatcher#pullMany()` unit tests are failing :wink:

test/unit/remote/change.js
sebn: I'm almost sure for this one: the property is not part of the type and when logging `FileDeletion` on some scenario it has a `doc` but not a `was`.

taratatach:
> @taratatach `RemoteWatcher#pullMany()` unit tests are failing wink

Fixed

PR #1584, opened 2019-05-13
gui/js/proxy: Refactor to test

sebn:
- Stop parsing command-line arguments on module import
- Export multiple functions so we can test them independently:
  * Replacing `setupProxy()` with `proxy.setup()`.
- Pass `config` to `proxy.setup()` so we can later test multiple cases.
- Pass original functions before monkey-patching to callback, so they
  can be reverted.
- Introduce self-signed pfx certificate for testing purpose
- Start adding tests for the various patched global functions (still no
  proxy setup testing yet):
  * Checks user-agent
  * At least ensures functions don't crash

Remaining work (to be completed later):

- Find a way to revert `allowNTLMCredentialsForDomains()`
- Find a way to revert `setProxy()`
- Find a way to test `fetch()` with our self-signed pfx
- Test various proxy configs

---

- [x] PR is not too big


sebn:
As discussed, rebased to merge partial work.

sebn:
Unlink the previous attempt to call `setupProxy()` before running every tests, it doesn't seem to break the CI, maybe because we're partially reverting stuff...

sebn:
I wonder whether we should keep passing `app`, `session` and `userAgent` to `proxy.setup()`.
We never pass anything else than the default ones. And it wouldn't prevent testing anyway.

PR #1586, opened 2019-05-16
core: Use our own rev to determine changes to apply

taratatach:
We don't control the Pouch's `_rev` attribute set on each document.
When we move a document, the destination will be a new document  in
Pouch. This means its `_rev` will probably start with 1 (it can be
higher if the path has previously existed but it does not matter
here).

Since our mechanism to determine which changes need to be applied on
which side is based on the `_rev` value (i.e. its short rev or
beginning number to be precise) and the `sides` we maintain, we
cannot keep the existing `sides` value since they will probably not be
correlated with the new `_rev` anymore.
This means we're losing information when doing a move.

This loss prevents us from being able to merge some combinations of
changes.
e.g. With the client stopped:
1. add document to a folder locally
2. move that folder remotely
3. launch client to synchronize changes

The file will be moved as a child of the moved directory and its
`sides` attribute will contain `{ remote: 1 }` (i.e. we're losing the
local addition information).
In this situation, we're not able to move the file since it doesn't
exist remotely and we don't have any remote id for it and we won't try
to upload it to the Cozy since we see a remotely created file.

By maintaining our own `rev` attribute, we're able to keep a
correlation between the `rev` and the `sides` values even when moving
a document.
In our previous example, we'll be easily able to mark the file as created
locally in the new directory.

We decided to add this `rev` attribute inside the `sides` attribute
since they work together.
The `MetadataSidesInfos` type looks like this now:
```
{
  rev: number,
  local?: number,
  remote?: number
}
```

Out faked docs coherence was also improved by:
- Improving Metdata builder's `upToDate()` method to take into account
  existing sides (i.e. if doc is based on another one)
- Using builder's helper methods to set sides whenever possible
- Avoiding faking movements when possible (to avoid unnecessary side
  effects when running the move method we're testing)

- [ ] PR is not too big

### DIFF
--- a/core/merge.js
+++ b/core/merge.js
@@ -5,7 +5,6 @@
 
 const autoBind = require('auto-bind')
 const _ = require('lodash')
-const { clone } = _
 const path = require('path')
 
 const IdConflict = require('./IdConflict')
@@ -138,7 +137,7 @@ class Merge {
     side /*: SideName */,
     doc /*: Metadata */,
     was /*: ?Metadata */
-  ) {
+  ) /*: Promise<Metadata> */ {
     log.warn(
       { path: doc.path, oldpath: was && was.path, doc, was },
       'resolveConflictAsync'
@@ -219,8 +218,7 @@ class Merge {
     }
     await this.ensureParentExistAsync(side, doc)
 
-    const result = await this.pouch.put(doc)
-    return this.fixSideInPouch({ side, result, doc })
+    return this.pouch.put(doc)
   }
 
   // Update a file, when its metadata or its content has changed
@@ -331,8 +329,7 @@ class Merge {
     }
     await this.ensureParentExistAsync(side, doc)
 
-    const result = await this.pouch.put(doc)
-    return this.fixSideInPouch({ side, result, doc })
+    return this.pouch.put(doc)
   }
 
   // Rename or move a file
@@ -387,8 +384,7 @@ class Merge {
       } else if (file && !doc.overwrite && doc.path === file.path) {
         const dst = await this.resolveConflictAsync(side, doc, file)
         was.moveTo = dst._id
-        dst.sides = {}
-        dst.sides[side] = 1
+        dst.sides = { target: 1, [side]: 1 }
         return this.pouch.bulkDocs([was, dst])
       } else if (file && doc.overwrite) {
         doc._rev = file._rev
@@ -397,8 +393,7 @@ class Merge {
       } else {
         await this.ensureParentExistAsync(side, doc)
 
-        const results = await this.pouch.bulkDocs([was, doc])
-        return this.bulkFixSideInPouch({ side, results, docs: [doc] })
+        return this.pouch.bulkDocs([was, doc])
       }
     } else {
       // It can happen after a conflict
@@ -482,8 +477,8 @@ class Merge {
     )
 
     for (let doc of docs) {
-      let src = clone(doc)
-      let dst = clone(doc)
+      let src = _.cloneDeep(doc)
+      let dst = _.cloneDeep(doc)
       dst._id = makeDestinationID(doc)
       dst.path = doc.path.replace(was.path, folder.path)
       if (src.sides && src.sides[side] && !src.sides[otherSide(side)]) {
@@ -510,8 +505,7 @@ class Merge {
       else delete dst.incompatibilities
       bulk.push(dst)
     }
-    const results = await this.pouch.bulkDocs(bulk)
-    return this.bulkFixSideInPouch({ side, results, docs: bulk })
+    return this.pouch.bulkDocs(bulk)
   }
 
   async restoreFileAsync(
@@ -575,7 +569,7 @@ class Merge {
       return this.pouch.put(oldMetadata)
     }
     delete oldMetadata.errors
-    const newMetadata = clone(oldMetadata)
+    const newMetadata = _.cloneDeep(oldMetadata)
     metadata.markSide(side, newMetadata, oldMetadata)
     newMetadata._id = doc._id
     newMetadata._rev = doc._rev
@@ -621,8 +615,7 @@ class Merge {
         } else {
           // When "unlinked" from the local side, a folder doesn't have sides
           // information.
-          was.sides = {}
-          was.sides[otherSide(side)] = 1
+          was.sides = { target: 1, [otherSide(side)]: 1 }
         }
         return this.putFolderAsync(otherSide(side), was)
       }
--- a/core/metadata.js
+++ b/core/metadata.js
@@ -84,6 +84,7 @@ type RemoteRev = string
 export type RemoteRevisionsByID = { [RemoteID] : RemoteRev}
 
 export type MetadataSidesInfo = {
+  target: number,
   remote?: number,
   local?: number
 }
@@ -161,6 +162,8 @@ module.exports = {
   sameBinary,
   markSide,
   incSides,
+  side,
+  target,
   wasSynced,
   buildDir,
   buildFile,
@@ -374,16 +377,12 @@ function extractRevNumber(doc /*: Metadata|{_rev: string} */) {
 }
 
 // Return true if the remote file is up-to-date for this document
-function isUpToDate(side /*: SideName */, doc /*: Metadata */) {
-  let currentRev = doc.sides[side] || 0
-  let lastRev = extractRevNumber(doc)
-  return currentRev === lastRev
+function isUpToDate(sideName /*: SideName */, doc /*: Metadata */) {
+  return side(doc, sideName) === target(doc)
 }
 
-function isAtLeastUpToDate(side /*: SideName */, doc /*: Metadata */) {
-  let currentRev = doc.sides[side] || 0
-  let lastRev = extractRevNumber(doc)
-  return currentRev >= lastRev
+function isAtLeastUpToDate(sideName /*: SideName */, doc /*: Metadata */) {
+  return side(doc, sideName) >= target(doc)
 }
 
 function markAsUnsyncable(side /*: SideName */, doc /*: Metadata */) {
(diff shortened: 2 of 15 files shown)


test/unit/merge.js
```diff
@@ -1215,72 +1216,103 @@ describe('Merge', function() {
...
-        .metafile()
+      const file2 = builders
+        .metafile(file)
         .path('SRC/FILE2')
-        .moveFrom(orig)
-        .sides({ [this.side]: 3, [otherSide(this.side)]: 2 })
-        .remoteId(dbBuilders.id())
-        .create()
+        .build()
+      await this.merge.moveFileAsync(
```
sebn: Maybe we could wrap the build+merge+get logic with some helper to make this test a bit shorter & reuse it in the following one too?

Either a plain function:

```js
const was = await builders.move(file, 'SRC/FILE2')
```

Or a builder class in case it makes sense:

```js
const was = await builders.move(file).to('SRC/FILE2').create()
```

Assuming `Merge` is injected into `Builders` together with `Pouch`:

```js
// replacing:
builders = new Builders({ pouch: this.pouch })
// with:
builders = new Builders(this)
```

test/unit/merge.js
```diff
@@ -2182,11 +2206,13 @@ describe('Merge', function() {
           .metadir()
           .path('SRC_DIR')
           .upToDate()
+          .remoteId(dbBuilders.id())
           .create()
         const child = await builders
           .metadata()
           .path(`SRC_DIR/${childName}`)
           .upToDate()
+          .remoteId(dbBuilders.id())
```
sebn: Shouldn't `.upToDate()` imply `.remoteId()`?

test/unit/merge.js
```diff
@@ -1215,72 +1216,103 @@ describe('Merge', function() {
...
+      await this.merge.moveFolderAsync(
+        'local',
+        _.cloneDeep(dst),
+        _.cloneDeep(src)
+      )
+      const was = await this.pouch.db.get(
+        builders
+          .metafile()
+          .path('DST/FILE')
+          .build()._id
```
sebn: `metadata.id('DST/FILE')`?
Ideally we should rename `Pouch#byPath()` to something less confusing, e.g. `#byParentPath()`, then reuse the `Pouch#byPath()` name for this.

core/local/atom/initial_diff.js
```diff
@@ -89,6 +89,10 @@ async function initialState(
   const byInode /*: Map<number|string, Metadata> */ = new Map()
   const docs /*: Metadata[] */ = await opts.pouch.byRecursivePathAsync('')
   for (const doc of docs) {
+    if (metadata.needsRevMigration(doc)) {
+      await metadata.migrateRev(doc, opts)
```
sebn: Maybe rename to something like `#needsTargetMigration()` & `#migrateTarget()`, same as `sides.rev` vs `sides.target`?

core/metadata.js
```diff
@@ -675,3 +678,21 @@ function shouldIgnore(
     isFolder: doc.docType === 'folder'
   })
 }
+
+function needsRevMigration(existing /*: Metadata */) /*: boolean %checks */ {
+  return existing.sides == null || existing.sides.rev == null
```
sebn: Maybe explain that the first case is not supposed to happen but is here to make flow happy because of our permissive type annotations?

core/local/chokidar/initial_scan.js
```diff
@@ -48,6 +48,10 @@ const detectOfflineUnlinkEvents = async (
   let unappliedMoves = []
 
   for (const doc of docs) {
+    if (metadata.needsRevMigration(doc)) {
+      await metadata.migrateRev(doc, { pouch })
```
sebn: Maybe we could `pouch` directly?
(I suppose there's no reason we would need anything else in `.migrateRev()`)

core/metadata.js
```diff
@@ -675,3 +678,21 @@ function shouldIgnore(
...
+  { pouch } /*: { pouch: Pouch } */
+) /*: Promise<Metadata> */ {
+  log.info({ path: existing.path, _rev: existing._rev }, 'Migrating _rev')
+  const doc = _.defaultsDeep(
+    { sides: { rev: extractRevNumber(existing) } },
+    existing
+  )
+  const { rev: _rev } = await pouch.put(doc)
+  doc._rev = _rev
+  return doc
```
sebn: Both mutating & returning new value?

core/merge.js
```diff
@@ -386,8 +383,7 @@ class Merge {
       } else if (file && !doc.overwrite && doc.path === file.path) {
         const dst = await this.resolveConflictAsync(side, doc, file)
         was.moveTo = dst._id
-        dst.sides = {}
-        dst.sides[side] = 1
+        dst.sides = { rev: 1, [side]: 1 }
```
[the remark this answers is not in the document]
taratatach: That's what we're doing :) (probably not in the commit you were looking at?)

core/pouch.js
```diff
@@ -354,7 +354,7 @@ class Pouch {
   /* Helpers */
 
   // Retrieve a previous doc revision from its id
-  getPreviousRev(id, shortRev, callback) {
+  getPreviousRev(id, revDiff, callback) {
```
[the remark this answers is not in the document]
taratatach: I missed the type :+1:

core/sync.js
```diff
@@ -349,8 +349,8 @@ class Sync {
       log.debug({ path: doc.path }, `Applying else for ${doc.docType} change`)
       let old
       try {
-        old = await this.pouch.getPreviousRevAsync(doc._id, rev)
-      } catch (_) {
+        old = await this.pouch.getPreviousRevAsync(doc._id, doc.sides.rev - rev)
```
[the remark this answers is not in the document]
taratatach: We don't care what's the document's `_rev` value here. But this is the mechanism Pouch uses to mark revisions and we have to deal with it.
We're lucky the short rev is incremental since it gives us a way to fetch the previous nth revision.

core/sync.js
```diff
@@ -514,8 +514,9 @@ class Sync {
         await this.pouch.put({
           ...unsynced,
           sides: {
-            [side]: metadata.extractRevNumber(doc) + 1,
-            [other]: unsynced.sides[other] + 1
+            rev: unsynced.sides.rev + 1, // increase rev because of new merge
+            [side]: doc.sides.rev,
+            [other]: unsynced.sides[other] + 1 // increase side to mark change as applied
```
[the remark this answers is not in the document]
taratatach: Yes unfortunately. Maybe some functions would help by removing the need to remember and documenting it.

test/unit/merge.js
taratatach: I've thought about it yes. I wanted to avoid creating new errors in other tests but that would be desirable.

test/unit/merge.js
taratatach: The issue with `metadata.id()` is that it does not normalize paths. So in order to make this test pass, we either need to require the `path` module and rename every occurrence of the `path` variable to avoid collision or use the builder instead which already normalizes paths.

I don't really understand the link with `Pouch#byPath`.

core/metadata.js
taratatach: I don't remember writing it for Flow's sake but I might have done it without thinking.
Are we sure this can never happen?

core/local/chokidar/initial_scan.js
taratatach: It was convenient to pass the options this way, especially in Atom's dispatch step.
But I'm fine with passing `Pouch` only.

core/metadata.js
taratatach: Yes I'm not super happy about it but I wanted to avoid another request to Pouch to get the new `doc`.
I'm open to suggestions.

core/metadata.js
```diff
@@ -541,25 +542,29 @@ function markSide(
   doc /*: Metadata */,
   prev /*: ?Metadata */
 ) /*: Metadata */ {
-  let rev = 0
-  if (prev) {
-    rev = extractRevNumber(prev)
-  }
+  const prevSides = prev && prev.sides
+  let prevRev = rev(prevSides)
```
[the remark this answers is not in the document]
taratatach: Why not :)

core/local/atom/initial_diff.js
```diff
@@ -89,6 +89,10 @@ async function initialState(
   const byInode /*: Map<number|string, Metadata> */ = new Map()
   const docs /*: Metadata[] */ = await opts.pouch.byRecursivePathAsync('')
   for (const doc of docs) {
+    if (metadata.needsRevMigration(doc)) {
```
nono: I wouldn't have put the migration inside the local watcher. Instead, I'd have added a design doc to pouch with a schema version, and when the cozy-desktop starts and initializes the database, it loads this document and if the schema version is too old, it plays the migration(s) from the old schema version up to the current schema version. And only after that, you can call `sync`.

I'm not saying that my method is better, or that you must change to it. It's just a personal preference to not have the migration stuff inside the watcher logic.

taratatach: I'm not well versed into Pouch so I just didn't know about this method.

I find it much better than mine!

taratatach:
It looks like we don't need all this after all. I'm closing until it is found necessary.

taratatach:
It seems after all that the simpler solution to the problem of https://github.com/cozy-labs/cozy-desktop/pull/1637 does not work and we'll need this PR for it to work.

core/sync.js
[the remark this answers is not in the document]
taratatach: Actually, `metadata.markSide()` does this job in most places. We can't use it here because we're modifying both sides with different values but this is not the common update to sides.

test/support/builders/metadata/base.js
```diff
@@ -173,7 +173,9 @@ module.exports = class BaseMetadataBuilder {
   }
 
   upToDate() /*: this */ {
-    this.doc.sides = { rev: 2, local: 2, remote: 2 }
+    const sides = this.doc.sides || { rev: 1 }
+    const newRev = sides.rev + 1
```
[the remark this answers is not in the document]
taratatach: I'm actually using `metadata.markAsUpToDate()` here so we don't have to re-implement it.

core/metadata.js
taratatach: This is done via a migration now.

core/metadata.js
```diff
@@ -407,7 +408,7 @@ function upToDate(doc /*: Metadata */) /*: Metadata */ {
 
   return _.assign(clone, {
     errors: undefined,
-    sides: { local: rev, remote: rev }
+    sides: { rev, local: rev, remote: rev }
```
nono: I'd have expected `target: rev`. Why is it just `rev` here?

taratatach: Good catch. It's a leftover from a previous version of the PR in which `target` was still `rev`.
And Flow didn't catch this…

taratatach: Actually, the function is not used anymore. I've removed it.

PR #1590, opened 2019-05-20
core/app: Temporarily disable Sentry

sebn:
To stop spamming infra. Until we find a solution.

---

- [x] PR is not too big


sebn:
Disabling sentry setup was breaking the app.
So I disabled message handling directly: It's less convoluted and the app still works.

PR #1591, opened 2019-05-20
test/scenarios: New disabled scenario syntax

sebn:
In order to disable a test scenario, we were previously appending a
`.DISABLED` extension to either the `scenario.js` file, some subdir,
or some `*.json` capture file.

Now, one can set the `disabled` property in the scenario file without
having to rename anything.

To disable the whole scenario:

```js
{
  // ...
  disabled: 'Some explanation'
}
```

To disable some tests:

```js
{
  // ...
  disabled: {
    'atom/win32': 'Some explanation',
    'local/darwin': 'Another explanation',
    'stopped': 'Yet another one',
    'remote': 'Explain again'
  }
}
```

For each case, an explanation should be provided so we know why the
scenario is currently unsupported.

Now linting & syntax highlighting should work for disabled files.
And disabled cases listing should be easier to implement.

---

- [x] PR is not too big


sebn:
- [ ] Do not disable capture that were already removed

test/scenarios/restore/scenario.js
```diff
@@ -4,18 +4,18 @@
 
 module.exports = ({
   side: 'remote',
+  disabled:
+    "Not sure why we don't handle this case yet. At least because of init trashed property.",
   init: [
     { ino: 1, path: 'parent/' },
-    { ino: 2, path: 'parent/dir/', trashed: true },
```
taratatach: Why did you comment `trashed: true`?

test/support/helpers/scenarios.js
```diff
@@ -90,8 +86,9 @@ module.exports.loadFSEventFiles = scenario => {
   const disabledEventsFile = name => {
     if (process.platform === 'win32' && name.indexOf('win32') === -1) {
       return 'darwin/linux test'
-    } else if (name.endsWith(disabledExtension)) {
-      return 'disabled case'
+    } else {
+      const disabled = scenario[`local/${name}`]
```
taratatach: Shouldn't this be ```scenario.disabled[`local/${name}`]```?

test/support/helpers/scenarios.js
```diff
@@ -116,8 +113,9 @@ module.exports.loadAtomCaptures = scenario => {
       return 'linux test'
     } else if (process.platform === 'linux' && name.indexOf('win32') >= 0) {
       return 'win32 test'
-    } else if (name.endsWith(disabledExtension)) {
-      return 'disabled case'
+    } else {
+      const disabled = scenario[`atom/${name}`]
```
taratatach: Shouldn't this be ```scenario.disabled[`atom/${name}`]```?

test/scenarios/restore/scenario.js
sebn: Flow was complaining about it because it is not part of the `Scenario` type.
I assumed we would figure it when working to make the scenario pass.

- [ ] Add a comment.

sebn:
Rebased on master, fixed up commits according to review + a bit of refactoring. Waiting for AppVeyor cozies to be available.

sebn:
- [ ] There are still some `*.DISABLED` files/dirs

sebn:
Waiting for beta publication before restarting builds.

test/support/helpers/scenarios.js
```diff
@@ -83,20 +79,34 @@ if (module.exports.scenarios.length === 0) {
...
+  typeof scenario.disabled === 'string' && scenario.disabled
+
+/**
+ * For historical reasons, `disabledScenarioTest()` may receive test names
+ * with file extension. Although it may be useless in many cases, stripping
+ * them will make sure code always works.
+ */
+const disabledScenarioTest = (scenario, testName) =>
+  disabledScenario(scenario) ||
+  _.get(scenario, ['disabled', testName.replace(path.extname(testName), '')])
```
taratatach: Why the extra pair of quotes at the end?

test/support/helpers/scenarios.js
```diff
@@ -83,20 +79,34 @@ if (module.exports.scenarios.length === 0) {
...
-    } else if (name.endsWith(disabledExtension)) {
-      return 'disabled case'
+    } else {
+      return disabledScenarioTest(scenario, `local/${name}`)
     }
   }
   return eventFiles.map(f => {
     const name = path.basename(f)
-    const disabled = scenario.disabled || disabledEventsFile(name)
+    const disabled = disabledScenario(scenario) || disabledEventsFile(name)
```
taratatach: It seems redundant to call `disabledScenario(scenario)` here since it will be called by `disabledScenarioTest()` in `disabledEventsFile()`.

test/support/helpers/scenarios.js
```diff
@@ -116,13 +126,13 @@ module.exports.loadAtomCaptures = scenario => {
...
-    } else if (name.endsWith(disabledExtension)) {
-      return 'disabled case'
+    } else {
+      return disabledScenarioTest(scenario, `atom/${name}`)
     }
   }
   return eventFiles.map(f => {
     const name = path.basename(f)
-    const disabled = scenario.disabled || disabledEventsFile(name)
+    const disabled = disabledScenario(scenario) || disabledEventsFile(name)
```
taratatach: Same here.

test/support/helpers/scenarios.js
```diff
@@ -140,7 +150,7 @@ module.exports.loadRemoteChangesFiles = scenario => {
 
   return glob.sync(pattern).map(f => {
     const name = path.basename(f)
-    const disabled = scenario.disabled || f.endsWith(disabledExtension)
+    const disabled = disabledScenarioTest(scenario, 'remote')
```
taratatach: Shouldn't we use `shouldSkipRemote()` here?

test/support/helpers/scenarios.js
sebn: Right, but it won't be called first in `disabledEventsFile()`.
Which means the reason why it is disabled won't be the same.
I didn't wanted to change behavior here.
But all those `disabled*()` functions will definitely need refactoring.

test/support/helpers/scenarios.js
sebn: We would have a dependency problem.
Both `test/support/helpers/scenario` & `test/scenarios/run` need refactoring, but this is out of the scope of this PR IMHO.

test/support/helpers/scenarios.js
sebn: Because otherwise, `undefined` would be coerced to a string, ending up looking for `atom/linuxundefined` for example?

test/support/helpers/scenarios.js
sebn: Same as above 😉

PR #1596, opened 2019-05-22
core/local/steps/overwrite: Detect overwriting creations

taratatach:
When LibreOffice Calc saves a new version of a file, it does so
  through a temporary file and replaces the old file with a new one.
  We should detect those and mark the newly created file as an
  overwriting change.

- [x] PR is not too big


test/scenarios/update_file_via_tmp_file/scenario.js
```diff
@@ -0,0 +1,22 @@
+/* @flow */
+
+/*:: import type { Scenario } from '..' */
+
+module.exports = ({
+  side: 'local',
+  init: [{ ino: 1, path: 'file', content: 'initial content' }],
+  actions: [
+    { type: 'mv', src: 'file', dst: 'file.special-tmp' },
```
sebn: Maybe rename file to match the specific case, .e.g. `file.ods.osl-tmp`?
And rename scenario to something like `ods_update_through_osl_tmp_file`?
Possibly explaining the case, either as a comment inside the `scenario.js`, or in a separate `README.md`?

test/scenarios/update_file_via_tmp_file/scenario.js
```diff
@@ -0,0 +1,22 @@
...
+    { type: 'create_file', path: 'file' },
+    { type: 'update_file', path: 'file', content: 'updated content #1' },
+    { type: 'delete', path: 'file.special-tmp' },
+    { type: 'update_file', path: 'file', content: 'updated content #2' }
```
sebn: We should probably make sure we don't break a simple file replacement case, aka `deleted file + created file`.

core/local/steps/overwriting_move.js
```diff
@@ -96,6 +96,29 @@ const ignoreDeletedBeforeOverwritingMove = (event, state) => {
   }
 }
 
+/** Possibly change event action to 'ignored'.
+ *
+ * In case it is a deleted event preceding a replacement.
+ * We expect the Merge step to be able to merge the created event as is, even
```
sebn: `s/Merge step/Merge/`?
`s/as is/as a file update/`?

sebn:
- [ ] `test/scenarios/replace_file_with_file/atom/win32.json`
- [ ] `test/scenarios/replace_file_with_file/local/darwin.json` (disabled in case it doesn't work)
- [ ] `test/scenarios/ods_update_through_osl_tmp_file/local/darwin.json` (idem)

PR #1607, opened 2019-06-04
core: Skip remote change with missing parent in Pouch

sebn:
Trying to merge it again will fail anyway until some new remote change
is made.
    
So the sync stops "blinking" and slowing done because of successive
retries from the failed seq.

---

- [x] PR is not too big


test/unit/helpers_merge.js
```diff
@@ -10,68 +10,76 @@ const pouchHelpers = require('../support/helpers/pouch')
...
-      }
-      let child = {
-        _id: 'exists/child',
-        docType: 'folder'
-      }
-      await this.pouch.db.put(doc)
-      await this.merge.ensureParentExistAsync(this.side, child)
-    })
+        it('works when in the root folder', async function() {
+          await this.merge.ensureParentExistAsync(this.side, { _id: 'foo' })
```
sebn: Pass side directly?

sebn: (making `beforeEach` useless?)

test/unit/remote/watcher.js
```diff
@@ -198,6 +202,12 @@ describe('RemoteWatcher', function() {
           .should.be.rejectedWith(new RegExp(remoteDocs[0]._id))
       })
 
+      it('ignores MergeMissingParentError', async function() {
+        // $FlowFixMe
+        err = new MergeMissingParentError()
```
sebn: Maybe find a way not to use `let` + a `FloxFixMe` annotation?

PR #1608, opened 2019-06-04
gui/updater: Increase UPDATE_CHECK_TIMEOUT

sebn:
So users with slow connection have more chances to start downloading the
update before it is skipped.

---

- [x] PR is not too big

### DIFF
--- a/gui/js/updater.window.js
+++ b/gui/js/updater.window.js
@@ -8,7 +8,20 @@ const log = require('../../core/app').logger({
   component: 'GUI:autoupdater'
 })
 
-const UPDATE_CHECK_TIMEOUT = 5000
+/** The delay starting from the update info request after which it is skipped.
+ *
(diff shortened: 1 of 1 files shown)


gui/js/updater.window.js
```diff
@@ -35,9 +48,9 @@ module.exports = class UpdaterWM extends WindowManager {
     autoUpdater.autoDownload = false
     autoUpdater.on('update-available', info => {
       this.clearTimeoutIfAny()
-      log.info({ update: info, skipped: this.skipped }, 'Update available')
       // Make sure UI don't show up in front of onboarding after timeout
       if (!this.skipped) {
+        log.info({ update: info, skipped: this.skipped }, 'Update available')
```
taratatach: Is there a point in logging the `skipped` value now? And wasn't the point of this log line to know when we were skipping updates that were available?

Maybe we should still log something when we have an available update but skip it.

PR #1615, opened 2019-06-10
core: Ignore "System Volume Information/" by default

sebn:
- We better not touch it on Windows
- We may not have the necessary permissions anyway

---

- [x] PR is not too big


sebn:
@taratatach Could you test it with your existing setup?

PR #1617, opened 2019-06-11
gui/js/updater: Improve error handling

sebn:
- Extract #skipUpdate() from #checkForUpdate() so we can reuse it on
  error, not only on timeout.
- Ignore error after skipped/timeout
- Retry on error after 1s
- Skip on ENOENT error (dev-app-update.yml)
- Remove stuff already logged by autoUpdater.logger
- Make skipping reason clearer

---

- [x] PR is not too big

### DIFF
--- a/gui/js/updater.window.js
+++ b/gui/js/updater.window.js
@@ -1,3 +1,4 @@
+const Promise = require('bluebird')
 const WindowManager = require('./window_manager')
 const { autoUpdater } = require('electron-updater')
 const { translate } = require('./i18n')
@@ -22,6 +23,8 @@ const log = require('../../core/app').logger({
  * available or unavailable update answer anyway.
  */
 const UPDATE_CHECK_TIMEOUT = 10000
+
+const UPDATE_RETRY_DELAY = 1000
 
 module.exports = class UpdaterWM extends WindowManager {
   windowOptions() {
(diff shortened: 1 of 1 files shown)


gui/js/updater.window.js
```diff
@@ -116,23 +123,23 @@ module.exports = class UpdaterWM extends WindowManager {
   }
 
   checkForUpdates() {
-    log.info('Looking for updates...')
     this.timeout = setTimeout(() => {
```
taratatach: What happens when creating new timeouts on each call?

gui/js/updater.window.js
```diff
@@ -75,11 +78,15 @@ module.exports = class UpdaterWM extends WindowManager {
...
-      this.clearTimeoutIfAny()
-      this.afterUpToDate()
+    autoUpdater.on('error', async err => {
+      if (this.skipped) {
+        return
+      } else if (err.code === 'ENOENT') {
+        this.skipUpdate('assuming development environment')
+      } else {
+        await Promise.delay(UPDATE_RETRY_DELAY)
+        autoUpdater.checkForUpdates()
```
taratatach: Shouldn't we still clear the timeout here?

gui/js/updater.window.js
sebn: It would be a pain, but I don't think the current PR changes anything regarding this.

gui/js/updater.window.js
sebn: Wouldn't this make for an infinite update check?

sebn:
@taratatach Please note `autoUpdater.checkForUpdates()` and `this.checkForUpdates()` are not the same, the former is third-party code, while the latter is the only one setting up the update check timeout.

gui/js/updater.window.js
taratatach: I was confused. I took this call to `autoUpdate.checkForUpdates()` for one to `this.checkForUpdates()`…
We don't need to clear anything here.

gui/js/updater.window.js
taratatach: Nevermind. I thought we were calling this method multiple times.

PR #1619, opened 2019-06-12
gui: Synchronization incomplete as catch all status

taratatach:
Stating to the user the synchronisation is impossible when we get an
  error that is not specific enough for us to display a more compelling
  message can be misleading and overly alarming.

  This status can indeed be displayed in situations where we
  successfully merged and synced some changes but not all.
  We prefer to state the synchronisation was incomplete.

- [x] PR is not too big


gui/locales/fr.json
```diff
@@ -58,6 +58,7 @@
   "Dashboard Recent activities": "Activités récentes",
   "Dashboard Show more files": "Montrer plus de fichiers",
   "Dashboard Synchronization impossible": "Synchronisation impossible",
+  "Dashboard Synchronization incomplete": "Synchronization incomplète",
```
tpetitberghien: pas de Z en français : "Synchronisation"

taratatach: Ah oui bien vu ! Merci :)

PR #1622, opened 2019-06-13
local/atom/filter_ignored: Handle renamed paths

taratatach:
Until now, the `filterIgnored` module was ignoring events based on
  their `path` attribute only.
  However, `renamed` events also have an `oldPath` attribute. Movements
  can happen from or to an ignored location and this should be handled
  as follows:
  - ignored to not ignored: event should be considered as a `created`
    event at the not ignored destination path, the `path` attribute
  - not ignored to ignored: event should be considered as a `deleted`
    event at the not ignored source path, the `oldPath` attribute
  - not ignored to not ignored: event is untouched
  - ignored to ignored: event is completely ignored

- [x] PR is not too big

### DIFF
--- a/core/local/atom/filter_ignored.js
+++ b/core/local/atom/filter_ignored.js
@@ -15,6 +15,7 @@ const log = logger({
 })
 
 module.exports = {
+  STEP_NAME,
   loop
 }
 
--- a/core/local/atom/filter_ignored.js
+++ b/core/local/atom/filter_ignored.js
@@ -1,10 +1,13 @@
 /* @flow */
 
+const _ = require('lodash')
+
 const logger = require('../../logger')
+const metadata = require('../../metadata')
 
 /*::
 import type Channel from './channel'
-import type { AtomEvent } from './event'
+import type { AtomEvent, EventKind } from './event'
 import type { Ignore } from '../../ignore'
 */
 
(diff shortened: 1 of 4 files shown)


test/unit/local/atom/filter_ignored.js
```diff
@@ -29,142 +33,186 @@ onPlatforms(['linux', 'win32'], () => {
...
+      .path('tmp/isIgnored.txt')
+      .kind('file')
+      .build()
+    const notIgnoredRenamedEvent = builders
+      .event()
+      .action('renamed')
+      .oldPath('notIgnored.txt')
+      .path('stillNotIgnored.txt')
+      .kind('file')
+      .build()
```
sebn: Wouldn't separate tests be easier to read?

test/unit/local/atom/filter_ignored.js
```diff
@@ -29,142 +33,186 @@ onPlatforms(['linux', 'win32'], () => {
...
       beforeEach(() => {
-        channel.push([ignoredEvents[0], notIgnoredEvents[0]])
-        channel.push([ignoredEvents[1], notIgnoredEvents[1]])
+        channel.push(_.cloneDeep([notIgnoredScanEvent, ignoredScanEvent]))
+        channel.push(_.cloneDeep([notIgnoredCreatedEvent, ignoredCreatedEvent]))
+        channel.push(_.cloneDeep([notIgnoredDeletedEvent, ignoredDeletedEvent]))
+        channel.push(
+          _.cloneDeep([ignoredRenamedSrcEvent, ignoredRenamedDstEvent])
+        )
+        channel.push(_.cloneDeep([notIgnoredRenamedEvent, INITIAL_SCAN_DONE]))
```
sebn: Do we really need to test all those cases again when the *filter_ignore* is not aggregating batches in any way?

core/local/atom/filter_ignored.js
```diff
@@ -28,24 +31,68 @@ function loop(
...
+        } else if (isPathIgnored) {
+          const deletedEvent = movedToIgnoredPath(event)
+          batch.push(deletedEvent)
+        } else if (isOldPathIgnored) {
+          const createdEvent = movedFromIgnoredPath(event)
+          batch.push(createdEvent)
+        } else {
+          batch.push(event)
+        }
+      } else if (isPathIgnored) {
```
sebn: Could probably be DRYed by including each predicate + corresponding (possibly made-up) event in a separate function then using `||` and pushing resulting not ignored event if any / logging ignored event otherwise.

test/unit/local/atom/filter_ignored.js
taratatach: I just wanted to make sure we have tests for both 1 batch and multi-batches. Besides it was already done that way.
But it might be overkill indeed.

taratatach: It's just we know some steps can have different behaviors depending on the batches they're given and event if this one does not __at the moment__ I'd feel safer it this was tested.

core/local/atom/filter_ignored.js
taratatach: I thought I was DRYing up the code actually :D. I don't really see what you're thinking about though.
Let's discuss it directly.

test/unit/local/atom/filter_ignored.js
sebn: (was already like this => out of the scope of this PR)

core/local/atom/filter_ignored.js
sebn: No blocker :ship:

PR #1623, opened 2019-06-13
core/local/atom/add_checksum: Log only computed checksums

taratatach:
It can be hard to know when reading logs if a checksum was actually
  computed or not since we log before testing for the existence of a
  checksum on the given event.
  This would help figure out when we compute checksums when we don't
  want to (i.e. we see computed checksums logs while no changes were
  made on the file).
  This should also reduce the size of the log files.

- [x] PR is not too big

### DIFF
--- a/core/local/atom/add_checksum.js
+++ b/core/local/atom/add_checksum.js
@@ -10,10 +10,12 @@ const STEP_NAME = 'addChecksum'
 const log = logger({
   component: `atom/${STEP_NAME}`
 })
+const contentActions = new Set(['created', 'modified', 'renamed', 'scan'])
 
 /*::
 import type Channel from './channel'
 import type { Checksumer } from '../checksumer'
+import type { AtomEvent } from './event'
 */
 
 module.exports = {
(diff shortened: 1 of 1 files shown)


core/local/atom/add_checksum.js
```diff
@@ -60,3 +59,13 @@ function loop(
...
   })
 }
+
+function isFileWithContent(event /*: AtomEvent */) /*: boolean %checks */ {
+  return (
+    (event.action === 'created' ||
+      event.action === 'modified' ||
+      event.action === 'scan' ||
+      event.action === 'renamed') &&
+    event.kind === 'file'
```
sebn: Maybe use a `Set`?

```js
const contentActions = new Set(['created', 'modified', 'renamed', 'scan'])
const isFileWithContent = ({ action, kind }) => kind === 'file' && contentActions.has(action)
```

taratatach: This method is a Flow predicate. Its body needs to be a single return expression for Flow to understand it.

PR #1629, opened 2019-06-18
core/local/atom/producer: Do not #scan() ignored files/dirs

sebn:
So we don't try to scan:

- `.system-tmp-cozy-drive/` contents
- Windows system files/dirs

---

- [x] PR is not too big

### DIFF
--- a/core/local/atom/producer.js
+++ b/core/local/atom/producer.js
@@ -22,6 +22,12 @@ const log = logger({
   component: 'atom/Producer'
 })
 
+const isIgnored = ({ path, kind }, ignore) =>
+  ignore.isIgnored({
+    relativePath: path,
+    isFolder: kind === 'directory'
+  })
+
 // This class is a producer: it watches the filesystem and the events are
 // created here.
 //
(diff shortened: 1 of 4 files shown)


test/unit/local/atom/producer.js
```diff
@@ -17,17 +18,42 @@ onPlatforms(['linux', 'win32'], () => {
...
+
+        beforeEach(async () => {
+          ignore = new Ignore([ignoredDir])
+          producer = new Producer({ syncPath, ignore })
+
+          await syncDir.makeTree([
+            `${ignoredDir}/`,
+            `${ignoredDir}/subdir/`,
+            `${ignoredDir}/subdir/file`,
+            `${notIgnoredFile}/`
```
taratatach: There shouldn't be any `/` at the end of this path, right?

core/local/atom/producer.js
```diff
@@ -115,7 +124,9 @@ module.exports = class Producer {
         }
         if (stats) scanEvent.stats = stats
         if (incomplete) scanEvent.incomplete = incomplete
-        entries.push(scanEvent)
+        if (!isIgnored(scanEvent, this.ignore)) {
```
taratatach: I wonder if we could have situations where we want to ignore some directories for which we can't get the `stats` and thus won't match against the rule.

sebn: Wouldn't the issue be the same in the `add_infos` step anyway?

sebn:
Rebased on master + fixed trailing `/`.

core/local/atom/producer.js
taratatach: Probably. This would be a separate issue then

PR #1633, opened 2019-06-18
[WIP] Code signing playground

sebn:


sebn:
This PR was only used to fix code signing. It is not of any use now. Closing.

PR #1635, opened 2019-06-19
core/metadata: No ancestors duplication in conflict names

taratatach:
If a file, already in conflict, should be renamed because of a
  conflicting document and has ancestors, then those ancestors should
  not be duplicated in the generated path.

- [x] PR is not too big

### DIFF
--- a/core/metadata.js
+++ b/core/metadata.js
@@ -607,7 +607,7 @@ function createConflictingDoc(doc /*: Metadata */) /*: Metadata */ {
   let ext = path.extname(doc.path)
   let dir = path.dirname(doc.path)
   let base = path.basename(doc.path, ext)
-  const previousConflictingName = conflictRegExp('(.*)').exec(doc.path)
+  const previousConflictingName = conflictRegExp('(.*)').exec(base)
   const filename = previousConflictingName ? previousConflictingName[1] : base
   // 180 is an arbitrary limit to avoid having files with too long names
   const notToLongFilename = filename.slice(0, 180)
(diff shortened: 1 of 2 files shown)


core/metadata.js
```diff
@@ -606,12 +606,12 @@ function createConflictingDoc(doc /*: Metadata */) /*: Metadata */ {
...
-  const filename = previousConflictingName ? previousConflictingName[1] : base
-  // 180 is an arbitrary limit to avoid having files with too long names
-  const notToLongFilename = filename.slice(0, 180)
-  dst.path = `${path.join(dir, notToLongFilename)}-conflict-${date}${ext}`
+  let basename = path.basename(doc.path)
+  const previousConflictingName = conflictRegExp('(.*)').exec(basename)
+  const filename = previousConflictingName
+    ? previousConflictingName[1].slice(0, 180)
+    : path.basename(basename, ext).slice(0, 180)
+  dst.path = `${path.join(dir, filename)}-conflict-${date}${ext}`
```
sebn: I believe the main issue is that I missed the `dir` + `filename` joining here. I don't think applying the RegExp to the whole `path` was an issue.

So wouldn't just removing `path.join(dir, ...)` in the old code work and prevent duplicating the slicing logic? E.g.:

```js
dst.path = `${filename}-conflict-${date}${ext}`
```

sebn: (it would also make the `let dir = ...` line useless)

sebn: Oh, but you're right the slicing should not be done on the whole path...

sebn: And looks like you also fixed cases like `dir-conflict-1970-01-01T13_37_00.666Z/file` where some ancestor directory has a conflict suffix but not the actually conflicting file...

sebn: So :+1: in the end, I would just keep the separate `notTooLongFilename` (adding the missing `o`) to prevent duplication.

sebn: (no blocker)

taratatach: I factorised the call to `slice()`

PR #1637, opened 2019-06-20
core/merge: Transform child move if file was created on other side

taratatach:
If a file is added to a directory on one side while the directory is
moved on the other side, we should not try to apply the move on the
child file but transform it instead so it is interpreted as a file
addition at the new location (i.e. in the directory's new path).
This is allowing us to merge both changes during the initial scan
where the client applies changes coming from both sides before
synchronizing them.

- [x] PR is not too big

### DIFF
--- a/core/merge.js
+++ b/core/merge.js
@@ -481,10 +481,10 @@ class Merge {
       let dst = _.cloneDeep(doc)
       dst._id = makeDestinationID(doc)
       dst.path = doc.path.replace(was.path, folder.path)
-      if (src.sides && src.sides[side] && !src.sides[otherSide(side)]) {
-        metadata.markAsUnsyncable(side, src)
-        metadata.markAsNew(dst)
-        metadata.markSide(side, dst)
+
+      const singleSide = metadata.detectSingleSide(src)
+      if (singleSide) {
+        move.convertToDestinationAddition(singleSide, src, dst)
       } else {
         move.child(side, src, dst)
       }
--- a/core/move.js
+++ b/core/move.js
@@ -17,6 +17,7 @@ import type { SideName } from './side'
 
 module.exports = move
 move.child = child
+move.convertToDestinationAddition = convertToDestinationAddition
 
 // Modify the given src/dst docs so they can be merged then moved accordingly
 // during sync.
@@ -48,4 +49,19 @@ function move(side /*: SideName */, src /*: Metadata */, dst /*: Metadata */) {
 function child(side /*: SideName */, src /*: Metadata */, dst /*: Metadata */) {
   src.childMove = true
   move(side, src, dst)
+}
+
+function convertToDestinationAddition(
+  side /*: SideName */,
+  src /*: Metadata */,
+  dst /*: Metadata */
+) {
+  // Delete source
+  if (src.moveTo) delete src.moveTo
+  metadata.markAsUnsyncable(side, src)
+
+  // Create destination
+  if (dst.moveFrom) delete dst.moveFrom
+  metadata.markAsNew(dst)
+  metadata.markSide(side, dst)
 }
--- a/test/integration/move.js
+++ b/test/integration/move.js
@@ -366,6 +366,50 @@ describe('Move', () => {
           ]
         })
       })
+
+      it('on initial scan with parent move', async () => {
+        should(await helpers.local.tree()).deepEqual([
+          'parent/',
+          'parent/dst/',
+          'parent/src/',
+          'parent/src/dir/',
+          'parent/src/dir/empty-subdir/',
+          'parent/src/dir/file',
+          'parent/src/dir/subdir/',
+          'parent/src/dir/subdir/file'
+        ])
+        const was = await pouch.byRemoteIdAsync(dir._id)
+        await helpers._remote.moveFolderAsync(
+          {
+            ...was,
+            path: path.normalize('parent/dst/dir')
+          },
+          was
+        )
+        await helpers.pullAndSyncAll()
+        should(await helpers.trees('metadata', 'remote')).deepEqual({
+          remote: [
+            'parent/',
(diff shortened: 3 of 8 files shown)


core/merge.js
```diff
@@ -484,6 +484,10 @@ class Merge {
         metadata.markAsUnsyncable(side, src)
         metadata.markAsNew(dst)
         metadata.markSide(side, dst)
+      } else if (src.sides && !src.sides[side] && src.sides[otherSide(side)]) {
+        metadata.markAsUnsyncable(otherSide(side), src)
+        metadata.markAsNew(dst)
+        metadata.markSide(otherSide(side), dst)
```
sebn: Looks similar to the other branch...

```js
const singleSide = detectSingleSide(src)
if (singleSide) {
  convertToDestinationAddition(src, dst, otherSide(singleSide))
  // or maybe reverse args, same as move* methods? e.g.
  convertToDestinationAddition(otherSide(singleSide), dst, src)
}

const sideNames = new Set(['local', 'remote'])

const detectSingleSide = doc => {
  if (doc.sides) {
    for (const sideName of sideNames) {
      if (doc.sides[sideName] && !doc.sides[otherSide(sideName)]) {
        return sideName
      }
    }
  }
}

const convertToDestinationAddition = (src, dst, side) => {
  metadata.markAsUnsyncable(otherSide(side), src)
  metadata.markAsNew(dst)
  metadata.markSide(side, dst)
}
```

I wonder whether renaming `markAsUnsyncable()` to `forget()` would be clearer. Do we even need to pass it a side? Out of the scope of this PR anyway.

test/integration/move.js
```diff
@@ -366,6 +366,50 @@ describe('Move', () => {
           ]
         })
       })
+
+      it('on initial scan with parent move', async () => {
```
sebn: weird test description :wink:

taratatach: What's the weird part? I would prefer to have the description follow an `it` word but it's not the way other test descriptions have been written in this file :/

core/merge.js
```diff
@@ -792,6 +792,24 @@ const needsFileidMigration = (
   fileid /*: ?string */
 ) /*: boolean %checks */ => existing.fileid == null && fileid != null
 
+const sideNames = new Set(['local', 'remote'])
+
+const detectSingleSide = doc => {
```
sebn: Maybe move to `core/metadata`?

sebn: Or even better rebase on #1674 and move to `./core/side`? :stuck_out_tongue:

taratatach: It takes a document as single argument. It's probably best in `core/metadata`.

taratatach: However, the `sideNames` Set is probably best in `core/side` :grin:

core/merge.js
```diff
@@ -792,6 +792,24 @@ const needsFileidMigration = (
...
+  if (doc.sides) {
+    for (const sideName of sideNames) {
+      if (doc.sides[sideName] && !doc.sides[otherSide(sideName)]) {
+        return sideName
+      }
+    }
+  }
+}
+
+const convertToDestinationAddition = (side, dst, src) => {
```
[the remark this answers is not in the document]
taratatach: I'll think about it. Maybe do both?

core/merge.js
sebn: Yeah, my bad, I was thinking about `sideNames` :+1:

core/merge.js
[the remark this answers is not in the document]
sebn: Doing both may look strange since move default export is already a function with another function as property:

```js
move()
move.child()
move.convertMoveToDestinationAddition() // move x2
move.convertToDestinationAddition() // better?
```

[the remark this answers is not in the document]
taratatach: ah right. I thought `move()` was exported as a property.
The second one looks good to me.

core/move.js
```diff
@@ -40,8 +41,8 @@ function move(side /*: SideName */, src /*: Metadata */, dst /*: Metadata */) {
   if (!dst.overwrite) {
     delete dst._rev
   }
-  delete dst.sides
-  metadata.markSide(side, dst)
+  reinitializeSides(dst, src)
```
sebn: I would have expected this to break stuff, I'm surprised it doesn't...

sebn: Not sure why it is needed either... Wouldn't having `{[side]: 1}` already ensure the addition to the destination is synced?

taratatach: Tests are indeed passing without this commit. I removed it and will keep it in another branch just in case I figure out a situation where it's necessary.

core/move.js
```diff
@@ -50,3 +51,18 @@ function child(side /*: SideName */, src /*: Metadata */, dst /*: Metadata */) {
...
   move(side, src, dst)
 }
+
+function convertToDestinationAddition(
+  side /*: SideName */,
+  src /*: Metadata */,
+  dst /*: Metadata */
+) {
+  // Delete source
+  if (src.moveTo) delete src.moveTo
```
taratatach: @nono 	I added the 2 `delete` calls after writing unit tests for this method. If you're ok with those changes, I'll merge the PR.

nono: OK for the second delete, but for the first delete, I don't see in which case it happens. Can you enlighten me?

taratatach: When we call `move(side, src, dst)`, we add a `moveTo` attribute on `src` and a `moveFrom` attribute to `dst`.
So I'm removing both to make sure nothing's left from the move.

core/move.js
taratatach: I figured out why it's needed :).

For successive child moves. If we don't keep both existing sides in child moves and later want to move those children, we'll end up detecting only one side on them and convert those moves to additions later resulting in sync errors because their remote revisions would not have been updated.

PR #1650, opened 2019-06-26
core/remote/cozy: Expect cozy-client-js FetchErrors

taratatach:
Based on #1621 

The `cozy-client-js` lib defines its own `FetchError` objects when
throwing a network error.
This means it is not an instance of the `electron-fetch` defined
`FetchError` type and that calls to `instanceof` won't qualify errors
thrown by calls to `cozy-client-js` methods as `Fetch` errors.

However, we can detect all `Fetch` errors in
`.handleCommonCozyErrors()` by checking if their name is
"FetchError".

- [x] PR is not too big

### DIFF
--- a/core/remote/cozy.js
+++ b/core/remote/cozy.js
@@ -2,7 +2,6 @@
 
 const autoBind = require('auto-bind')
 const CozyClient = require('cozy-client-js').Client
-const { FetchError } = require('electron-fetch')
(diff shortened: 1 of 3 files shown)


sebn:
To be rebased on https://github.com/cozy-labs/cozy-desktop/pull/1653

PR #1647, opened 2019-06-25
core/path_restrictions: Allow colon (`:`) on macOS

sebn:
So it allows slash (`/`) from a macOS user standpoint.

Please make sure the following boxes are checked:

- [x] PR is not too big
- [x] it improves UX & DX in some way
- [ ] it includes unit tests matching the implementation changes
- [x] it includes scenarios matching a new behaviour or has been manually tested
- [x] it includes relevant documentation

PR #1651, opened 2019-06-26
core/metadata: Dissociate adding and replacing conflict suffixes

taratatach:
The body of `createConflictingDoc()` was getting very complex and
error prone (e.g. keeping the previous conflit's milliseconds part
when a document has no extension).

We have 2 very different situations:
- either the document already has a conflict suffix
- or it does not

In the first case, we want to replace the previous suffix with a new
one while we want to add a conflict suffix in the second case, after
the file's basename but before its extension.
We can then create one function for each case, returning a new version
of the file's path and return a copy of the document with new `path`
and `_id` attributes.

- [x] PR is not too big

### DIFF
--- a/test/unit/metadata.js
+++ b/test/unit/metadata.js
@@ -27,7 +27,7 @@ const {
   upToDate,
   outOfDateSide,
   createConflictingDoc,
-  conflictRegExp
+  CONFLICT_REGEXP
 } = metadata
 const { Ignore } = require('../../core/ignore')
 const { FILES_DOCTYPE } = require('../../core/remote/constants')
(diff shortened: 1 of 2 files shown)


test/unit/metadata.js
```diff
@@ -1013,20 +1013,19 @@ describe('metadata', function() {
...
-        .and.match(pathRegExp)
+        .and.match(CONFLICT_PATTERN)
     })
     it('should get the correct _id', () => {
       const doc = {
         path: 'docname'
       }
       const newDoc = createConflictingDoc(doc)
-      const pathRegExp = new RegExp(conflictRegExp(doc.path).source, 'i')
+      const idRegExp = new RegExp(CONFLICT_PATTERN.source, 'i')
```
sebn: - [ ] The refactored unit tests don't check the part (dirname + basename) before the conflict suffix anymore since we dropped the `conflictRegExp()` function. Which means all of them will still pass even when returning only the conflict suffix. But maybe adding some `.and.startWith(...)` assertions would be good enough, without having to reintroduce the old function.

sebn:
- [ ] 2 tests broken because of [me merging macOS colon while electron-mocha ignored `--forbid-only` option was hiding them](https://github.com/cozy-labs/cozy-desktop/pull/1647#issuecomment-506265758)
- [ ] 1 conflict test to fix

PR #1661, opened 2019-07-01
core/local: Re-use untouched files' checksum

taratatach:
In order to limit the number of checksum computations during the
initial scan, we try to reuse the existing PouchDB record's local
`md5sum` if the event's update time matches the record's local
`updated_at`.

We compare the event's `mtime` field because we care only about the
file content updates to trigger checksum computations.
Also, we now compare it with the recently added local `updated_at`
which will now contain the latest local `mtime` value instead of
either the remote `updated_at` value or the max between the local
`mtime` and `ctime` that can be stored in the record's main
`updated_at` attribute.
Finally, we re-use the local `md5sum` because the main `md5sum`
attribute can have changed after a remote update was merged but not
applied on the local filesystem.

Since we're changing the way we compute the local `updated_at`, all
existing files' records will be updated during the first launch
including this commit.
This can take a while but all later launches will benefit from it.

- [x] PR is not too big

### DIFF
--- a/core/local/atom/initial_diff.js
+++ b/core/local/atom/initial_diff.js
@@ -177,7 +177,7 @@ async function initialDiff(
           }
         } else if (foundUntouchedFile(event, was)) {
           _.set(event, [STEP_NAME, 'md5sumReusedFrom'], was.path)
-          event.md5sum = was.md5sum
+          event.md5sum = was.local.md5sum
         }
       }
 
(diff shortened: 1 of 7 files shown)


core/local/atom/initial_diff.js
```diff
@@ -320,20 +320,20 @@ function fixPathsAfterParentMove(renamedEvents, event) {
...
 function foundUntouchedFile(event, was) /*: boolean %checks */ {
   return (
-    was != null &&
-    was.md5sum != null &&
     event.kind === 'file' &&
-    eventUpdateTime(event) === docUpdateTime(was)
+    was != null &&
+    was.local != null &&
+    was.local.md5sum != null &&
+    contentUpdateTime(event) <= docUpdateTime(was.local)
```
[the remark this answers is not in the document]
taratatach: Not from the top of my mind. It probably seemed logical at the time not to recompute the checksum of an older file.
I'll think about it a bit to see if I can come up with a scenario that would need the less than part.

core/metadata.js
```diff
@@ -814,6 +816,13 @@ function shouldIgnore(
...
-  doc.local = _.pick(newLocal || doc, LOCAL_ATTRIBUTES)
+function updateLocal(doc /*: Metadata */, newLocal /*: ?Object */ = {}) {
+  // Boolean attributes not set in doc when false will not override an existing
+  // truthy value.
+  // This is the case for `executable` and we need to provide a default falsy
+  // value to override the `newLocal` executable value in all cases.
+  doc.local = _.pick(
+    _.defaults(newLocal, { executable: false }, doc),
+    LOCAL_ATTRIBUTES
+  )
```
nono, line 827: Is this change related to this PR? It looks like it is more a bugfix that may be in a separated commit.

taratatach: The use of `_.defaults()` is needed to set the `updated_at` value here : https://github.com/cozy-labs/cozy-desktop/pull/1661/files#diff-b2fbff412cdb1b20ee3b7c797ef5b609R772.

But using this method means the `executable: true` attribute from `doc` can override the absence of the `executable` attribute in `newLocal` so we need to provide the intermediary default `executable:false` value.
So it is related to this PR.

PR #1668, opened 2019-07-04
core: Refactor to ./incompatibilities/platform

sebn:
So we can later introduce new types of incompatibilities.

The `platform` name will make it easier to know who files/dirs are
incompatible with (is it the current platform? the Cozy ? the
synchronization client ?) and also in which direction the checks
are relevant.

Moving it to a subdir will make it easier to group/find all the
incompatibility types.

Using the *incompatibilities*  wording for module/types/functions (vs
the old *issues* suffix in function names & the old module name) better
matches the rest of the codebase (`Metadata#incompatibilities`,
`gui/js/incompatibilitiesmsg`... same name everywhere!).

Also replace the *path* wording with *platform* when relevant, to make
it easier to understand the link between types/functions & the new
module. Renaming conflicting type/function in `./metadata`.

Also move from `function` style to arrow functions, improve a few
comments & make them jsdoc-compatible.

This commit is refactoring only.
New types of incompatibilities will be introduced in subsequent ones.

---

- [ ] PR is not too big


sebn:
@taratatach Early feedback welcome!

test/unit/incompatibilities/platform.js
```diff
@@ -2,19 +2,22 @@
 
 const should = require('should')
 
-const pathRestrictions = require('../../core/path_restrictions')
-const { detectNameIssues, detectPathLengthIssue } = pathRestrictions
-
-describe('path_restrictions', () => {
-  describe('detectNameIssues', () => {
+const pathRestrictions = require('../../../core/incompatibilities/platform')
```
taratatach: `platformIncompatibilities`?

PR #1677, opened 2019-07-12
[WIP] core: Use AtomWatcher on macOS

sebn:
So we can:

- Drop the whole chokidar namespace (sort of)
- Get rid of the 10s delay
- Lower resources consumption

---

- [x] PR is not too big


core/local/atom/producer.js
```diff
@@ -61,12 +62,21 @@ class Producer {
...
   constructor(opts /*: { syncPath: string, ignore: Ignore } */) {
     this.channel = new Channel()
     this.syncPath = opts.syncPath
     this.ignore = opts.ignore
     this.watcher = null
+    if (process.platform === 'darwin') {
+      const timeoutInMs = process.env.NODE_ENV === 'test' ? 1000 : 10000
+      this.macOSBuffer = new LocalEventBuffer(timeoutInMs, batches => {
+        for (const batch of batches) {
+          this.process(batch)
```
sebn: await?

core/local/atom/producer.js
```diff
@@ -90,10 +100,15 @@ class Producer {
...
+    const macOSBuffer = this.macOSBuffer
+    const onEventBatch = macOSBuffer
+      ? macOSBuffer.push.bind(macOSBuffer)
+      : this.process
+
     this.watcher = await watcher.watchPath(
       this.syncPath,
       { recursive: true },
-      this.process
+      onEventBatch
```
sebn: should we regenerate all captures with buffer?

sebn: also is the buffer necessary or not?

PR #1679, opened 2019-07-12
Test sync feedback in remote scenarios

taratatach:
To make sure we don't have bad interactions between the
synchronization of remote changes and the local watcher, we need to
run the local watcher during our remote scenarios.

This involves running the watcher, sending a magic event after we've
run all scenarios actions to signal the end of the test and using real
inodes to make sure the local watcher does not detect and merge inode
changes on init files during its initial scan.

- [x] PR is not too big

### DIFF
--- a/core/local/chokidar/event_buffer.js
+++ b/core/local/chokidar/event_buffer.js
@@ -71,12 +71,12 @@ class EventBuffer /*:: <EventType> */ {
     }
   }
 
-  flush() /*: void */ {
+  async flush() {
     this.clearTimeout()
     if (this.events.length > 0) {
       const flushedEvents = this.events
       this.events = []
-      this.flushed(flushedEvents)
+      return this.flushed(flushedEvents)
     }
   }
 
--- a/test/scenarios/run.js
+++ b/test/scenarios/run.js
@@ -249,6 +249,7 @@ async function runLocalStopped(scenario, helpers) {
 
 async function runRemote(scenario, helpers) {
   if (scenario.init) {
+    const useRealInodes = true
     let relpathFix = _.identity
     if (process.platform === 'win32') {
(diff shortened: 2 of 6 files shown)


test/scenarios/run.js
```diff
@@ -257,18 +258,21 @@ async function runRemote(scenario, helpers) {
...
 
+  await helpers.local.side.watcher.start()
   await helpers.remote.pullChanges()
   // TODO: Don't sync when scenario doesn't have target FS/trash assertions?
-  for (let i = 0; i < scenario.actions.length + 1; i++) {
-    await helpers.syncAll()
-  }
+  await helpers.syncAll()
+  // Wait for all local events to be dispatched
+  await helpers.local.simulateAtomEvents([])
```
sebn: No guarantee that the magic event is not pushed before the FS ones.

taratatach: All tests are green but since this PR is about tests lying I'll add a wait time before sending the magic event.
500 ms should be enough.

core/local/chokidar/event_buffer.js
```diff
@@ -76,7 +76,7 @@ class EventBuffer /*:: <EventType> */ {
     if (this.events.length > 0) {
       const flushedEvents = this.events
       this.events = []
-      this.flushed(flushedEvents)
+      return this.flushed(flushedEvents)
```
nono: Why the `return` here? I'm surprised because it doesn't match the signature: `flush() /*: void */`

taratatach: I need to be able to wait for the flush to complete in the scenarios runner but the result of `this.flushed()` which is an async function was not returned by `flush()`.
So I decided to return it. It has no effect on the production code but I forgot to update the method's signature.

PR #1682, opened 2019-07-17
core/pouch/migrations: Introduce Pouch migrations

taratatach:
We don't control the Pouch's `_rev` attribute set on each document.
When we move a document, the destination will be a new document  in
Pouch. This means its `_rev` will probably start with 1 (it can be
higher if the path has previously existed but it does not matter
here).

Since our mechanism to determine which changes need to be applied on
which side is based on the `_rev` value (i.e. its short rev or
beginning number to be precise) and the `sides` we maintain, we
cannot keep the existing `sides` value since they will probably not be
correlated with the new `_rev` anymore.
This means we're losing information when doing a move.

This loss prevents us from being able to merge some combinations of
changes.
e.g. With the client stopped:
1. add document to a folder locally
2. move that folder remotely
3. launch client to synchronize changes

The file will be moved as a child of the moved directory and its
`sides` attribute will contain `{ remote: 1 }` (i.e. we're losing the
local addition information).
In this situation, we're not able to move the file since it doesn't
exist remotely and we don't have any remote id for it and we won't try
to upload it to the Cozy since we see a remotely created file.

By maintaining our own `rev` attribute, we're able to keep a
correlation between the `rev` and the `sides` values even when moving
a document.
In our previous example, we'll be easily able to mark the file as created
locally in the new directory.

We decided to add this `rev` attribute inside the `sides` attribute
since they work together.
The `MetadataSidesInfos` type looks like this now:
```
{
  rev: number,
  local?: number,
  remote?: number
}
```

Our faked docs coherence was also improved by:
- Improving Metdata builder's `upToDate()` method to take into account
  existing sides (i.e. if doc is based on another one)
- Using builder's helper methods to set sides whenever possible
- Avoiding faking movements when possible (to avoid unnecessary side
  effects when running the move method we're testing)

- [ ] PR is not too big


core/pouch/index.js
```diff
@@ -112,6 +141,16 @@ class Pouch {
     })
   }
 
+  async runMigrations() {
+    log.debug('Running migrations...')
+    for (const migration of migrations) {
+      const result = await migrate(migration, this)
+      const { errors, msg } = migrationLog(migration, result)
+      log.debug({ errors }, msg)
```
nono: Only a debug log if the migration has some errors?

core/pouch/index.js
```diff
@@ -12,19 +12,48 @@ const _ = require('lodash')
...
 })
 
+/** Represents an Error as returned by Pouch when saving a document
+ *
+ * Some errors will not be returned as such since they're treated differently by
+ * Pouch itself.
+ * Examples include:
+ * - 'bad_format', thrown when a doc's _rev attribute's format is bad
+ */
+class PouchError extends Error {
```
nono: [nit] I'd have put `PouchError` in its own file

core/pouch/migrations.js
```diff
@@ -0,0 +1,150 @@
...
+  _id: SCHEMA_DOC_ID,
+  version: SCHEMA_INITIAL_VERSION
+}
+const MIGRATION_RESULT_NOOP = 'MigrationNoop'
+const MIGRATION_RESULT_COMPLETE = 'MigrationComplete'
+const MIGRATION_RESULT_FAILED = 'MigrationFailed'
+
+const migrations /*: Migration[] */ = []
+
+function MIGRATION_NOOP() {
```
nono: I'm not sure it is very idiomatic to call a function with a name all in upper case

core/pouch/migrations.js
```diff
@@ -0,0 +1,150 @@
...
+    }
+    return result
+  }
+}
+
+async function save(
+  docs /*: Metadata[] */,
+  pouch /*: Pouch */
+) /*: Promise<MigrationResult> */ {
+  const migrationResult /* MigrationResult */ = new MIGRATION_NOOP()
```
nono: It feels weird to use `new` with something declared as a function, and not as a class

core/pouch/index.js
taratatach: I'd like to know when migrations are run and their result (noop, failed, complete).
In case there aren't any errors, it will print only one line per migration. This seems small enough to me.

nono: I'm OK with that. My point is that if there is an error, shouldn't we do something like alerting the user or sending the error to sentry?

core/pouch/migrations.js
taratatach: It's idiomatic javascript though. I'm only using classes for very large classes or when Flow doesn't let me use a function as a type.

taratatach: Sorry I misunderstood your question. I've renamed the function to `migrationNoop` and will use it as a simple function.

core/pouch/index.js
taratatach: Errors will be logged but we should probably send them to Sentry indeed.

core/pouch/migrations.js
```diff
@@ -0,0 +1,244 @@
...
+  const date = new Date()
+  const dateString = `${date.getFullYear()}-${date.getMonth() +
+    1}-${date.getDate()}`
+  const originalDBInfo = await originalDB.info()
+  return `${originalDBInfo.db_name}-migration-${dateString}`
+}
+
+async function replicateDB(fromDB /*: PouchDB */, toDB /*: PouchDB */) {
+  return new Promise(async (resolve, reject) => {
+    fromDB.replicate
```
nono: `replicate` doesn't delete the documents in `toDB` that are not in `fromDB`. I think it may cause some issues for migrations that change the `_id` of documents. For example, we may want to play a migration to add a `/` at the start of the `_id` of each doc to fix the issue for files starting with an underscore.

taratatach: There shouldn't be any issues here.
When we create the migration DB, `toDB` is empty.
When we replace the user's DB with the content of the migration DB, we have previously reset the user's DB so `toDB` is empty in this situation as well.

nono: It's probably my mistake, but I don't see where the user's DB is reset

taratatach: It's done in the `migrate` method: https://github.com/cozy-labs/cozy-desktop/pull/1682/files/c5b3d72078270f2201bdf986c5a3da278005c30d#diff-71833bc7fc5def55987ed47e8e557c16R98

PR #1684, opened 2019-07-26
core/local/chokidar/prepare_events: Manage HFS+ re-encoding

taratatach:
When applying remote changes on the local filesystem for paths
containing utf8 characters, the encoding can be automatically and
silently modified by the filesystem itself.
For example, HFS+ volumes will normalize UTF-8 characters using NFD
(c.f. https://en.wikipedia.org/wiki/Unicode_equivalence#Normalization).

When this happens, the local watcher will fire events for paths which
aren't exactly equal to the ones that were merged from the remote
changes, leading to the detection of a `move` which will then be
synchronized with the remote and so on.

To avoid those "leaks" which can create conflicts or at least
extraneous changes on existing documents, we can normalize on the fly
both the stored path and the event's path using NFC so we won't detect
movements when only the encoding has changed.
Also, this means that future paths will always be stored using NFC
encoding while ids will still use NFD.

All other modifications will be merged and synchronized.

- [x] PR is not too big


test/scenarios/create_dir_with_utf8_character/local/darwin_hfs+.json
```diff
@@ -0,0 +1,12 @@
+[
+  {
+    "type": "addDir",
+    "path": "Partages reçus",
```
taratatach: FYI, this path is encoded using NFD while the path for APFS is encoded using NFC.

test/scenarios/create_dir_with_utf8_character/remote/changes.json
```diff
@@ -0,0 +1,13 @@
+[
+  {
+    "_id": "5a4845b78561d1c5970243cbe40012a5",
+    "_rev": "1-58193d0119d3f773f8982101206b5dc6",
+    "created_at": "2019-07-26T14:04:13.704599113Z",
+    "dir_id": "io.cozy.files.root-dir",
+    "name": "Partages reçus",
+    "path": "/Partages reçus",
```
taratatach: The name and path are encoded using NFC as this seems to be the encoding used in SWIFT.

test/scenarios/create_dir_with_utf8_character/scenario.js
```diff
@@ -0,0 +1,13 @@
+/* @flow */
+
+/*:: import type { Scenario } from '..' */
+
+//const save = 'Partages reçus/'
+
+module.exports = ({
+  actions: [{ type: 'mkdir', path: 'Partages reçus' }],
```
taratatach: Encoded with NFC.

test/scenarios/create_dir_with_utf8_character/scenario.js
```diff
@@ -0,0 +1,13 @@
...
+  expected: {
+    tree: ['Partages reçus/'],
```
taratatach: Encoded with NFC but local expectations will be normalized using NFD when tests are run on a HFS+ volume (see https://github.com/cozy-labs/cozy-desktop/pull/1684/files#diff-e535224c0b54370fef0dc6d963d6c27aR304)

taratatach:
I added the documentation inside the code using JSDoc.

PR #1686, opened 2019-07-29
core/metadata: Child conflicts do not replace parent conflict suffix

taratatach:
The regexp matching existing conflict suffixes in document paths was
too open and would match conflicts in the document's parents (e.g.
`parent-conflict-2019-07-29T14_00_01.223Z/document.pdf`).

This would result in replacing the parent's conflict suffix instead
of adding a suffix to the child document's path.

The regexp now ensures via a look-ahead that the conflict suffix being
matched is not followed by a `/`, marking a parent directory.

- [x] PR is not too big

### DIFF
--- a/core/metadata.js
+++ b/core/metadata.js
@@ -63,7 +63,11 @@ const log = logger({
 
 const { platform } = process
 
-const CONFLICT_REGEXP = /-conflict-\d{4}(?:-\d{2}){2}T(?:\d{2}_?){3}\.\d{3}Z/
+const DATE_REGEXP = '\\d{4}(?:-\\d{2}){2}T(?:\\d{2}_?){3}\\.\\d{3}Z'
+const SEPARATOR_REGEXP = `(?!.*\\${path.sep}.*)`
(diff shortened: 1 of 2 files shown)


core/metadata.js
```diff
@@ -63,7 +63,11 @@ const log = logger({
 
 const { platform } = process
 
-const CONFLICT_REGEXP = /-conflict-\d{4}(?:-\d{2}){2}T(?:\d{2}_?){3}\.\d{3}Z/
+const DATE_REGEXP = '\\d{4}(?:-\\d{2}){2}T(?:\\d{2}_?){3}\\.\\d{3}Z'
+const SEPARATOR_REGEXP = `(?!.*\\${path.sep}.*)`
```
nono: I'd have written this as `[^\\${path_sep}]*$`, but both versions should work

taratatach: I remember trying this version (I'm not particularly familiar with the look-ahead) but it wouldn't work for some reason.
It could be a syntax error.

PR #1746, opened 2019-10-24
test/unit/local/atom/dispatch: Harden tests

taratatach:
When running this unit tests file enough (or maybe on resources
constrained device like Travis), we can end up with some tests failing
because the expected built document has a different update time than
the one created by `dispatch`.

This is actually unexpected that those tests would not fail more often
since we build the expected document's update time depends on its
build time while the tested document's update time comes from the
event pushed into the events channel.

By setting the event's update time with the same value than the one
used when building the expected document we can make sure those values
will be the same when comparing the 2 documents.

### DIFF
--- a/test/unit/local/atom/dispatch.js
+++ b/test/unit/local/atom/dispatch.js
@@ -114,6 +114,7 @@ describe('core/local/atom/dispatch.loop()', function() {
 
   context('when channel contains a scan file event', () => {
     const filePath = 'foo'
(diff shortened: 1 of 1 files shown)


nono:
@taratatach what should we do of this PR?

taratatach:
I'll try to rebase it on https://github.com/cozy-labs/cozy-desktop/pull/1783 and merge it.
Thanks for the notice.

PR #1749, opened 2019-10-30
Load docs in bulk in initial scans and migrations

taratatach:
When loading all documents in migrations or the initial scan of either
  local watchers, we end up using a low-level db iterator which seems to
  fail at parsing documents when they're too deeply nested (i.e. lots of
  `moveFrom` inside `moveFrom`, probably resulting from conflict loops).

  PouchDB uses a fallback mechanism when parsing JSON documents fails
  and uses the vuvuzela library instead (see
  https://github.com/nolanlawson/vuvuzela) which does not suffer from
  deeply nested documents parsing errors.

  By using the PouchDB's `allDocs()` method to get all the saved
  documents, we hope not to use the low-level iterator that's failing
  us.

- [ ] PR is not too big

### DIFF
--- a/core/local/atom/initial_diff.js
+++ b/core/local/atom/initial_diff.js
@@ -87,7 +87,7 @@ async function initialState(
   // which files/folders have been deleted, as it is stable even if the
   // file/folder has been moved or renamed
   const byInode /*: Map<number|string, Metadata> */ = new Map()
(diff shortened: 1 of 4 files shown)


taratatach:
Yes, avoiding conflict loops in the goal but, in case the current work on this subject doesn't cover all possible sources of such loops, avoiding crash apps is an improvement for users.

PR #1765, opened 2019-11-13
core/local/atom/add_infos: Try to guess deleted event kind

taratatach:
Sometimes, AtomWatcher won't know the kind of a deleted document and
will trigger a `deleted` event with an `unknown` type.
In those situations, we decided to always force the kind to `file`.

However, if the deleted document is actually a directory, the deletion
won't happen because we'll get an `Mismatch on doctype for doTrash`
error during merge.

It seems more reasonable today to try and fetch the document with the
event's `_id` from Pouch and use its doctype to set the event's kind.
If we don't have any document in Pouch with the event's `_id`, we can
resort back to forcing the kind to `file`.

- [x] PR is not too big

### DIFF
--- a/core/local/atom/add_infos.js
+++ b/core/local/atom/add_infos.js
@@ -22,11 +22,17 @@ const log = logger({
 })
 
 /*::
+import type { Pouch } from '../../pouch'
+import type { Metadata } from '../../metadata'
 import type Channel from './channel'
 import type { AtomEvent } from './event'
 */
 
+const kind = (doc /*: Metadata */) =>
+  doc.docType === 'folder' ? 'directory' : doc.docType
+
 module.exports = {
+  STEP_NAME,
   loop
 }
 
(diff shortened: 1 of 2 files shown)


core/local/atom/add_infos.js
```diff
@@ -65,7 +71,10 @@ function loop(
             // If kind is unknown, we say it's a file arbitrary
             if (event.kind !== 'directory' && event.kind !== 'file') {
               _.set(event, [STEP_NAME, 'kindConvertedFrom'], event.kind)
-              event.kind = 'file'
+
+              const doc =
+                event._id && (await opts.pouch.byIdMaybeAsync(event._id))
```
nono: Can it fail? (it == `await opts.pouch.byIdMaybeAsync(event._id)`)

nono: And, if it can fail, what happens?

taratatach: It can fail if we don't have any document with this `_id` in Pouch.
If it fails, `doc` will be `undefined` and we'll resort back to forcing the event `kind` to `file` as I don't see any other solution for now.

PR #1777, opened 2019-11-22
core/app: Update client version when updating app

taratatach:
The OAuth client stored on the Cozy contains the app's version. This
is used in io.cozy.files cozyMetadata to tell which version of which
app uploaded or updated documents for example.
This could also be useful to know which versions of the app are
installed.

However, the Desktop client would not update its OAuth client after
installing a new version. This means that the version used in metadata
and available in the OAuth client is the version that was installed
when connecting the client to the Cozy.

Updating the app's version in the OAuth client is a first step towards
being able to block some versions of the Desktop client from
connecting to the Cozy by allowing us to delete OAuth clients for old
versions and force a re-connection (with an app update).

We also take this opportunity to fix the device name display in the `Preferences` tab and to use it as Sentry's `server_name` tag's value so we have the same reference for clients across the board (and have a stable `server_name` independent from the user's network configuration).

- [x] PR is not too big

### DIFF
--- a/core/remote/cozy.js
+++ b/core/remote/cozy.js
@@ -127,6 +127,10 @@ class RemoteCozy {
     return this.client.auth.unregisterClient()
   }
 
+  update() /*: Promise<void> */ {
+    return this.client.auth.updateClient()
+  }
+
   diskUsage() /* Promise<*> */ {
     return this.client.settings.diskUsage()
   }
--- a/core/app.js
+++ b/core/app.js
@@ -334,12 +334,40 @@ class App {
   }
 
   setup() {
-    log.info(this.clientInfo(), 'user config')
-    sentry.setup(this.clientInfo())
+    const clientInfo = this.clientInfo()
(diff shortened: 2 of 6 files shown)


core/remote/cozy.js
```diff
@@ -127,6 +127,10 @@ class RemoteCozy {
     return this.client.auth.unregisterClient()
   }
 
+  update() /*: Promise<void> */ {
+    return this.client.auth.updateClient()
```
nono: I'm surprised to not see some params here: `this.client.auth.updateClient(params)`, coming from https://github.com/cozy-labs/cozy-desktop/blob/master/core/remote/registration.js#L61-L88

taratatach: The params are set automatically in `cozy-client-js` https://github.com/cozy/cozy-client-js/blob/master/src/index.js#L220.

This is the same thing for `unregisterClient()`.

PR #1781, opened 2019-12-06
core/remote/watcher: Retry failed changes merge

taratatach:
We can fail to merge some changes for various reasons that are not
  definitive.
  For example, we could encounter very brief network issues or the
  remote changes can be in an order that prevents some of them to be
  merged (i.e. we have merged the creation of their parent folder yet).

  The order issue should not happen in theory since we order the changes
  prior to merging them but we know that in some cases, we fail to
  correctly order all changes, especially when there's a lot of them
  (like during the first synchronisation on a new device).

  We decided to implement a similar retry process than the one done
  during the propagation of merged changes by pushing to the end of the
  list changes which we've failed to merge.
  We hope that later changes will make it possible to merge the earlier
  ones (e.g. we finally get to the parent folder creation change which
  will let us merge its descendants).
  This process should also help with short network issues.

  We stop retrying to merge changes once we've been through the entire
  changes list without merging any changes. In this situation, we know
  the order of changes is not the problem and the network outage might
  longer than expected (e.g. no network at all) and retrying again and
  again would be pointless.

  Once we're done with all our retries, we gather the latest encountered
  errors and pass them along to the watcher.
  For now, we don't change the way we handle them: if there were any
  errors, the remote sequence isn't updated. The next time we fetch the
  remote changes, all merged changes will still be included but the
  watcher will detect they're up-to-date and won't try to merge them
  again.
  The errors are then handled one by one, network errors forcing the
  client into offline mode and the first not network related being
  thrown and stopping the synchronization with a status message and
  notification.

- [ ] PR is not too big

### DIFF
--- a/core/remote/watcher/index.js
+++ b/core/remote/watcher/index.js
@@ -8,7 +8,6 @@ const Promise = require('bluebird')
 const _ = require('lodash')
 
 const metadata = require('../../metadata')
-const { MergeMissingParentError } = require('../../merge')
 const remoteChange = require('../change')
 const { handleCommonCozyErrors } = require('../cozy')
 const { inRemoteTrash } = require('../document')
@@ -98,7 +97,7 @@ class RemoteWatcher {
   }
 
   async watch() {
-    let errors = []
+    let errors /*: Array<{ err: Error, change?: * }> */ = []
 
     try {
       const seq = await this.pouch.getRemoteSeqAsync()
@@ -112,7 +111,7 @@ class RemoteWatcher {
 
       try {
         let target = -1
-        errors = errors.concat(await this.pullMany(docs))
+        errors = await this.pullMany(docs)
         if (errors.length === 0) {
           target = (await this.pouch.db.changes({ limit: 1, descending: true }))
             .last_seq
@@ -125,7 +124,7 @@ class RemoteWatcher {
         log.debug('No more remote changes for now')
       }
     } catch (err) {
-      errors.push(err)
+      errors.push({ err })
     }
 
     for (const err of errors) {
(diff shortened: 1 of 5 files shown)


core/remote/watcher/index.js
```diff
@@ -112,7 +112,7 @@ class RemoteWatcher {
 
       try {
         let target = -1
-        errors = errors.concat(await this.pullMany(docs))
+        errors = await this.pullMany(docs)
```
nono: You are sure about that change? I think that `errors` should be an array

core/remote/watcher/index.js
```diff
@@ -359,131 +358,133 @@ class RemoteWatcher {
...
+  async applyAll(
+    changes /*: Array<RemoteChange> */,
+    errorsCount /*: number */ = 0
+  ) /*: Promise<Array<{ change: RemoteChange, err: Error }>> */ {
+    const errors = await Promise.mapSeries(
+      changes,
+      async change => await this.apply(change)
+    ).filter(err => err)
+    if (errors.length === errorsCount) return errors
+    return this.applyAll(errors.map(error => error.change), errors.length)
```
nono: If we have 2 errors on the first call of `applyError`, and that the 2 errors can be resolved on the second call, we will enter an infinite loop, no? I think we are missing an happy ending path here: `if (errors.length === 0) return []`

core/remote/watcher/index.js
taratatach: `pullMany` is still returning an array but since we don't have any `errors` prior to this call I figured we didn't need to concatenate the result.

core/remote/watcher/index.js
taratatach: I don't think so. Here's my thought process:
1. On first call, `errorsCount` is 0 and `errors.length` is 2
2. Since they differ, we call `applyAll` again
3. On second call, `errorsCount` is 2 and `errors.length` is 0
4. Since they differ, we call `applyAll` again
5. On third call, `errorsCount` is 0 and `errors.length` is 0 (we don't have any changes to merge)
6 Since they're equal, line 369 gets us out of the loop

nono: OK, it works, but I still think it would be clearer if we didn't have an `errorsCount` parameter and line 369 would be:

```js
if (errors.length === 0 || errors.length === changes.length) return errors
```

PR #1783, opened 2019-12-10
core/utils/sentry: Format errors to help grouping

taratatach:
A lot of errors that are sent to Sentry don't have the required
fingerprinting attributes for Sentry to group them together and their
messages differ slightly with file paths, request domains or other
dynamic elements in system calls or network requests.

This generates a lot of noise in our Sentry dashboard, especially for
network requests as those events happen quite often and are not
grouped together at the moment.

To have Sentry group those events, we make sure that every exception
reported will have the appropriate `type` and `message` so they get
grouped together while still retaining the important dynamic parts in
other fields which will get attached to each individual event.

This requires some changes in the bunyan error serializer since the
standard serializer removes a lot of interesting fields from errors
and this would prevent us from doing some pattern matching on them
to format those errors for Sentry.

We start from an object as the standard serializer would return and
add back a few specific attributes that we'll use later in Sentry
errors formatting.

- [x] PR is not too big

### DIFF
--- a/test/unit/utils/sentry.js
+++ b/test/unit/utils/sentry.js
@@ -1,8 +1,32 @@
 /* eslint-env mocha */
 
-require('should')
+const should = require('should')
+const fs = require('fs')
 
 const sentry = require('../../../core/utils/sentry')
+const { FetchError } = require('electron-fetch')
+
+// This class is a copy of the `cozy-client-js` package's `FetchError` as it is
+// not exported and could therefore not be imported.
+class CozyClientFetchError extends Error {
+  constructor(res, reason) {
+    super()
+    if (Error.captureStackTrace) {
+      Error.captureStackTrace(this, this.constructor)
+    }
+    this.name = 'FetchError'
+    this.response = res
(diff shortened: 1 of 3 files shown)


test/unit/utils/sentry.js
```diff
@@ -1,8 +1,31 @@
...
 const sentry = require('../../../core/utils/sentry')
+const { FetchError } = require('electron-fetch')
+
+class CozyClientFetchError extends Error {
+  constructor(res, reason) {
+    super()
+    if (Error.captureStackTrace) {
+      Error.captureStackTrace(this, this.constructor)
+    }
+    // XXX We have to hardcode this because babel doesn't play nice when extending Error
```
nono: Babel? I thought babel was no longer used on cozy-desktop

taratatach: Ah, I just copy/pasted this class from `cozy-client-js` as it's not exported. I'll remove this line and add a comment to explain where the class is coming from.

PR #1793, opened 2020-01-03
core/sync: Create conflict copy on 409 errors

taratatach:
When the stacks responds to a Sync request with a 409 status error, it
means the change we're trying to push is conflicting with an existing
document.
This can only happen when we try to push a new file as updates with
the wrong revision number will result in 412 status errors and
conflicts on directories are resolved by linking the two together.

In this situation, to avoid losing data by overriding any of the
documents (i.e. the existing one and the one we're trying to push)
we'll rename the existing remote document with a conflict suffix and
mark our local change with an increased error counter so it can be
retried later (and work this time as the existing document has been
renamed).

- [x] PR is not too big

### DIFF
--- a/core/remote/cozy.js
+++ b/core/remote/cozy.js
@@ -52,6 +52,25 @@ class DirectoryNotFound extends Error {
   }
 }
 
+class FileNotFound extends Error {
+  /*::
+  path: string
+  cozyURL: string
+  */
+
+  constructor(path /*: string */, cozyURL /*: string */) {
+    super(`File ${path} was not found on Cozy ${cozyURL}`)
+
+    if (Error.captureStackTrace) {
+      Error.captureStackTrace(this, FileNotFound)
+    }
+
+    this.name = 'FileNotFound'
+    this.path = path
+    this.cozyURL = cozyURL
+  }
+}
+
 const COZY_CLIENT_REVOKED_ERROR = 'CozyClientRevokedError'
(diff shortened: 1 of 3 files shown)


nono:
> This can only happen when we try to push a new file

No, it can also happen when trying to rename/move a file to a destination that already exists but has not yet been fetched by the remote watcher.

> conflicts on directories are resolved by linking the two together

IIRC, the conflict resolution happens in the merge, not in the sync. So, we can have a 409 for directories if one directory is renamed in local to a name for a new directory that has not yet been fetched by the remote watcher.

> In this situation, to avoid losing data by overriding any of the
documents (i.e. the existing one and the one we're trying to push)
we'll rename the existing remote document with a conflict suffix and
mark our local change with an increased error counter so it can be
retried later (and work this time as the existing document has been
renamed).

In theory, we were just increasing the error counter and waiting that the remote watcher fetches the document on the Cozy, so that Merge can resolve the conflict. And a next retry should succeed. Can you explain why this strategy doesn't work here?

taratatach:
In theory we would catch conflicts at Merge time indeed. I came up with those changes as a strategy to cope with an unknown issue a user was facing.

I went back to analyzing this user's logs and came up with another, more realistic, explanation for his issues and this PR seems unnecessary now.

I'll try not to get satisfied with coping with symptoms in the future. Thanks for holding me back here :).

PR #1795, opened 2020-01-07
core/sync: Properly stop synchronizations

taratatach:
The way the synchronization is stopped at the moment does not properly
end infinite loops that take care of waiting for changes either local
or remote.
This is an issue especially when requesting a manual synchronization
as it works by stopping the current synchronization and starting a new
one. If the former is not completely stopped, we can end up with
multiple loops working on changes making for unpredictable behavior.

We also suspect that this may cause errors when trying to disconnect
the client from the Cozy with a process still using the PouchDB logs
file while trying to remove it.

We now use a promise that resolves when we call the `Sync#stop()`
method and we race it again synchronization promises so we can know
when to break out of the loops.

- [x] PR is not too big

### DIFF
--- a/core/app.js
+++ b/core/app.js
@@ -321,17 +321,12 @@ class App {
   // Start the synchronization
   startSync(mode /*: SyncMode */) {
     this.config.saveMode(mode)
-    log.info('Run first synchronisation...')
     return this.sync.start(mode)
   }
 
   // Stop the synchronisation
   stopSync() {
-    if (!this.sync) {
-      // $FlowFixMe
-      return Promise.resolve()
-    }
-    return this.sync.stop()
+    return this.sync && this.sync.stop()
   }
 
   async setup() {
--- a/core/sync.js
+++ b/core/sync.js
@@ -60,8 +60,10 @@ class Sync {
   local: Local
   pouch: Pouch
   remote: Remote
-  stopped: ?boolean
   moveTo: ?string
+  stopRequested: Promise
+  started: ?Promise<boolean>
+  running: ?Promise
 
   diskUsage: () => Promise<*>
   */
@@ -82,6 +84,7 @@ class Sync {
     this.events = events
     this.local.other = this.remote
     this.remote.other = this.local
+    this.stopRequested = false
 
     autoBind(this)
   }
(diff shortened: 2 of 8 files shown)


core/sync.js
```diff
@@ -123,23 +135,30 @@ class Sync {
...
-      this.changes = null
+  async stop() {
+    log.info('Stopping Sync...')
+    if (this.runningResolve) {
+      this.runningResolve()
+      this.runningResolve = null
     }
-    return Promise.all([this.local.stop(), this.remote.stop()])
+    await this.stopped
+    log.info('Sync stopped')
```
nono: I don't know if it is an issue, but if there was a change being applied, `stop` can return before it has finished, and when the change will finish, it will write to PouchDB. Maybe we should look if we can cancel the change.

core/sync.js
```diff
@@ -60,8 +60,9 @@ class Sync {
   local: Local
   pouch: Pouch
   remote: Remote
-  stopped: ?boolean
   moveTo: ?string
+  stopped: Promise
```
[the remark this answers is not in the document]
taratatach: Right. Flow isn't complaining but that might be because of the Promise.

core/app.js
```diff
@@ -321,16 +321,12 @@ class App {
...
     return this.sync.start(mode)
   }
 
   // Stop the synchronisation
   stopSync() {
-    if (!this.sync) {
-      // $FlowFixMe
-      return Promise.resolve()
-    }
+    if (!this.sync) return
```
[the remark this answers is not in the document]
taratatach: We do return a Promise. It resolves to `undefined`.
I'm OK with the simplification.

core/sync.js
```diff
@@ -107,14 +109,24 @@ class Sync {
...
       await started
     }
-    await new Promise(
+    this.stopped = new Promise(
       async function(resolve, reject) {
         Promise.all(sidePromises).catch(err => reject(err))
+
+        this.runningResolve = async () => {
+          await Promise.all([this.local.stop(), this.remote.stop()])
+          resolve()
```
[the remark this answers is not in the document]
taratatach: Sounds good.

core/sync.js
taratatach: It would be better to wait until the change is applied. The goal of the PR is to properly stop the synchronization so it should be part of it.

nono: Even if it is a long download?

taratatach: Since we don't rollback changes saved in Pouch but not completely applied, yes we should.

core/sync.js
```diff
@@ -95,51 +98,88 @@ class Sync {
...
+    })
+
+    this.started = new Promise(async (resolve, reject) => {
+      if (this.stopRequested) reject()
+
+      log.info('Starting Sync...')
+
+      try {
+        let sidePromises = []
+        if (mode !== 'pull') {
```
nono: Maybe it is time to remove the push and pull modes. I don't think they are still used.

core/sync.js
```diff
@@ -60,8 +60,10 @@ class Sync {
   local: Local
   pouch: Pouch
   remote: Remote
-  stopped: ?boolean
   moveTo: ?string
+  stopRequested: Promise
+  started: ?Promise
+  running: ?Promise<boolean>
```
nono: It looks like the code is checking the `started` promise as a boolean, but not the `running` promise:

```js
started: ?Promise<boolean>
running: ?Promise
```

core/sync.js
```diff
@@ -95,51 +98,88 @@ class Sync {
...
+      if (this.changes) {
+        this.changes.cancel()
+        this.changes = null
+      }
+
+      await Promise.all([this.local.stop(), this.remote.stop()])
+
+      this.started = Promise.resolve(false)
+      if (runningResolve) {
+        runningResolve(false)
```
nono: `runningResolve()` should be enough

core/sync.js
```diff
@@ -95,51 +98,88 @@ class Sync {
...
+      throw err
+    } finally {
+      if (this.changes) {
+        this.changes.cancel()
+        this.changes = null
+      }
+
+      await Promise.all([this.local.stop(), this.remote.stop()])
+
+      this.started = Promise.resolve(false)
```
nono: Or `this.started = null`

core/sync.js
```diff
@@ -95,51 +98,88 @@ class Sync {
...
+        Promise.all(sidePromises).catch(err => {
+          if (runningReject) runningReject(err)
+        })
+      } catch (err) {
+        log.error({ err }, 'Could not start watchers')
+        this.local.stop()
+        this.remote.stop()
+        reject(err)
+      }
+      resolve()
```
nono: `resolve(true)` for L101

core/sync.js
```diff
@@ -95,51 +98,88 @@ class Sync {
...
-        } catch (err) {
-          reject(err)
+    let runningResolve, runningReject
+    this.running = new Promise((resolve, reject) => {
+      runningResolve = resolve
+      runningReject = reject
+    })
+
+    this.started = new Promise(async (resolve, reject) => {
+      if (this.stopRequested) reject()
```
nono: We should also return here!

core/sync.js
nono: Does it mean that the client won't exit when the user asks to quit until the download as finished?

taratatach: I'll run a test to see what happens.

taratatach: I've ran a test and we do wait for the end of the download before exiting the client.
I think it's a good thing knowing how we manage inconsistencies in Pouch today but I understand this might be upsetting to some users :/.

core/sync.js
taratatach: After removing it I wonder if this is actually something we want to do or if we could use that to give users the option to use Cozy Desktop as a backup software :thinking:.
I'll create another PR to discuss this.

core/sync.js
nono: I don't follow you. As I understand, the sync doesn't save anything in PouchDB until the download has finished. And if we stop the download before its end, we can find a trick to not write at all in PouchDB for this operation. So, we will have no inconsistencies.

taratatach: Indeed the Sync won't write in PouchDB until the download is done.
But the Merge has already written in PouchDB the new version of the document. If by any chance this was not a new document but an update, we'll end up loosing the update by overwriting the remote version with out-of-date local one (on top of a conflict reanming :/).

nono: Can we find another way to fix this issue? Because, if I understand correctly, if several files are updated on the Cozy, not stopping the download will only save the first one, the two other will still be corrupted when cozy-desktop will be restarted.

taratatach: You're right, this won't help with multiple downloads. This wasn't mean as a fix for this issue though.

taratatach: I've introduced a specific method for the manual sync that will wait for the current sync operation to end while leaving the app won't wait.

PR #1797, opened 2020-01-14
core/merge: Don't override unsynced file updates

taratatach:
When starting Cozy Desktop, we gather and merge local changes then
remote changes and last we try to synchronize them.
This can create issues when the same document was modified both
locally and remotely.

So, to avoid loosing data, we decided to rename the document with a
conflict suffix on the side that would overwrite the existing
document. In the case of a startup sync, this would be the remote
side.

However, if the application was stopped right after a remote update
was merged and before it was propagated to the local file system, when
starting it up again, we'll wrongly detect a local update (i.e. the
local version does not reflect the remote version and its Pouch
representation). In this case, the conflict resolution would happen on
the local file system, thus generating a local `renamed` event that
would lead to the remote version being overwritten with the local
version on top of being renamed with a conflict suffix.

We could decide this is fine now that we have versions management on
the remote Cozy but users don't seem to think about looking at those
and believe their changes are gone.
Besides, the change would be propagated to other Cozies in case the
file was shared and this can be even harder to understand for the
owners of those Cozies.

So we decided to overwrite the local version with the unsynced remote
version. Since, the local version could have really been updated,
we'll create a backup copy of the local file (i.e. with the `.bck`
file extension) and trash it for safe keeping in case that version was
the one the user wanted to keep.

- [x] PR is not too big

### DIFF
--- a/core/local/chokidar/local_change.js
+++ b/core/local/chokidar/local_change.js
@@ -664,7 +664,7 @@ function includeChangeEventIntoFileMove(
   const moveChange /*: ?LocalFileMove */ = maybeMoveFile(sameInodeChange)
   if (!moveChange) return
   log.debug({ path: e.path }, 'FileMove + change')
-  moveChange.md5sum = moveChange.old.md5sum || moveChange.md5sum
+  moveChange.md5sum = e.md5sum
   moveChange.update = _.defaults(
     {
       // In almost all cases, change event has the destination path.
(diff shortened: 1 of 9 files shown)


taratatach:
> Let's say it is OK. I don't know, I don't see anything wrong but core/local/chokidar is a black box for me.

OK. I'll spend time to particularly test this anyway.

PR #1800, opened 2020-01-22
core: Get instance subdomains capabilities

taratatach:
We'll need to know if a Cozy uses flat or nested subdomains to address
applications so we can open Cozy Notes on the user's Cozy.
This information is stored in the instance's capabilities part of the
settings and can be access via the `/settings/capabilities` route.

Since those settings should not change very often, we can avoid making
making a request each time we want to know if an instance uses flat or
nested subdomain by storing it in the Desktop client's config.

Now that we can store the flat/nested subdomains capabilities of the
remote Cozy, we need a way to get the information.

The `/settings/capabilities` route is accessible via the `cozy-client`
package but not via the `cozy-client-js` package that we already use.
Since `cozy-client-js` is deprecated, we won't add this route to the
package and we need to use `cozy-client`.

To avoid transitioning all Desktop right now, we'll use the
`fromOldOAuthClient` method of `cozy-client` which lets us instantiate
a CozyClient object from an old CozyClient object.

From there, we can simply query the `io.cozy.settings` doctype for
the `capabilities` document and get the information we need.

- [x] PR is not too big

### DIFF
--- a/core/remote/index.js
+++ b/core/remote/index.js
@@ -54,6 +54,7 @@ export type RemoteOptions = {
 class Remote /*:: implements Reader, Writer */ {
   /*::
   other: Reader
+  config: Config
   pouch: Pouch
   events: EventEmitter
   watcher: RemoteWatcher
@@ -62,6 +63,7 @@ class Remote /*:: implements Reader, Writer */ {
   */
 
   constructor({ config, prep, pouch, events } /*: RemoteOptions */) {
+    this.config = config
     this.pouch = pouch
     this.events = events
     this.remoteCozy = new RemoteCozy(config)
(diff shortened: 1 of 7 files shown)


core/remote/index.js
```diff
@@ -404,6 +408,32 @@ class Remote /*:: implements Reader, Writer */ {
...
+        flatSubdomains = false
+      } catch (err) {
+        if (err.syscall === 'getaddrinfo' && err.code === 'ENOTFOUND') {
+          flatSubdomains = true
+        } else {
+          throw err
+        }
+      }
+
+      this.config.capabilities = { flatSubdomains }
```
nono: [nit] I'm not a fan of using `set` to make the assignation perform a persist

taratatach: It's kind of a hidden behavior yes. But that would mean changing the whole config module.
It'll probably happen someday :).

PR #1801, opened 2020-01-27
tools: Associate `.cozy-note` files with Desktop

taratatach:
Create a file association on the user's computer via the
`electron-builder` configuration.
This will create the `text/vnd.cozy.note+markdown` MIME type on macOS
and Windows and will set Cozy Desktop as the default application to
open files with the `cozy-note` extension.

On Linux, since we don't install the application but hand an
executable, we need to create the MIME type by hand by putting a
declaration file in the local MIME types database.
The association will only work if a `desktop` file with the right MIME
type association points to the Desktop app. This is done automatically
if using `appimaged`.

- [x] PR is not too big

### DIFF
--- a/build/launcher-script.sh
+++ b/build/launcher-script.sh
@@ -3,6 +3,49 @@
 # Uncomment for debugging purposes:
 # set -eux
 
+APP_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+RESOURCES_DIR="$APP_DIR/resources"
+DESKTOP_DATABASE_DIR="$HOME/.local/share/applications"
+MIME_DATABASE_DIR="$HOME/.local/share/mime"
+CUSTOM_MIME_DIR="$MIME_DATABASE_DIR/packages"
+COZY_NOTE_MIME_TYPE_DECLARATION_FILENAME="vnd.cozy.note+markdown.xml"
+
+function isCozyNoteMimeTypeDeclared() {
+    # does our MIME type declaration file exist in the MIME database folder?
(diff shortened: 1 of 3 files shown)


build/launcher-script.sh
```diff
@@ -3,6 +3,45 @@
...
+function isCozyNoteMimeTypeDeclared() {
+    [[ -f "$CUSTOM_MIME_DIR/$COZY_NOTE_MIME_TYPE_DECLARATION_FILENAME" ]] && return
+
+    false
+}
+
+function declareCozyNoteMimeType() {
+    cp "$RESOURCES_DIR/$COZY_NOTE_MIME_TYPE_DECLARATION_FILENAME" "$CUSTOM_MIME_DIR/$COZY_NOTE_MIME_TYPE_DECLARATION_FILENAME"
+    # rebuild the local MIME database to make sure our new type is known
+    update-mime-database $MIME_DATABASE_DIR
```
clochix: If the folder `~/.local/share/mime` doesn’t exist, this command will fail. So we could add

    mkdir -p "$CUSTOM_MIME_DIR"

before `update-mime-database`

build/launcher-script.sh
```diff
@@ -3,6 +3,49 @@
 # Uncomment for debugging purposes:
 # set -eux
```
nono: I think we should always have `set -eux`, not just for debugging.

taratatach: In our case, we can't because the definition of `UNPRIVILEGED_USERNS_ENABLED` relies on a command call that will error out if it isn't enabled.

PR #1805, opened 2020-01-30
gui: Open note in browser when invoked with path

taratatach:
Cozy Desktop is declared as the default application to open
`.cozy-note` files (or files with the `text/vnd.cozy.note+markdown`
MIME type on linux).

We're now detecting when the app is invoked with arguments (i.e. a
file path) and we'll try to open it within the Cozy Notes web
application.

Some restrictions apply:
- we first check if the path is matching a document within our local
Pouch database. If not, we won't open it.
- then we check if it is a valid Cozy Note. If not, we won't open it
- then we check if we own this note. For now, we won't open it if we
don't own it since opening it in the owner's Cozy Note application
requires some work on the stack side.

We'll reconstruct the Cozy Note's url on the user's Cozy and open it
in the default browser.
For now, if the user's offline, we'll show an error saying we couldn't
find the note in the user's Cozy. This should evolve in the future.

When fetching a remote document from the Cozy, we were selecting only
some attributes to save in our local database.
This prevents us from having access to new fields unless we've
explicitly listed them. We haven't updated this list in a while so our
documents are missing the new `cozyMetadata` attribute which we'll
need to determine if a Note is owned by the Desktop's user or later to
synchronize dates correctly.

We don't see any value in making a selection in the attributes so
we'll stop filtering out some of them so new attributes can be saved
as soon as a document is updated with them on the Cozy.

We'll need to build a migration to update existing documents since
we've already marked them as seen via the remote sequence and the
remote watcher won't fetch them again unless they're updated on the
Cozy.

- [x] PR is not too big


gui/notes/index.js
```diff
@@ -0,0 +1,147 @@
...
+      ? `${slug}.${url.host}`
+      : url.host
+          .split('.')
+          .map((x, i) => (i === 0 ? x + '-' + slug : x))
+          .join('.')
+  url.hash = nativePath
+  return url.toString()
+}
+
+const options = async (
```
nono: You don't use the new route from the stack that works also for notes from a shared directory (https://github.com/cozy/cozy-stack/pull/2345). Is it on purpose?

gui/notes/index.js
```diff
@@ -0,0 +1,147 @@
...
+ * computeNoteURL - Constructs a link to the Note within the web app
+ *
+ * @param {object} Options               Object of options
+ * @param {string}   options.noteId        Id of the CouchDB document representing the given note
+ * @param {string}   options.cozyUrl       Base URL of the cozy, eg. cozy.tools or test.mycozy.cloud
+ * @param {string}   options.subDomainType Whether the cozy is using flat or nested subdomains. Defaults to flat.
+ *
+ * @return {string} Generated URL
+ */
+const computeNoteURL = async ({ noteId, cozyUrl, subDomainType }) => {
```
nono: I thought that this function would be shared between apps. Is there a reason to not add it to cozy-client?

gui/notes/index.js
```diff
@@ -0,0 +1,147 @@
...
+
+const isMyNote = (
+  note /* CozyFile */,
+  { config } /*: { config: Config } */
+) => {
+  const isNote = models.file.isNote(note)
+  const isMine = note.cozyMetadata.createdOn.startsWith(config.cozyUrl)
+
+  return isNote && isMine
+}
```
nono: Maybe we can add some tests for the functions in this file

gui/notes/index.js
Crash--: @nono +1.

We already have something similar : https://github.com/cozy/cozy-client/blob/master/packages/cozy-client/src/models/note.js#L6

gui/notes/index.js
Crash--: We can also extract the isMine method to `models.file.isMine` since it'll also be shared 👍

gui/notes/index.js
taratatach: Ah yes, I added this function at the very beginning of my experiments when I was not using `cozy-client`.
I'll take a look at how I can replace that.

gui/notes/index.js
taratatach: I didn't know it was implemented already :). When do you think it's gonna be merged so I can test it?

gui/notes/index.js
taratatach: @Crash-- that's a good idea. I'll make a PR.

PR #1808, opened 2020-02-10
gui/main: Display Markdown on note open failure

taratatach:
We want to be able to display the markdown content of Cozy Notes we
could not open in the web applications (e.g. because the computer is
oflline).

We'll use `react-markdown` to parse and display an HTML version of the
markdown content so we create a new HTML base that will load a special
javascript file which will import React and the other required
components.
The content of the note, its filename and other information like the
title and details of the error that lead us to open the viewer will be
passed via an asynchronous message sent to the window itself.
The error will be displayed in a banner at the top of the window so
the user knows why we did not open the web application and displayed
the note's content instead.

When we fail to open a note in the Cozy Notes web application either
because we're offline or the note could not be found, we'll try to
load the Markdown content from the file itself and display it within
the Markdown viewer window with a human readable error message within
the banner so the user can understand why we're not showing her note
within the web application.
If we can't the error was of another kind, we'll display the error
name and message within the banner.

If we fail to load the file's content and open it within the Markdown
viewer, we'll show a dialog box with the error name and message.

- [ ] PR is not too big


gui/markdown-viewer.js
```diff
@@ -0,0 +1,74 @@
...
+        e('span', { className: 'banner__error-title' }, title)
+      ]),
+      e('p', { className: 'banner__error-details' }, details),
+      e(
+        'div',
+        {
+          className: 'banner__close',
+          onClick: this.close
+        },
+        '\u2715'
```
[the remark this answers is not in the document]
taratatach: It's an `X`

gui/markdown-viewer.js
```diff
@@ -0,0 +1,76 @@
+const React = require('react')
+const ReactDOM = require('react-dom')
+const ReactMarkdown = require('react-markdown')
+const e = React.createElement
+
+class Banner extends React.Component {
+  constructor() {
```
Crash--:
```suggestion
  constructor(props) {
```

gui/markdown-viewer.js
```diff
@@ -0,0 +1,76 @@
...
+    super()
```
Crash--:
```suggestion
    super(props)
```

gui/markdown-viewer.js
```diff
@@ -0,0 +1,76 @@
...
+
+    this.close = this.close.bind(this)
+  }
+
+  close() {
+    this.setState({ closed: true })
+  }
+
+  render() {
+    const { translate, level, title, details } = this.props
```
Crash--: Since you're used to have type checking with Flow, you can use something like https://www.npmjs.com/package/prop-types to check and warn about the type you're using. 

It's far from flow or TS, but at least, you can define the "API" of the component. 

Banner.propTypes = {
  translate: PropTypes.func.required, 
  level: PropTypes...
}

gui/markdown-viewer.js
```diff
@@ -0,0 +1,76 @@
...
+    const { content: source, filename } = this.props
+
+    return e('div', { className: 'markdown-viewer' }, [
+      e('header', { className: 'markdown-viewer__filename' }, [
+        e('h1', null, filename)
+      ]),
+      e(ReactMarkdown, { source }, null)
+    ])
+  }
+}
```
Crash--: If you extends Component, you have to call super(props). But here what you need it's just a "render component". So let's remove this Component class and use a functional component : 

```jsx
function MarkdownViewer(props){
 const { content: source, filename } = props
    return e('div', { className: 'markdown-viewer' }, [
      e('header', { className: 'markdown-viewer__filename' }, [
        e('h1', null, filename)
      ]),
      e(ReactMarkdown, { source }, null)
    ])
}
```

gui/styles/_markdown_viewer.styl
```diff
@@ -0,0 +1,110 @@
+black = #000
+chablis = #fff2f2
+charcoalGrey = #32363f
+coolGrey = #95999d
+paleGrey = #f5f6f7
+pomegranate = #f52d2d
```
Crash--: Come on! Our only source of colors should be cozy-ui! (https://github.com/cozy/cozy-ui/blob/8c26e8a02220bd3816e03893441429e54d01299e/stylus/settings/palette.styl#L37) 

We should work on using it in the desktop app, right? ;)

gui/markdown-viewer.html
```diff
@@ -0,0 +1,12 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta charset="utf-8">
+    <title>Cozy Drive</title>
+    <link rel="stylesheet" href="app.css">
```
Crash--: how this css file is generated? Is it the preexisting one so we're loading all the CSS (https://github.com/cozy-labs/cozy-desktop/blob/86b7e6c230ace0cc6d273a0dc209380ff057bd13/gui/styles/app.styl) of the desktop app? 

Since in `_markdown_viewer.styl` I don't see any import, it does mean that this file is pretty standalone?

gui/markdown-viewer.js
taratatach: Ah yes I could use PropTypes!

gui/markdown-viewer.js
taratatach: Thanks! I wanted to use a function rather than a component but I started with `Banner` which has a state and didn't think about it for the viewer :)

gui/styles/_markdown_viewer.styl
taratatach: Totally agreed. But I didn't want to go through the upgrade of `cozy-ui` right now. I'll give it a try and see if it requires work or not :).

gui/markdown-viewer.html
taratatach: I'm actually relying on the style of the rest of the app. So I need to import `app.css`.

gui/styles/_markdown_viewer.styl
taratatach: We tried to upgrade `cozy-ui` but it's too much work for now. We'll continue in https://github.com/cozy-labs/cozy-desktop/pull/1810

PR #1836, opened 2020-03-16
core/merge: Mark documents for deletion

taratatach:
Until now, when we detected the deletion of a document either on the
local file system or the remote Cozy, we would merge it by deleting
it from the PouchDB database.
Then, The Sync module would detect that change via the PouchDB changes
feed and propagate it.

However, deleted documents in PouchDB cannot be queried by `get()`.
Also, when multiple modifications are made to the same document (i.e.
saved with the same `_id` attribute), the changes feed will return
only the latest version of the document.

Those 2 "restrictions" make merging some changes harder. For example:
- the remote addition of files or folders into a locally deleted
  folder (i.e. the parent folder can't be queried so it's "missing")
- the deletion and creation of a file at the same path/_id (i.e. we
  lose the _rev of the deleted document and can't propagate it as an
  update)

The solution we found to this problem is to mark documents for
deletion during the Merge step so they can still be queried until the
deletion has been propagated. This way we can decide to cancel the
deletion or take any other appropriate action to merge other changes
that could not be merged (or merged in a way that would result in an
incorrect or impossible propagation).

This commit introduces the `deleted` marker and does not make use of
it to solve our problems yet. Its goal is to achieve feature parity.

- [x] PR is not too big

### DIFF
--- a/core/merge.js
+++ b/core/merge.js
@@ -103,7 +103,7 @@ class Merge {
 
     try {
       const folder = await this.pouch.db.get(parentId)
-      if (folder) {
+      if (folder && !folder.deleted) {
         return
       }
     } catch (err) {
@@ -163,6 +163,10 @@ class Merge {
     metadata.markSide(side, doc, file)
     metadata.assignMaxDate(doc, file)
     if (file) {
+      if (file.deleted) {
+        return this.updateFileAsync(side, doc)
+      }
+
       const idConflict /*: ?IdConflictInfo */ = IdConflict.detect(
         { side, doc },
         file
@@ -261,7 +265,7 @@ class Merge {
         if (doc.mime == null) {
           doc.mime = file.mime
         }
-      } else if (!metadata.isAtLeastUpToDate(side, file)) {
+      } else if (!file.deleted && !metadata.isAtLeastUpToDate(side, file)) {
         if (side === 'local') {
           // We have a merged but unsynced remote update and we can't create a
           // conflict because the local rename will trigger an overwrite of the
@@ -357,7 +361,7 @@ class Merge {
   ) /*: Promise<*> */ {
     log.debug({ path: doc.path, oldpath: was.path }, 'moveFileAsync')
     const { path } = doc
-    if (!metadata.wasSynced(was)) {
+    if (!metadata.wasSynced(was) || was.deleted) {
       metadata.markAsUnsyncable(side, was)
       await this.pouch.put(was)
       return this.addFileAsync(side, doc)
@@ -388,6 +392,10 @@ class Merge {
 
       const file /*: ?Metadata */ = await this.pouch.byIdMaybeAsync(doc._id)
       if (file) {
+        if (file.deleted) {
+          doc.overwrite = file
+        }
+
         const idConflict /*: ?IdConflictInfo */ = IdConflict.detect(
           { side, doc, was },
           file
@@ -462,6 +470,10 @@ class Merge {
 
     const folder /*: ?Metadata */ = await this.pouch.byIdMaybeAsync(doc._id)
     if (folder) {
+      if (folder.deleted) {
+        doc.overwrite = folder
+      }
+
       const idConflict /*: ?IdConflictInfo */ = IdConflict.detect(
         { side, doc, was },
         folder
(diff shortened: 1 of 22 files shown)


taratatach:
> I agree with the goal, but it looks like just merging this PR will make things worse. Is it correct?

Well, it should not. It should be mergeable by itself. If you see issues, I'm interested.
If implemented correctly, documents with the `deleted` attribute set to `true` should be considered as erased documents.

nono:
Well, I didn't think a lot about it. Maybe I'm wrong, but I would have expected a lot of edge cases (like moving a file several times or temporary files that are created and removed without ever being pushed to the cozy instance).

taratatach:
I forgot to check the result of a bunch of getters from `Pouch`… Even though the tests are all green, we'd probably see problems indeed.
I'm fixing that right now and will push another version.

core/merge.js
```diff
@@ -548,7 +560,15 @@ class Merge {
 
       const existingDstRev = existingDstRevs[dst._id]
       if (existingDstRev && folder.overwrite) {
-        dst._rev = existingDstRev
+        if (src.deleted) {
+          // Don't create destination that would overwrite existing file
+          continue
+          // TODO: document with test deleting child file then move its parent
+          // folder, overwriting another folder with child file of same name.
+          // The file in the desitnation folder should not be deleted.
```
nono, line 568: *destination

core/sync.js
```diff
@@ -332,13 +336,12 @@ class Sync {
...
         delete doc.overwrite
+        if (doc.deleted) doc._deleted = true
       }
 
       log.trace({ path, seq }, `Applied change on ${sideName} side`)
       await this.pouch.setLocalSeqAsync(change.seq)
-      if (!change.doc._deleted) {
-        await this.updateRevs(change.doc, sideName)
-      }
+      await this.updateRevs(doc, sideName)
```
nono, line 344: No migration phase is planned? I mean if a user deletes a file locally with the current version of the desktop client while offline, and migrates to the new version of the desktop client, the client won't try to synchronize this deletion. Right?

core/sync.js
```diff
@@ -307,6 +307,10 @@ class Sync {
 
     if (metadata.shouldIgnore(doc, this.ignore)) {
       return this.pouch.setLocalSeqAsync(change.seq)
+    } else if (!metadata.wasSynced(doc) && doc.deleted) {
+      doc._deleted = true
```
nono, line 311: [nit] I think we can even do `doc = { _deleted: true }` to not keep the other fields (changes feed)

core/sync.js
```diff
@@ -588,7 +591,13 @@ class Sync {
     doc /*: Metadata */,
     side /*: SideName */
   ) /*: Promise<*> */ {
-    metadata.markAsUpToDate(doc)
+    if (doc.remote != null) {
+      metadata.markAsUpToDate(doc)
+    } else if (doc.sides && doc.sides.remote != null) {
+      // Avoid Metadata has sides.remote but no remote error when saving the an
```
nono, line 597: s/the an/an/

core/merge.js
taratatach: Thanks. And I should implement the test.

core/sync.js
taratatach: We still need the `_id`. Do you think we'd get some performance or storage gain from this change?

core/sync.js
taratatach: Right, I need to think about this.

taratatach:
It seems like this could get more work. I'll push back its integration into the beta release.

core/sync.js
nono: It would take a bit less storage. But the main motivation should be to avoid getting stale data (for example, it may be used to know that the file has been removed remotely during the migration phase).

taratatach: Noted. To make the difference between documents `_deleted` during the migration phase and those "erased" after we've synchronized the deletion, I was thinking of using the sides. If they're up-to-date then it means there's nothing to do (except on that specific line where I check for both `_deleted` and `deleted` now).

taratatach: Ah, I can't do what I proposed, because moves are still erasing source documents.
I'll try your idea.

taratatach:
I should have addressed your concerns about the transition and how to handle erased documents in Sync.

PR #1862, opened 2020-04-15
core/merge: Don't trash file moved on other side

taratatach:
We believe users expect the path of their documents to be their
identifier and thus that deleting a file on one side while it was
moved on the other should not result in the file being deleted since
its path has changed.

To meet those expectations, we undo the trashing/deletion of files
when we detect they have been moved on the other side.

We note that when we merge the local deletion of a file before we
merge its remote movement, we cannot undo the merged trashing since
the PouchDB record has been erased.
However, the local deletion is not propagated to the remote Cozy
thanks to the movement updating the remote revision of the document
and thus implying a rejected trashing.
This situation can be stabilized once we'll have deletion markers
since we'll be able to fetch the deleted record when merging the
movement.

We're now preventing the deletion of files moved on the other side but
if the file was completely erased on the Cozy (i.e. removed from the
trash as well), we can't restore it and need to re-upload it.

Besides, it's managed through another method,
`Merge.deleteFileAsync()` and needs its own behavior changes.

- [x] PR is not too big

### DIFF
--- a/core/merge.js
+++ b/core/merge.js
@@ -671,6 +671,26 @@ class Merge {
       log.error({ doc, was, sentry: true }, 'Mismatch on doctype for doTrash')
       return
     }
+    if (was.moveFrom) {
+      // The file was moved and we don't want to delete it as we think users
+      // delete "paths".
+      if (side === 'remote') {
+        // We update the remote rev so we can send the file again and undo the
+        // remote trashing.
+        was.remote._rev = doc.remote._rev
+        // We keep the `moveFrom` hint so we will update the remote file and
+        // restore it from the trash instead of re-uploading it.
+        was.moveFrom.remote._rev = doc.remote._rev
+      } else {
+        // We remove the hint that the file should be moved since it has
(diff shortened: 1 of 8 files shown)


core/merge.js
```diff
@@ -725,6 +750,23 @@ class Merge {
...
+        // The file was completely destroyed on the Cozy so it cannot be
+        // restored fromt the trash and should be re-uploaded.
+        delete file.remote
+        delete file.sides.remote
+        delete file.moveFrom
+        return this.pouch.put(file)
+      } else {
+        log.warn(
+          { path: doc.path, doc, file },
+          'deleteFileAsync should not be called for local side'
```
nono, line 766: You are sure about about that? I can be wrong, but it looks like https://github.com/cozy-labs/cozy-desktop/blob/master/core/prep.js#L132 could do that.

taratatach: Ah crap, I missed this one. Good catch.

PR #1865, opened 2020-04-22
core: Refactor sync modules lifecycle management

taratatach:
We found out that we were not catching all errors coming from the remote watcher thus not alerting the user and not completely stopping the synchronization process.

We want to make sure this does not happen anymore and that other similar situations cannot happen.
To do so, we need to move away from the complex lifecycle management we have today in the watchers and sync modules.

- [ ] PR is not too big

### DIFF
--- a/core/sync.js
+++ b/core/sync.js
@@ -131,14 +131,12 @@ class Sync {
 
       try {
         await this.local.start()
-        const localRunning = this.local.watcher.running
-        const {
-          running: remoteRunning,
-          started: remoteStarted
-        } = this.remote.start()
-        await remoteStarted
+        await this.remote.start()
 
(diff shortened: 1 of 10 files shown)


core/sync.js
```diff
@@ -112,73 +111,51 @@ class Sync {
...
+      return
+    }
+    this.lifecycle.end('start')
+
+    Promise.all([
+      this.local.watcher.running,
+      this.remote.watcher.running
+    ]).catch(err => {
+      this.error(err)
+      this.stop()
```
nono, line 141: No `await` here?

taratatach: We're not awaiting the two promises here so we're not await the `catch` handler either so awaiting `this.stop()` would not change anything.

PR #1878, opened 2020-05-07
tests/scenarios: Use defined files content in captures

taratatach:
When creating files in a test scenario, we can define its content via
the `content` attribute. If it's not present, the content will default
to "whatever".
However, this attribute was not taken into account in the capture's
init phase, resulting in inadequate captures and thus test failures or
incorrectness.

- [ ] PR is not too big


taratatach:
The failing tests are known and will be fixed in separate PRs as the failures are not due to scenarios' helpers or orchestration incorrect behaviors.

PR #1882, opened 2020-05-12
core/merge: Create conflict on local Note edit

taratatach:
Cozy Notes are a special kind of document. Their content lives within
their CouchDB record metadata and the associated file is only a
markdown export.
Besides, the stack is not able to re-import markdown into a Cozy Note
document for the moment.

This means that if a synchronized Cozy Note file is edited on the
user's local filesystem, we won't be able to import the new content
into the remote Cozy Note. On a side note, the client does not handle
remote records metadata so the edited CouchDB record will lose the
original Cozy Note content.

We've added some protections by forbidding the propagation of local
Note edits but this means the Note will be de-synchronized with
different contents locally and on the remote Cozy.
In case the Note is edited again, via the Cozy Note app this time, the
local content will be overwritten with the new markdown export and the
locally edited content will be lost.

To avoid losing this content we can create a conflict on local Notes
edits so that the original Note will keep its metadata and actual
content and a new markdown file will be uploaded to the remote Cozy
with the newly edited content.
Since users can keep their editor open when saving the new local
content, we can't change the name of the local file and have to rename
the remote file, thus the original Note.

- [x] PR is not too big

### DIFF
--- a/core/merge.js
+++ b/core/merge.js
@@ -135,17 +135,13 @@ class Merge {
   // A suffix composed of -conflict- and the date is added to the path.
   async resolveConflictAsync(
     side /*: SideName */,
-    doc /*: Metadata */,
-    was /*: ?Metadata */
+    doc /*: Metadata */
   ) /*: Promise<Metadata> */ {
-    log.warn(
-      { path: doc.path, oldpath: was && was.path, doc, was },
-      'resolveConflictAsync'
-    )
     const dst = metadata.createConflictingDoc(doc)
+    log.warn({ path: dst.path, oldpath: doc.path }, 'Resolving conflict')
     try {
       // $FlowFixMe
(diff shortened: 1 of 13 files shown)


core/merge.js
```diff
@@ -135,23 +136,63 @@ class Merge {
...
     } catch (err) {
       throw err
     }
     return dst
   }
 
+  // Resolve Cozy Note conflict when its content is updated on the local
+  // filesystem and would therefore break the actual Note.
+  // We always rename the remote document since this method should only be
+  // called in csae of a local update and the local file could still be open in
```
nono, line 155: s/csae/case/

taratatach:
> I didn't see a test for this behavior on conflict for a local note edition. Am I missing it?

Arf, I created the tests in a new file that I forgot to add. I'll push a new version.

PR #1890, opened 2020-05-29
core/sync: Pause local watcher during local writes

taratatach:
When we fetch multiple changes on the same document from the remote
Cozy, the Sync will apply them all one after the other.
For example, if a file was moved and updated, we'll move the file on
the local filesystem and then download the remote version to update
the local content.

Those actions are run after the PouchDB record for the document has
been updated to reflect the final expected state of the document
metadata.

However, the local watcher will receive events for each action when
it's executed. When it receives an event, it will fetch the document's
stats and calculate the checksum of files. In our example, it means
the event fired after the old file was moved will fetch the stats and
checksum of the old version, different from what is stored in PouchDB.
When the event reaches the Merge module, we'll detect an unexpected
file change from the new version to the old.

To avoid this timing issue we will pause the local watcher and buffer
events until all Sync actions have been executed on the local
filesystem so that when the watcher is resumed, the fetched stats and
checksums will reflect the final, expected state of the document.

The modified scenario does not fail without those changes.
However, if we run it in debug mode and observe the produced logs, we
can see that a file update is detected and merged without the pause
and that the file is found up-to-date with the pause.

- [x] PR is not too big

### DIFF
--- a/core/local/atom/pause.js
+++ b/core/local/atom/pause.js
@@ -1,0 +1,67 @@
+/** This step bufferizes events while the watcher is paused.
+ *
+ * @module core/local/atom/pause
+ * @flow
+ */
+
+const STEP_NAME = 'pause'
+
+/*::
+import type Channel from './channel'
+import type EventEmitter from 'events'
+
+type WaitState = {
+  paused: boolean
+}
+*/
+
+module.exports = {
+  STEP_NAME,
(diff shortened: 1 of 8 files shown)


taratatach:
This approach raises other issues that are too big to be good.

PR #1891, opened 2020-05-29
core/local: Discard changes from remote propagation

taratatach:
When we propagate remote changes to the local filesystem, the local watcher will receive events from the filesystem and try to merge those "changes".
In some situations like when a Cozy Note was moved and updated, we end up reverting the changes we are applying.

We try to be smarter about the local changes in 2 ways:
- be more aggressive in the way we aggregate events in the Atom watcher to limit irrelevant changes
- try and detect already applied moves to that we don't merge creations when dispatching local `renamed` events with no source PouchDB records

- [x] PR is not too big

### DIFF
--- a/core/local/atom/await_write_finish.js
+++ b/core/local/atom/await_write_finish.js
@@ -65,7 +65,7 @@ function countFileWriteEvents(events /*: AtomEvent[] */) /*: number */ {
     }
     if (
       event.kind === 'file' &&
-      ['created', 'modified'].includes(event.action)
+      ['created', 'modified', 'renamed'].includes(event.action)
     ) {
       nbCandidates++
     }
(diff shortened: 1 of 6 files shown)


core/local/atom/await_write_finish.js
```diff
@@ -107,25 +107,35 @@ function isAggregationCandidate(event) {
...
+    }
   }
 
   if (recentEvent.action === 'modified') {
     addDebugInfo(recentEvent, oldEvent)
     // Preserve the action from the first event (it can be a created file)
     recentEvent.action = oldEvent.action
+
+    if (oldEvent.action === 'renamed') {
+      recentEvent.oldPath = oldEvent.oldPath
```
nono, line 137: I don't understand how this part works: do we have an `oldPath` on `modified` event that is used somewhere?

core/local/atom/await_write_finish.js
```diff
@@ -192,16 +202,28 @@ function debounce(waiting /*: WaitingItem[] */, events /*: AtomEvent[] */) {
               addDebugInfo(event, e)
               // Preserve the action from the first event (it can be a created file)
               event.action = e.action
+
+              if (e.action === 'renamed') {
+                event.oldPath = e.oldPath
```
nono, line 207: Same here

core/local/atom/await_write_finish.js
```diff
@@ -192,16 +202,28 @@ function debounce(waiting /*: WaitingItem[] */, events /*: AtomEvent[] */) {
...
+                log.debug(
+                  { createdEvent: e, deletedEvent: event },
+                  `Ignore ${e.kind} ${e.action} then ${event.action}`
+                )
+                events.splice(i, 1)
+                i--
+              } else if (e.action === 'renamed') {
+                addDebugInfo(event, e)
+                // Delete document at oldPath instead of moving then deleting
+                if (e.oldPath) {
```
nono, line 223: Why do we check that `e` has an `oldPath`? Can we have a `renamed` event without it?

taratatach: It's for Flow. I could make a check method that checks both if you find it clearer.

core/local/atom/await_write_finish.js
taratatach: The `recentEvent` action is changed on line 134 into `renamed`. In this case we want to keep the `oldPath` attribute.

core/local/atom/await_write_finish.js
taratatach: Same answer :)

core/local/atom/await_write_finish.js
nono: Does it mean that a `renamed` event can change the checksum of a file?

taratatach: Yes it can. This is the case already when we have to merge the offline edition and movement of a file.

PR #1895, opened 2020-06-08
core/merge: Remove Note metadata on conflicts

taratatach:
When we detect a local Cozy Note export modification and create a
conflict, we want to mark the new markdown file as such and not a real
Cozy Note.

We won't change the file's extension in case the file is still open in
the user's editor and so we won't change the document's mime type
either. However, we'll remove the Cozy Note's content and schema
stored in the record's `metadata` attribute since the actual file's
content has changed and won't be re-imported as a Note.

Removing these attributes should give Drive Web hints that the file
should be displayed as a markdown file.

- [x] PR is not too big

### DIFF
--- a/core/merge.js
+++ b/core/merge.js
@@ -187,9 +187,10 @@ class Merge {
       delete doc.overwrite
     }
 
-    // We create it at the destination location
+    // We create it at the destination location as a normal local file
     metadata.markAsNew(doc)
     metadata.dissociateRemote(doc)
+    metadata.removeNoteMetadata(doc)
     return this.addFileAsync('local', doc)
   }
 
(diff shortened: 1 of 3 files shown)


taratatach:
> I must be missing something. When reading the code, it looks like the metadata is deleted in PouchDB. But the commit message talks of Drive Web. Does this modification also deleted the metadata on the Cozy? I don't see code that take the metadata from PouchDB and send them to the Cozy.

We remove the metadata from the PouchDB record which will be saved as a new record. So we won't send the metadata to the remote Cozy while we would have otherwise.
My commit message is probably not clear enough.

taratatach:
I updated the message. It now states that those attributes won't be included in the new record and thus not sent to the Cozy.

PR #1896, opened 2020-06-08
core: Refetch remote Cozy Notes after upgrade

taratatach:
The new Desktop release fixes a bug which lead to local Cozy Note exports not being updated on Windows when propagating a remote note update and the creation of a conflict when detecting a local modification on local exports to avoid breaking the remote note.

The association of the 2 would lead to the generation of conflicts for all exports that became unsynchronized because of the bug on Windows.

We want to avoid those dispensable conflicts and propose to rely on the unapplied remote update detection to forcefully download and replace the local export with the remote version.
We add a migration that fakes a remote update on all Cozy Note PouchDB records so that `Merge` will detect them and choose to discard the local modification it will receive from the initial scan after the release install.

- [x] PR is not too big

### DIFF
--- a/core/pouch/migrations.js
+++ b/core/pouch/migrations.js
@@ -89,6 +89,32 @@ const migrations /*: Migration[] */ = [
         return doc
       })
     }
+  },
+  {
+    baseSchemaVersion: 2,
+    targetSchemaVersion: 3,
+    description: 'Marking Cozy Notes for refetch to avoid conflicts',
+    affectedDocs: (docs /*: Metadata[] */) /*: Metadata[] */ => {
+      return docs.filter(
+        doc =>
+          doc.mime === 'text/vnd.cozy.note+markdown' &&
+          doc.metadata &&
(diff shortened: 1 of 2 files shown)


core/pouch/migrations.js
```diff
@@ -89,6 +89,30 @@ const migrations /*: Migration[] */ = [
...
     }
+  },
+  {
+    baseSchemaVersion: 2,
+    targetSchemaVersion: 3,
+    description: 'Marking Cozy Notes for refetch to avoid conflicts',
+    affectedDocs: (docs /*: Metadata[] */) /*: Metadata[] */ => {
+      return docs.filter(
+        doc =>
+          doc.mime === 'text/vnd.cozy.note+markdown' &&
```
nono, line 100: Shouldn't we check for `doc.metadata.content`?

taratatach: This might be a bit safer indeed.

PR #1904, opened 2020-06-17
core/merge: Save local file metadata separately

taratatach:
We want to keep track of the current local state of files, whether
they're in sync or not.

At the moment, we store altogether in our PouchDB records the 3
different states of documents — local, remote and last synced — and a
few technical attributes that help us make sense of it all (e.g. which
side has received changes that need to be synchronized, what actions
need to be applied…).
This mechanism is limited because we don't store separately the values
of attributes on each side. For example, if a file's `size` has
changed on the remote Cozy and we've saved it in PouchDB, we've lost
the local file's `size`.
When loosing this information, it's harder for us to make sense of our
data and to make the right decision and even impossible to make some.

A good example of such an impossible decision to make can be described
by this scenario:
0. A file `a` is synchronized between the local filesystem and the
   remote Cozy
1. `a` is updated on the remote Cozy and thus its `md5sum` stored in
   PouchDB is changed
2. the client is stopped before the change is applied on the
   local filesystem

⇒ When the client is restarted, the local scan will generate an event
with the local, outdated, file `md5sum` which will be treated as an
update because we see a difference between the event's `md5sum` and
the one stored in PouchDB. We have no way to tell this was the
previous `md5sum` and this change should be discarded.

By storing the local state separately, we have this information and we
can use it to tell whether the detected local change is an actual one
or not by comparing the stored local state and the new one.

This commit only introduces local file states, the methods to update
them, to compare them with one another and the calls to keep them
up-to-date with the actual file's local state.

- [x] PR is not too big

### DIFF
--- a/core/metadata.js
+++ b/core/metadata.js
@@ -72,10 +72,34 @@ const CONFLICT_REGEXP = new RegExp(
   `-conflict-${DATE_REGEXP}${SEPARATOR_REGEXP}`
 )
 
+const LOCAL_ATTRIBUTES = [
+  'md5sum',
+  'class',
+  'docType',
+  'executable',
+  'updated_at',
+  'mime',
+  'size',
+  'ino',
+  'fileid'
+]
+
 /*::
 export type DocType =
   | "file"
   | "folder";
+
+export type MetadataLocalInfo = {
+  class?: string,
+  docType: DocType,
+  executable?: true,
+  fileid?: string,
+  ino?: number,
+  md5sum?: string,
+  mime?: string,
+  size?: number,
+  updated_at?: string,
+}
 
 export type MetadataRemoteInfo = {
   _id: string,
@@ -109,14 +133,15 @@ export type Metadata = {
   overwrite?: Metadata,
   childMove?: boolean,
   path: string,
+  local: MetadataLocalInfo,
   remote: MetadataRemoteInfo,
   size?: number,
   tags?: string[],
   sides: MetadataSidesInfo,
   trashed?: true,
   incompatibilities?: *,
-  ino?: ?number,
-  fileid?: ?string,
+  ino?: number,
+  fileid?: string,
   moveFrom?: Metadata,
   cozyMetadata?: Object,
   metadata?: Object
@@ -169,6 +194,7 @@ module.exports = {
   sameFolder,
   sameFile,
   sameFileIgnoreRev,
+  sameLocal,
   sameBinary,
   detectSingleSide,
   markSide,
@@ -181,7 +207,8 @@ module.exports = {
   outOfDateSide,
   createConflictingDoc,
   CONFLICT_REGEXP,
-  shouldIgnore
+  shouldIgnore,
+  updateLocal
 }
 
 function localDocType(remoteDocType /*: string */) /*: string */ {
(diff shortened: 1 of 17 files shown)


core/metadata.js
```diff
@@ -72,11 +72,35 @@ const CONFLICT_REGEXP = new RegExp(
   `-conflict-${DATE_REGEXP}${SEPARATOR_REGEXP}`
 )
 
+const LOCAL_ATTRIBUTES = [
```
nono, line 75: The  `name` is not kept here. What happens for this scenario?

1. A file `foo.txt` is synchronized
2. On the cozy, its content is updated
3. Cozy-desktop gets the changes feed
4. Before applying this change, cozy-desktop is stopped
5. The file is renamed locally
6. Cozy-desktop is restarted.

nono: And same question for the parent_id

core/metadata.js
```diff
@@ -72,11 +72,35 @@ const CONFLICT_REGEXP = new RegExp(
...
+  'md5sum',
+  'class',
+  'docType',
+  'executable',
+  'updated_at',
+  'mime',
+  'size',
+  'ino',
+  'fileid'
```
nono, line 84: What is the purpose of keeping `ino` and `fileid` here? In which case they will be different to the `ino` and `fileid` on the global document?

core/metadata.js
```diff
@@ -109,14 +133,15 @@ export type Metadata = {
   overwrite?: Metadata,
   childMove?: boolean,
   path: string,
+  local: MetadataLocalInfo,
```
nono, line 136: As I understand, `local` cannot be null or undefined (no `?`). So, why are you comparing it to `null` on https://github.com/cozy-labs/cozy-desktop/pull/1904/files#diff-6d18f9e7a792437c06b6f95186c97b88R322 ?

nono:
> This commit only introduces local file states, the methods to update
them, to compare them with one another and the calls to keep them
up-to-date with the actual file's local state.

I don't see how the `doc.local` is filled for new files or directories. Is it expected?

core/metadata.js
taratatach: The `name` or rather `path` for local files, will be introduced later. For now, I don't need it to solve some problems and keeping it in sync is just more work.

core/metadata.js
taratatach: It can actually be null, if the document is built by the remote watcher during a remote file addition.
However, marking it as potentially null or undefined means tons of unnecessary checks with Flow… I chose to go with the same pattern as for `remote` which can in fact be null while the type says it can't.

taratatach:
> > This commit only introduces local file states, the methods to update
> > them, to compare them with one another and the calls to keep them
> > up-to-date with the actual file's local state.
> 
> I don't see how the `doc.local` is filled for new files or directories. Is it expected?

It's a bit hidden. It's filled during calls to `prep.*` by the `metadata.buildFile()` method: https://github.com/cozy-labs/cozy-desktop/pull/1904/files#diff-b2fbff412cdb1b20ee3b7c797ef5b609R737

core/metadata.js
taratatach: They're probably not gonna be different but the goal is to move away from using the attributes of the global document so I'm adding them now.

PR #1911, opened 2020-06-30
core/remote/watcher: Keep pouch path normalization

taratatach:
To avoid detecting invalid movements when a file or folder whose path is encoded with different normalizations between the local PouchDB record and the remote Cozy is edited on the Cozy we do 2 things:
- compare changes' paths using the same normalization (NFC by default) every where within the remote watcher
- keep the existing (as in saved in PouchDB) path normalization for existing documents and parent folders

The later change is more of a safety net.

- [x] PR is not too big

### DIFF
--- a/core/remote/watcher/index.js
+++ b/core/remote/watcher/index.js
@@ -314,14 +314,14 @@ class RemoteWatcher {
           detail: `${docType} was created and trashed remotely`
         }
       }
+      const oldPath = was.path
       const previousMoveToSamePath = _.find(
         previousChanges,
         change =>
           (change.type === 'DescendantChange' ||
             change.type === 'FileMove' ||
             change.type === 'DirMove') &&
-          // $FlowFixMe
-          change.doc.path === was.path
+          metadata.samePath(change.doc, oldPath)
       )
 
       if (previousMoveToSamePath) {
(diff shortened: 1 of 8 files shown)


core/remote/watcher/index.js
```diff
@@ -13,6 +13,7 @@ const remoteChange = require('../change')
 const { handleCommonCozyErrors } = require('../cozy')
 const { inRemoteTrash } = require('../document')
 const squashMoves = require('./squashMoves')
+const normalizePaths = require('./normalizePaths')
```
nono, line 16: It goes against my intuition to have to normalize paths on the remote watcher, but I trust you

core/remote/watcher/index.js
```diff
@@ -366,10 +371,9 @@ class RemoteWatcher {
   async applyAll(
     changes /*: Array<RemoteChange> */
   ) /*: Promise<Array<{ change: RemoteChange, err: Error }>> */ {
-    const errors = await Promise.mapSeries(
-      changes,
-      async change => await this.apply(change)
-    ).filter(err => err)
+    const errors = await Promise.mapSeries(changes, this.apply).filter(
```
nono, line 374: Don't we need to bind `this` on `this.apply`?

core/remote/watcher/index.js
taratatach: I'm doing it to avoid changing the path saved in PouchDB. It might not be necessary but I thought it would be better to have all paths coherent with each other (e.g. an NFD encoded folder name should be NFD encoded in its children's paths).

I've made some changes and I'm only normalizing on macOS now since Windows and Linux see the same path encoded with 2 different normalizations as 2 different paths and can thus actually perform a move.

core/remote/watcher/index.js
taratatach: It feels like we should indeed but no tests are failing :thinking:. I'll dig a bit.

PR #1917, opened 2020-07-13
core/metadata: Use full mtime as updated_at

taratatach:
We'll now use the local content modification date (i.e. `mtime` field
of `Stats` objects) with its full precision as the new `updated_at`
value when propagating local changes to the remote Cozy instead of
using the max value between the content modification date and the
metadata modification date (i.e. `ctime` field), truncated to the
second.

We've decided to use the full `mtime` precision for the local state
`updated_at` attribute to be sure not to miss any file updates.
However, this presented us with a challenge in the test examples.
Indeed, we would sometimes have to compute this value from the main
`updated_at` attribute which could be equal to the local `ctime` value
and did not carry the milliseconds. We would thus have time differences
in our expectations and the examples would fail.

To avoid introducing hacks in the test files, we changed the way we
compute the main `updated_at`. We're now using the `mtime` value with
its full precision.
This change is in line with the experience we want to offer our users.
We believe the content modification time is the best time to use as
the "last modified" time which, in our opinion, should not be affected
by permission changes, movements, etc.

To keep the full precision of the date, we'll use the recent `updatedAt`
parameter in the `cozy-stack` API since the old `Date` header which 
expected a date serialized as a GMT datestring, missing the 
milliseconds.

- [x] PR is not too big


taratatach:
This is a pre-requisite for https://github.com/cozy-labs/cozy-desktop/pull/1661.

test/support/builders/metadata/base.js
```diff
@@ -158,22 +155,22 @@ module.exports = class BaseMetadataBuilder {
...
+    this.doc.updated_at = typeof date === 'string' ? date : date.toISOString()
     return this
   }
 
   newerThan(doc /*: Metadata */) /*: this */ {
-    this.doc.updated_at = timestamp
-      .fromDate(new Date(timestamp.fromDate(doc.updated_at).getTime() + 2000))
-      .toISOString()
+    this.doc.updated_at = new Date(
+      new Date(doc.updated_at).getTime() + 2000
```
y-lohse, line 165: I know it was there before but the 2000 looks like a magic number. I think it would be good to at least use a constant with a meaningful name for it.

test/unit/merge.js
```diff
@@ -2415,11 +2415,12 @@ describe('Merge', function() {
...
           _.defaults(
             {
+              _id: dstFile._id,
+              path: dstFile.path,
               sides: increasedSides(srcFile.sides, this.side, 1),
               moveFrom: movedSrcFile
             },
-            _.pick(srcFile, ['remote']),
-            _.omit(dstFile, ['_rev'])
+            _.omit(srcFile, ['_rev'])
```
y-lohse, line 2423: I don't understand how the changes in these files relate to the other changes in the PR?

taratatach: It rather seems like a simplification of the expectation and probably doesn't relate to the rest of the PR.
I'll try to move it to another PR.

core/remote/index.js
```diff
@@ -207,7 +209,8 @@ class Remote /*:: implements Reader, Writer */ {
     const options = {
       contentType: doc.mime,
       checksum: doc.md5sum,
-      lastModifiedDate: new Date(doc.updated_at),
+      updatedAt: doc.updated_at,
```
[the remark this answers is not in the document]
taratatach: Yes this is normal. This function sends a file update so we don't need to, nor should we, pass the `createdAt` date.

In any case, `cozy-stack` does not allow modifying the `created_at` attribute of remote documents from what I've seen.

core/utils/timestamp.js
```diff
@@ -60,3 +61,14 @@ function maxDate(d1 /*: Date */, d2 /*: Date */) /*: Date */ {
 function stringify(t /*: Timestamp */) {
   return t.toISOString().substring(0, 19) + 'Z'
 }
+
+function roundedRemoteDate(isoDate /*: string */) /*: string */ {
+  const date = new Date(isoDate)
+
+  const millisecond = isoDate.split(/\.|Z/)[1]
+  if (millisecond.length > 3) {
+    return new Date(date.getTime() + 1).toISOString()
```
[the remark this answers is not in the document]
taratatach: In this situation (i.e. more than 3 milliseconds digits) we need to "truncate" the date to fit our 3 milliseconds digits standard.
However, simply truncating is not an option because `cozy-stack` does not allow changing a Date attribute with an older value and truncating would mean, in most cases, getting a lower value (with the very rare exception of a 0 milliseconds date).

So we make sure we have a greater value that's close enough by adding 1 millisecond to the rounded date (hence the "rounded" part in the function name).

I can add a comment if you feel that's necessary.

test/support/builders/metadata/base.js
taratatach: t is a magic number but we don't really care about its value. I'm not sure using a constant would convey the idea that it really doesn't matter what the value is.

I can try with something like SOME_TIME if you think that's better but I'm not really convinced.

taratatach:
> Looking good overall, though I'm unsure about the `created_at` business. Especially the fact that we sometimes seem to force the `created_at` to be the same as `updated_at` — aren't we potentially overriding a more accurate creation date?

No the creation date is not set before that point in the Desktop client. It is only "forced" when we send a creation query to `cozy-stack`. At that point, the modification date is the closest date we have to the creation date. This will probably be modified later when we start using the birthtime available on some file systems (APFS for example).

If we don't set this date, the stack will set its value with the current date which is not what we want in a lot of cases. Besides, if the `creation date is more recent that the modification date, the stack will also set the modification date with the newly generated creation date to maintain consistency and we definitely don't want that.

> The commit messages where helpful to understand what's going on, so I wonder if some of them shouldn't be in the code for later reference?

Do you have specific messages in mind? I sometimes add some comments but not necessarily the whole story I can write in commit messages. The files would get a bit messy if I did.

core/utils/timestamp.js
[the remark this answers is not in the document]
y-lohse: Alright, yeah I definitely think a comment is need there 😁

test/support/builders/metadata/base.js
y-lohse: Could it describe why it's there (which i'm not super clear about either)? If I understand correctly it's an offset we use to force a document to be older or newer than another one?

test/unit/merge.js
y-lohse: imo a separate commit is fine :)

y-lohse:
> Do you have specific messages in mind? 

I was mostly thinking about the reason why we force a `createdAt` attributes (ie. your previous message), but there's no centralized place where it happens, so I'm not sure either how to do that in practical terms. Not knowing the codebase it wasn't obvious which calls were creations and which were updates for example. So maybe just something along the lines of "we force the creation date for files we create with the best we have"… somewhere?

test/support/builders/metadata/base.js
taratatach: Yes that's it. It's just a helper function to create PouchDB records with a newer modification date than another record.

taratatach:
> I was mostly thinking about the reason why we force a `createdAt` attributes 

I see I send the `createdAt` value for folder updates. This should not be the case and it might have tipped you off.
I'll remove this argument.

test/unit/merge.js
taratatach: Actually, this change is required. 

Before, we were expecting the final file record to keep all the attributes of the overwritten file (we're testing a move overwriting existing documents at the destination) besides the `remote` attribute (which holds the id and revision of a the associated remote document) which should be the one of the moved file.

This is not true anymore because we're not keeping the full precision of the modification time for the `updated_at` attribute (instead of removing the milliseconds) and this date is not the same for the overwritten file and the overwriting one.
The test was passing before only because the records were created within the same second so they had the same `updated_at` value.

taratatach:
@y-lohse I created a function to build the basic new document attributes, including the `updated_at` and `created_at` value so that we force the value in only place that could be commented.
I also added comments to the `roundedRemoteDate` method.

PR #1922, opened 2020-07-27
core/utils/timestamp: Round non-UTC remote dates

taratatach:
We should correctly round dates sent by the remote Cozy whether they're
in UTC format or not.

We recently started to round the `created_at` and `updated_at` dates
coming from the remote Cozy to make sure we have the expected ISO
format in our PouchDB records.
However, we were expecting to receive dates in UTC only while we can
receive dates in other timezones.
When the timezone is not UTC, our rounding function raises an error
after trying to read the `length` of `undefined`.

This error is raised because our method to get the milliseconds part
of the date as a string relied on the presence of the letter `Z` at
the end of the date string.
When the timezone is not UTC, the string ends with the time offset and
the milliseconds part could be `undefined` if milliseconds were not
present in the date.

We propose a new method based on a regular expression looking for the
presence of a dot followed by digits. If those are present then we
extract the digits as the milliseconds string. If they're not present,
we assign an empty string to the milliseconds.
In both cases, we can safely look at the `length` of the milliseconds.

- [x] PR is not too big

### DIFF
--- a/core/utils/timestamp.js
+++ b/core/utils/timestamp.js
@@ -77,8 +77,8 @@ function stringify(t /*: Timestamp */) {
 function roundedRemoteDate(isoDate /*: string */) /*: string */ {
   const date = new Date(isoDate)
 
-  const milliseconds = isoDate.split(/\.|Z/)[1]
-  if (milliseconds.length > 3) {
+  const [hasMilliseconds, milliseconds] = /\.(\d+)/.exec(isoDate) || [false, '']
+  if (hasMilliseconds && milliseconds.length > 3) {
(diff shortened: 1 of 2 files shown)


taratatach:
> Works for me, but can't we use [`getMilliseconds`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/getMilliseconds) instead of using regular expressions?

No we can't because go through the `Date` object directly. The stack uses a format with a nanoseconds precision which translates into 8 digits for the milliseconds part of the date string and the `Date` object truncates them after the third digit when we need to round it to the next millisecond to be sure the stack will accept further modifications :/.

PR #1929, opened 2020-08-06
gui/lastFiles: Securely persist last files list to disk

taratatach:
Node's documentation states that it is unsafe to make multiple calls
to `fs.writeFile()` on the same file in parallel
(see https://nodejs.org/api/fs.html#fs_fs_writefile_file_data_options_callback).

However, this is precisely what we were doing when persisting our
last files JSON object to disk. We only make one call to
`fs.writeFile()` per `lastFiles.persist()` call but calls to
`persist` are made asynchronously and not awaited by the caller (for
GUI performance).
This is probably the reason why we're seing malformed JSON files
persisted to disk.

We'll now use an `async` queue in which we'll push calls to
`fs.writeFile()` so they'll be run sequentially. We still won't
await calls to `lastFiles.persist()` in the GUI.

- [x] PR is not too big

### DIFF
--- a/gui/js/lastfiles.js
+++ b/gui/js/lastfiles.js
@@ -5,15 +5,22 @@
 
 const fs = require('fs')
 const path = require('path')
+const async = require('async')
 
 let lastFilesPath = ''
 let lastFiles = Promise.resolve([])
+let persistQueue = null
 
 const log = require('../../core/app').logger({
   component: 'GUI'
 })
 
+const writeJSON = (data, callback) => {
+  fs.writeFile(lastFilesPath, data, callback)
+}
+
 const init = desktop => {
+  persistQueue = Promise.promisifyAll(async.queue(writeJSON))
   lastFilesPath = path.join(desktop.basePath, 'last-files')
   lastFiles = new Promise(resolve => {
     fs.readFile(lastFilesPath, 'utf-8', (err, data) => {
@@ -34,11 +41,11 @@ const init = desktop => {
 
 const persist = async () => {
   const data = JSON.stringify(await lastFiles)
-  fs.writeFile(lastFilesPath, data, err => {
-    if (err) {
+  if (persistQueue) {
+    persistQueue.pushAsync(data).catch(err => {
       log.warn({ err }, 'Failed to persist last files')
-    }
-  })
+    })
+  }
 }
 
 const list = async () => await lastFiles
--- a/gui/js/lastfiles.js
+++ b/gui/js/lastfiles.js
@@ -13,7 +13,7 @@ const log = require('../../core/app').logger({
   component: 'GUI'
 })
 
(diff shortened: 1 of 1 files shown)


gui/js/lastfiles.js
```diff
@@ -5,15 +5,22 @@
...
   component: 'GUI'
 })
 
-module.exports.init = desktop => {
+const writeJSON = (data, callback) => {
+  fs.writeFile(lastFilesPath, data, callback)
+}
+
+const init = desktop => {
+  persistQueue = Promise.promisifyAll(async.queue(writeJSON))
```
Crash--, line 23: I'm wondering how it works since you're not returning the Promise in writeJSON ? :thinking:

taratatach: `writeJSON` passes the callback to `fs.writeFile` so the promise is resolved (or rejected) by `writeFile`?!

gui/js/lastfiles.js
```diff
@@ -5,15 +5,22 @@
...
 const log = require('../../core/app').logger({
   component: 'GUI'
 })
 
+const writeJSON = (data, callback) => {
+  fs.writeFile(lastFilesPath, data, callback)
+}
+
 const init = desktop => {
+  persistQueue = Promise.promisifyAll(async.queue(writeJSON))
```
y-lohse, line 23: Why are we using `promisifyAll` here?

taratatach: So that the `queue.push` call returns a promise (`queue.pushAsync` to be exact).
I'd prefer not to deal with callbacks.

y-lohse: I might be misreading it, but to me it sounds like `pushAsync` returns a promise anyway https://caolan.github.io/async/v3/docs.html#QueueObject ?

taratatach: We're using `async` v2.6.2 which does not have those promisified methods.
We should update the dependency at some point.

taratatach:
> LGTM! Adding a unit test might be good though :)

I tried to write a failing test but it would always pass. It's always hard to test race conditions and I'm not sure how to do it here.

gui/js/lastfiles.js
Crash--: How the promise is resolved or rejected is not my the point. 

Here `writeJSON` return `undefined` so how promisifyAll can know about waiting a promise?

y-lohse: That tripped me up too, but `async.queue` takes a function as param, and the function receives a node style `callback` that you call when you're done. Here we pass it to `writeFile` and it handles it from there ;)

gui/js/lastfiles.js
y-lohse: aha, gotcha!

PR #1932, opened 2020-08-08
gui: Update Recent list type icons

taratatach:
We want to match more types with a specific icon and avoid using the
generic `file` icon as much as possible.

We match 2 new types:
- files with the `.url` extension will use the `link` icon
- files matching the `vcard` mime type will use the `contact` icon

And we match more types with the `archive`, `code`, `presentation`,
`spreadsheet` and `audio` icons.

We also have a specific case for files with a missing mime type but 
it's not the first one tested while we do need the mime type for all 
other tests.
It seems cleaner and safer to make this check as soon as possible.

Finally, we have 2 cases which group multiple mime types together via
their `class` (i.e. the part before the slash in the mime type, like
`audio`).
Since most of our tests are matching specific parts of the full mime
type, we should run them before we check for those groups (in case one
specific type of these groups should be matched differently).
Moving them to the end of the list seems appropriate.

- [x] PR is not too big


gui/js/fileutils.js
```diff
@@ -4,32 +4,37 @@
  */
 
 module.exports.selectIcon = info => {
-  if (['image', 'video'].indexOf(info.class) !== -1) {
-    return info.class
-  } else if (info.class === 'music') {
-    return 'audio'
+  if (info.path.endsWith('url')) {
```
Crash--, line 7: Why not testing the class instead of just the end of the name?  class seems to be set by the stack only if the content is well formatted. 

```
"class": "shortcut",
  "mime": "application/shortcut"
```

Crash--: Also, on Drive Web we displayed the favicon of the target. I'm wondering if we should not do the same

gui/styles/_two_panes.styl
```diff
@@ -234,7 +234,7 @@
     background-size cover
     background-position center center
 
-    for mimetype in audio binary code file folder image pdf presentation spreadsheet text video archive cozy-note
+    for mimetype in archive audio binary code contact cozy-note file folder image link pdf presentation spreadsheet text video
         &.file-type-{mimetype}
             background-image url("./images/type-icons/icon-type-" + mimetype + ".svg")
```
Crash--, line 240: Openly, we should mutualize the file list between drive web and desktop... First step can be fro Desktop to use the SVG sprite of Cozy-UI. Like that, we would not have to add every new icons to desktop since we've them all in UI...

taratatach: Wouldn't the first step be to upgrade `cozy-ui` then? It's still probably a lot of work but this could be done after upgrading `electron`.
I found the same list of icons we have in Desktop in the Drive Web repo. Are those left overs?

gui/js/fileutils.js
taratatach: We use the `mime` library to get the mime types of files from their path and the returned type of `.url` files is `application/octet-stream`.
Since this is a very generic type I can't really use it here and I chose to check the extension directly.

I could also have moved this logic to our `mime` wrapper though, the same way I did for `.cozy-note`.

Crash--: What do we display in this "recent" panel? The modification on the local side? or the ones from the remote?

taratatach: We display files that were added, modified or moved and not removed from either side.

Crash--: So if we add a video without an internet connexion, we'll not have the `class` will not be set? So we display the generic `file` for an audio / video / image, right? https://github.com/cozy-labs/cozy-desktop/blob/681e360819d5b517eb024d1c0cee9cb8960c4885/gui/js/fileutils.js#L33-L37 

This is why you don't use `class === shortcut` to display the link icon?

gui/styles/_two_panes.styl
Crash--: > I found the same list of icons we have in Desktop in the Drive Web repo. Are those left overs?

Unfortunately, we still use them at one place we need to remove ;). 

Right for the UI upgrade. We should add this task somewhere in your roadmap ;)

gui/js/fileutils.js
taratatach: Hum no, we get the class. `mime` is a local dependency so we don't rely on an internet connection here.

Crash--: OK I got it. In fact Desktop can create the `class` attribute when the file is added locally (https://github.com/cozy-labs/cozy-desktop/blob/master/core/metadata.js#L763). 

What happen, when you send to the stack a file with an `application/octet-stream` mime and an `.url` extension? Do you get back the file with the `application/shortcut` and an `updatedAt` ... updated?

Crash--: Also, I'm wondering if you should not update your method to be closer to the stack https://github.com/cozy/cozy-stack/blob/master/model/vfs/vfs.go#L627-L661 

For instance, do we have the `slide` icon for a keynote document?

taratatach: So, I tested adding a shortcut in my local folder and checked both what we received back from the stack and what we saved in Pouch.
The stack does change the mime type from `application/octet-stream` to `application/internet-shortcut` but we don't save it in Pouch :/. I'll take a look at the reason behind this.

In any case it would indeed be good to align the mime/class modifications we make in Desktop and the stack. Also, I did not think about Apple's proprietary file types and they don't seem to be supported by `mime` so they're not detected as specific types but as generic files.

I'll create tickets for these two. In the meantime, my method works for all shortcuts created either on the local disk or the remote Cozy.

Crash--: > The stack does change the mime type from application/octet-stream to application/internet-shortcut

And in that case, does the stack update the updated_at attribute of the io.cozy.files? (just curious about the cycle of the file). (the most important part to me, is having that the io.cozy.files has the application/internet-shortcut mime on the cozy ;))

taratatach: The `updated_at` field hasn't changed. It might be the source of the issue.

PR #1939, opened 2020-08-24
pouch/migrations: Migrate changesfeed

taratatach:
When migrating an existing database, we copy all existing documents,
migrate those that need to be, replace their old version in the
database copy and then replace the old database with the copy.
In this process, we lose an essential information: the list of changes
from the old database's changesfeed that were not already processed by
the Sync module.

These changes can be found by requesting the changesfeed starting with
the saved local sequence. However, since the sequence increases with
each change made on a document and the new database has not seen the
same number of changes, the old database's local sequence has no
meaning in the new database context.
This is why we were just setting the new database's local sequence to
the last update sequence of the new database.

From there, only the changes made after the migration will have a
sequence greater than the local sequence and thus be found by the Sync
module.
In case some changes were merged but not applied (or synced) before a
migration was run, those won't appear in the changesfeed the next time
the Sync module looks for changes to apply and will be "lost".

We might apply those changes if another change on the same document is
merged later but we can't guarantee it will happen or that the result
will be the expected one.

For this reason, we now fetch the changes with a sequence greater than
the local sequence before applying a migration and touch their
associated documents after the migration has been run so that they
will appear in the changesfeed again.

- [x] PR is not too big

### DIFF
--- a/core/pouch/index.js
+++ b/core/pouch/index.js
@@ -471,6 +471,32 @@ class Pouch {
     return this.updater.push(task, callback)
   }
 
+  async unsyncedDocIds() {
+    const localSeq = await this.getLocalSeqAsync()
+    return new Promise((resolve, reject) => {
+      this.db
+        .changes({
+          since: localSeq,
+          filter: '_view',
+          view: 'byPath'
+        })
+        .on('complete', ({ results }) => resolve(results.map(r => r.id)))
+        .on('error', err => reject(err))
+    })
+  }
+
+  // Touch existing documents with the given ids to make sure they appear in the
+  // changesfeed.
+  // Careful: this will change their _rev value!
+  async touchDocs(ids /*: string[] */) {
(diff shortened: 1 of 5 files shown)


core/pouch/index.js
```diff
@@ -471,6 +471,30 @@ class Pouch {
...
+    })
+  }
+
+  // Touch existing documents with the given ids to make sure they appear in the
+  // changesfeed.
+  // Careful: this will change their _rev value!
+  async touchDocs(ids /*: string[] */) {
+    const existingDocs = await Promise.all(
+      ids.map(id => this.byIdMaybeAsync(id))
+    ).filter(d => d)
```
nono, line 494: I think it would be better to fetch the documents with a call to `allDocs` with a keys option: it makes a single call to pouchdb instead of a call per document.

nono (approved):

PR #1944, opened 2020-08-28
core/remote: Send body size with file modification

taratatach:
We're witnessing a lot of timeout errors (i.e. 504 response status)
when users send file modification requests while we don't really see
them for file creation requests.

We're not completely sure about the actual cause of those timeouts but
the `Content-Length` header is not sent in modification requests and
we believe this can have a role in this situation.

We're now sending the header and we'll monitor errors to see if we see
something different, either a drop in error rate or different errors
(we are expecting to receive 412 response statuses meaning the content
we're sending does not reflect the headers value and thus that the
content has changed on the local filesystem between the time the
record was merged and the time we stated propagating the change to the
remote Cozy).

- [x] PR is not too big

### DIFF
--- a/core/remote/index.js
+++ b/core/remote/index.js
@@ -208,6 +208,7 @@ class Remote /*:: implements Reader, Writer */ {
       {
         checksum: doc.md5sum,
         executable: doc.executable || false,
+        contentLength: doc.size,
         contentType: doc.mime,
         updatedAt: doc.updated_at,
         ifMatch: old && old.remote ? old.remote._rev : ''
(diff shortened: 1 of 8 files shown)


core/remote/index.js
```diff
@@ -200,16 +198,22 @@ class Remote /*:: implements Reader, Writer */ {
...
-      contentType: doc.mime,
-      checksum: doc.md5sum,
-      updatedAt: doc.updated_at,
-      executable: doc.executable || false,
-      ifMatch: ''
-    }
-    if (old && old.remote) {
-      options.ifMatch = old.remote._rev
-    }
+    // Object.assignn gives us the opportunity to enforce required options with
```
nono, line 201: s/assignn/assign/

nono (approved):

PR #1953, opened 2020-09-09
core/remote/watcher: Normalize path sorted changes

taratatach:
On macOS, we normalize the paths of remote changes to make sure we
keep the normalization of existing PouchDB records and avoid issues
like conflicts.

However, this process requires that ancestor paths be normalized
before their descendants' paths and remote changes are not necessarily
in this order (e.g. in case a folder is added with content and then
renamed, the folder addition change will come after its content
addition).

To prevent normalization issues, we now sort changes by destination
path before normalizing those paths.

It should not affect the final order of changes but this is a risk as
we know the remote changes sort method is not fully accurate.
Tests are passing so we hope we won't see new changes order issues in
the future.

- [x] PR is not too big

### DIFF
--- a/core/remote/change.js
+++ b/core/remote/change.js
@@ -82,7 +82,7 @@ export type RemoteDirTrashing = {
 export type RemoteIgnoredChange = {
   sideName: 'remote',
   type: 'IgnoredChange',
-  doc: Metadata|RemoteDoc|RemoteDeletion,
+  doc: *,
   was?: Metadata,
   detail: string
 }
@@ -139,7 +139,8 @@ module.exports = {
   isOnlyChildMove,
   includeDescendant,
   applyMoveInsideMove,
-  sort
+  sort,
+  sortByPath
 }
 
 const sideName = 'remote'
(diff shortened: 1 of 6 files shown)


core/remote/change.js
```diff
@@ -460,3 +461,22 @@ function sort(changes /*: Array<RemoteChange> */) /*: Array<RemoteChange> */ {
...
+    { doc: bDoc } /*: RemoteChange */
+  ) => {
+    if (!aDoc.path) return 1
+    if (!bDoc.path) return -1
+
+    const aPath = aDoc.path.normalize()
+    const bPath = bDoc.path.normalize()
+
+    if (aPath < bPath) return -1
+    if (bPath > aPath) return 1
```
nono, line 479: `aPath < bPath` and `bPath > aPath` are two ways to write the same condition, which means that the `return 1` here is dead code

nono (changes requested):

core/remote/change.js
taratatach: Good catch. I'll fix this right away.

nono (approved):

PR #1955, opened 2020-09-16
core/remote: Catch errors during file uploads

taratatach:
When a chunk encoded request, sent via HTTP/2 fails and the remote
Cozy returns an error (e.g. 413 because the file is too large),
Chromium will replace the response header status with a simple
`net:ERR_HTTP2_PROTOCOL_ERROR`, leaving us with no information about
the reason why the request failed.
See bugs.chromium.org/p/chromium/issues/detail?id=1033945

Besides, in this situation, Electron will reject a promise with a
`mojo result is not ok` error message in a way that does not let us
catch it.
See https://github.com/electron/electron/blob/1719f073c1c97c5b421194f9bf710509f4d464d5/shell/browser/api/electron_api_url_loader.cc#L190.

Our only recourse here is to make use of the Node's
`unhandledRejection` event handler to reject our own request and
make sure the client does not hang forever on this failed request.
To make sense of the situation, we also run checks in the event
handler on the remote Cozy to try and recreate the error that was
returned by the remote Cozy and take appropriate action down the
Sync process.

- [x] PR is not too big


package.json
```diff
@@ -89,11 +90,12 @@
     "btoa": "^1.1.2",
     "bunyan": "^2.0.2",
     "chokidar": "^3.2.2",
-    "cozy-client": "^11.2.1",
+    "cozy-client": "^14.6.0",
     "cozy-client-js": "^0.17.1",
+    "cozy-stack-client": "^14.1.2",
```
Crash--, line 95: no need to add it since you import the FetchError from cozy-client

Crash-- (approved):
After the removal of the dep 👍

package.json
taratatach: I need it for the linter :/

Crash--: OMG `const { FetchError } = require('cozy-stack-client')` my bad

PR #1959, opened 2020-09-28
Stop manually setting PouchDB records' `_id`

taratatach:
We stop manually setting the `_id` attribute of new PouchDB records
with a string computed from the record's `path` attribute. Ids will be
randomly generated by PouchDB itself and records will be fetched by
their `path` via the `byPath` view.

Although having ids based on documents' paths has some advantages like
being able to easily find a record in PouchDB from its path, it comes
with a big drawback when representing a directory's hierarchy: when a
document is moved or renamed, its id needs to change.
The main problem with this is the need to delete a record and create a
new one in a database that lacks transactions.

By not basing our records' ids on the documents' paths, we can change
their `path` attribute when the documents are moved or renamed and
thus change only one record.
We'll also be able to keep track of a given document via its record's
id throughout its life without the need to maintain those links.

- [ ] PR is not too big


core/pouch/migrations.js
```diff
@@ -216,6 +216,31 @@ const migrations /*: Migration[] */ = [
         return doc
       })
     }
+  },
+  {
+    baseSchemaVersion: 7,
+    targetSchemaVersion: 8,
+    description: 'Use generated ids instead of string based on path',
```
nono, line 223: Why do you need a migration? We can keep the old generated ids, even if we use them as an opaque value now.

core/pouch/migrations.js
```diff
@@ -216,6 +216,31 @@ const migrations /*: Migration[] */ = [
...
+    affectedDocs: (docs /*: Metadata[] */) /*: Metadata[] */ => {
+      return docs.filter(doc => !doc._deleted)
+    },
+    run: (docs /*: Metadata[] */) /*: Metadata[] */ => {
+      return docs.map(doc => {
+        if (doc.remote) {
+          // update remote with main object values?
```
nono, line 230: It looks like you started something, but forget it after.

core/merge.js
```diff
@@ -241,14 +245,16 @@ class Merge {
...
-    const file /*: ?Metadata */ = await this.pouch.byIdMaybe(doc._id)
+    const file /*: ?Metadata */ = await this.pouch.bySyncedPath(doc.path)
     metadata.markSide(side, doc, file)
     if (file && file.docType === 'folder') {
       throw new Error("Can't resolve this conflict!")
     }
     metadata.assignMaxDate(doc, file)
 
     if (file) {
+      // FIXME: $FlowFixMe use different types to manage records?
```
nono, line 256: It looks like a good idea

taratatach: And I've actually created it (https://github.com/cozy-labs/cozy-desktop/pull/1959/files/375579cf3ce69c02801f7490e5dee373a6fdc96c#diff-35b11e134833589d8cb2ae9238f6245ed0b49db0e2bbd8cc1221e4247de23ba1R32). I need to use it across the app though. I'll make a pass on the code base.

core/pouch/index.js
```diff
@@ -154,6 +161,9 @@ class Pouch {
       { path: doc.path, local, remote, _deleted: doc._deleted, doc },
       'Saving metadata...'
     )
+    // TODO: find something more useful than a generated uuid
+    // FIXME: $FlowFixMe find other types to manage records?
+    if (!doc._id) doc._id = uuid()
```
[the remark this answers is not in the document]
taratatach: The goal was to find something more useful than an `UUID` as _id but since I couldn't figure out something I might as well use `post()` indeed.

core/pouch/migrations.js
taratatach: ARf yes, I completely forgot the migration.

core/pouch/migrations.js
taratatach: Good idea. This would mean less records to migrate and less hassle to make sure they're all coherent.

core/pouch/index.js
```diff
@@ -145,26 +148,43 @@ class Pouch {
...
+    doc /*: T */
+  ) /*: Promise<SavedMetadata> */ {
     metadata.invariants(doc)
     const { local, remote } = doc.sides
     log.debug(
       { path: doc.path, local, remote, _deleted: doc._deleted, doc },
       'Saving metadata...'
     )
-    return this.db.put(doc)
+    // TODO: find something more useful than a generated uuid
```
nono, line 163: I think you can remove this comment

nono (approved):

PR #1963, opened 2020-10-05
core/metadata: Store remote doc in `Metadata`

taratatach:
The `remote` attribute of `Metadata` objects, stored locally in
PouchDB, only includes the `_id` and `_rev` of the associated remote
document.
We've started storing most of the other remote attributes in the main
section of the `Metadata` object but we're missing an opportunity to
make comparisons between the most recent remote state of a document
and the latest known remote state from any part of the synchronization
process.
We're forced to make those comparisons in the remote watcher while we
could benefit from them in the Merge or Sync modules.
Besides, storing the entire remote document means we have access to
all remote attributes and won't need to make further changes in the
future to access new attributes.

This commit only adds the entire remote document to `Metadata` objects
without making use of the newly stored attributes.

- [x] PR is not too big

### DIFF
--- a/core/remote/index.js
+++ b/core/remote/index.js
@@ -114,12 +114,12 @@ class Remote /*:: implements Reader, Writer */ {
     const parent /*: RemoteDoc */ = await this.remoteCozy.findDirectoryByPath(
       parentPath
     )
-    let dir /*: RemoteDoc */
 
     try {
-      dir = await this.remoteCozy.createDirectory(
+      const dir = await this.remoteCozy.createDirectory(
         newDocumentAttributes(name, parent._id, doc.updated_at)
       )
+      doc.remote = dir
     } catch (err) {
       if (err.status !== 409) {
         throw err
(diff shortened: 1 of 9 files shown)


core/remote/index.js
```diff
@@ -350,6 +330,9 @@ class Remote /*:: implements Reader, Writer */ {
         newRemoteDoc._id,
         referencedBy
       )
+      // addReferencedBy does not return a complete remote doc.
+      // Could it be an issue? We won't have the new reference in PouchDB and
+      // updating the _rev means the remote watcher won't do anything.
```
nono, line 335: I think you should clarify this before merging the PR

core/remote/index.js
```diff
@@ -392,6 +375,7 @@ class Remote /*:: implements Reader, Writer */ {
     }
   }
 
+  // TODO: change into assignNewRemote
```
nono, line 378: I think it should be better to fix this TODO now

nono (commented):
What is the migration path you have in mind?

taratatach:
Good calls. I did not review the commit before opening the PR and obviously I forgot to answer those questions when breaking it apart from the rest of my work on removing the path from the PouchDB _id.

taratatach:
About the migration, I'm kinda leaving it aside for now as I'm not making use of those attributes.
Clients should be migrating a lot of documents just by merging new changes and when we'll actually need them, we can fetch only the remote documents which have not been stored locally already and save some network and processing time.

nono (approved):

PR #1967, opened 2020-10-09
Update cozy-app-dev

nono:


taratatach:
I have a pending change which uses the tag `latest` instead of setting the version. What do you think about it?

nono:
@taratatach it is OK for me.

By the way, it looks like Travis build has failed because there was no space on device (https://travis-ci.org/github/cozy-labs/cozy-desktop/jobs/734280270#L5786). Maybe 10 MB is not enough on https://github.com/cozy-labs/cozy-desktop/blob/master/.travis.yml#L116.

taratatach:
Ah, good call! I thought this error was related to the Travis VM and thus cache. I'll increase the size of the HFS volume.

PR #1972, opened 2020-10-11
test/merge: Check local & remote attributes' value

taratatach:
Now that we're maintaining `local` and `remote` metadata separately we
should check that we have the expected value when calling different
Merge methods.

This meant improving the way we build and maintain those attributes in
the data builders as well.


taratatach:
@nono I need to keep going but if you find the time to review this PR later, I'm still interested in your thoughts about it.

PR #1976, opened 2020-10-13
core/metadata: Require `executable` file attribute

taratatach:
The `executable` attribute of `Metadata` objects is marked as optional
in the type definition. This leads to uncertainty as to whether the
attribute is defined in `Metadata` objects or not and forces us to
check for its presence every time we want to read or set it. Besides,
it allows us to build file `Metadata` objects without it.

Marking the attribute as required and setting it for every file
`Metadata` object will help us ensure our documents are valid and
simplify the use of this attribute.

/!\ Although it is marked as required, it will still be missing in
folder `Metadata` objects. This is made possible by cracks in our
types coverage but should be fixed once we'll have different type
definitions for file and folder metadata.

- [x] PR is not too big

### DIFF
--- a/core/pouch/migrations.js
+++ b/core/pouch/migrations.js
@@ -197,6 +197,25 @@ const migrations /*: Migration[] */ = [
         return doc
       })
     }
+  },
+  {
+    baseSchemaVersion: 7,
+    targetSchemaVersion: 8,
+    description: 'Set all files executable attribute',
+    affectedDocs: (docs /*: Metadata[] */) /*: Metadata[] */ => {
+      return docs.filter(
+        doc => doc.docType === 'file' && doc.executable == null
+      )
+    },
+    run: (docs /*: Metadata[] */) /*: Metadata[] */ => {
+      return docs.map(doc => {
+        doc.executable = false
+        if (doc.local && doc.local.executable == null) {
+          doc.local.executable = false
+        }
+        return doc
+      })
+    }
   }
 ]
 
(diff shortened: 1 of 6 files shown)


core/pouch/migrations.js
```diff
@@ -197,6 +197,25 @@ const migrations /*: Migration[] */ = [
...
       })
     }
+  },
+  {
+    baseSchemaVersion: 7,
+    targetSchemaVersion: 8,
+    description: 'Set all files executable attribute',
+    affectedDocs: (docs /*: Metadata[] */) /*: Metadata[] */ => {
+      return docs.filter(
+        doc => doc.docType === 'file' && doc.executable == null
```
nono, line 207: You are sure that we want to compare with `null`, and not with `undefined`?

core/pouch/migrations.js
```diff
@@ -197,6 +197,25 @@ const migrations /*: Migration[] */ = [
...
+      )
+    },
+    run: (docs /*: Metadata[] */) /*: Metadata[] */ => {
+      return docs.map(doc => {
+        doc.executable = false
+        if (doc.local && doc.local.executable == null) {
```
nono, line 213: Idem

core/pouch/migrations.js
taratatach: Using `==` both `null` and `undefined` values will return true here. I could specify `undefined` instead but if by any chance we do have some `null` values, I want to change them to false as well.

nono (approved):

PR #1979, opened 2020-10-14
core/pouch: Allow fetching docs via their `path`

taratatach:
We add the possibility to fetch documents from PouchDB via their
`path` attribute. This is done via a new `Pouch.bySyncedPath()`
method and a change in the `byPath` view.

The view will now emit the normalized parent path and name of each
document — both computed from the document's `path` — instead of the
parent path alone. This allows us to query documents via their full
path in addition to their parent path (thus preserving the
`Pouch.byPath()` and `Pouch.byRecursivePath()` querying methods).

We introduce a new querying method, `Pouch.bySyncedPath()` which will
return the first document matching the given path.

Being able to query documents by their `path` is a pre-requisite to
generating PouchDB record ids independant from the documents' paths.

- [x] PR is not too big

### DIFF
--- a/core/pouch/index.js
+++ b/core/pouch/index.js
@@ -221,11 +221,36 @@ class Pouch {
 
   // Return all the files and folders in this path, only at first level
   byPath(basePath) {
+    const key =
+      basePath === '' ? metadata.id(basePath) : metadata.id(basePath) + path.sep
     const params = {
-      key: basePath === '' ? basePath : basePath + path.sep,
+      startkey: [key, ''],
+      endkey: [key, '\ufff0'],
       include_docs: true
     }
     return this.getAll('byPath', params)
+  }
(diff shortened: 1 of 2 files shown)


core/pouch/index.js
```diff
@@ -288,17 +314,36 @@ class Pouch {
...
+            normalized = doc.path.toUpperCase()
+            break;
+          default:
+            normalized = doc.path
+        }
+        const parts = normalized.split(${sep})
+        const name = parts.pop()
+        const parentPath = parts.concat('').join(${sep})
+
+        return emit([parentPath, name], { rev: doc._rev })
```
nono, line 346: By curiosity, why do you emit `{ rev: doc._rev }`?

nono (approved):

PR #1989, opened 2020-11-03
ci: Use Github Actions instead of Travis

taratatach:
Allow testing and building the application on Github Actions instead of Travis CI.

- [x] PR is not too big

### DIFF
--- a/.github/workflows/linux.yaml
+++ b/.github/workflows/linux.yaml
@@ -1,0 +1,201 @@
+name: Linux
+
+# To debug a job, add the following step:
+# - name: Debug with tmate
+#   uses: mxschmitt/action-tmate@v3
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    paths-ignore:
+      - '.github/workflows/macos.yaml'
+
+env:
+  COZY_DESKTOP_DIR: "${{ github.workspace }}/cozy-desktop"
+  COZY_STACK_STORAGE: "${{ github.workspace }}/storage"
+  COZY_URL: "http://localhost:8080"
+  COZY_PASSPHRASE: "cozy"
+  GO111MODULE: "on"
+  NODE_ENV: "test"
+  COZY_DESKTOP_HEARTBEAT: "1000"
+  DISPLAY: ":99.0"
+  NODE_VERSION: "12.14.1"
+  GO_VERSION: "1.14"
+  COUCHDB_VERSION: "2.3.1"
+
+jobs:
+  unit:
+    runs-on: ubuntu-latest
+    name: Lint & Unit tests
+    steps:
+      - uses: actions/checkout@v2
+      - name: Cache node modules
+        id: cache-node-modules
+        uses: actions/cache@v2
+        env:
+          cache-name: cache-node-modules
+        with:
+          path: |
+            ${{ github.workspace }}/node_modules
+            ~/.cache/electron
+            ~/.cache/electron-builder
+          key: ${{ runner.os }}-v${{ env.NODE_VERSION }}-${{ env.cache-name }}-${{ hashFiles('**/yarn.lock') }}
+      - name: Setup node
+        uses: actions/setup-node@v2-beta
+        with:
+          node-version: ${{ env.NODE_VERSION }}
+          check-latest: true
+      - name: Setup golang
+        uses: actions/setup-go@v2
+        with:
+          go-version: ${{ env.GO_VERSION }}
+      - name: Setup CouchDB
+        uses: 'cobot/couchdb-action@11e2c9f532a0eee2516ff80d92990d8081f2792c' # v4
+        with:
+          couchdb version: ${{ env.COUCHDB_VERSION }}
+      - name: Setup cozy-stack
+        uses: ./.github/actions/setup-cozy-stack
+      - name: Setup local env
+        run: |
+          mkdir -p "$COZY_DESKTOP_DIR"
+          echo "NODE_ENV=test" > "${{ github.workspace }}/.env.test"
+      - name: Install dependencies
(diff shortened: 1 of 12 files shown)


.github/workflows/linux.yaml
```diff
@@ -0,0 +1,185 @@
...
+
+env:
+  COZY_DESKTOP_DIR: "${{ github.workspace }}/cozy-desktop"
+  COZY_STACK_STORAGE: "${{ github.workspace }}/storage"
+  COZY_URL: "http://localhost:8080"
+  COZY_PASSPHRASE: "cozy"
+  GOPROXY: "https://proxy.golang.org"
+  GO111MODULE: "on"
+  NODE_ENV: "test"
+  COZY_DESKTOP_HEARTBEAT: ${{ 1000 }}
```
nono, line 23: Why `${{ 1000 }}` and not just `"1000"`?

.github/workflows/linux.yaml
```diff
@@ -0,0 +1,185 @@
...
+  pull_request:
+    paths-ignore:
+      - '.github/workflows/macos.yaml'
+
+env:
+  COZY_DESKTOP_DIR: "${{ github.workspace }}/cozy-desktop"
+  COZY_STACK_STORAGE: "${{ github.workspace }}/storage"
+  COZY_URL: "http://localhost:8080"
+  COZY_PASSPHRASE: "cozy"
+  GOPROXY: "https://proxy.golang.org"
```
nono, line 20: I don't think the `GOPROXY` env variable is required

.github/workflows/linux.yaml
```diff
@@ -0,0 +1,185 @@
...
+    steps:
+      - uses: actions/checkout@v2
+      - name: Cache node modules
+        id: cache-node-modules
+        uses: actions/cache@v2
+        env:
+          cache-name: cache-node-modules
+        with:
+          path: ./node_modules
+          key: ${{ runner.os }}-build-${{ env.cache-name }}-${{ hashFiles('**/yarn.lock') }}
```
nono, line 87: You don't have the node version in the cache key, is it something we should add?

.github/workflows/macos.yaml
```diff
@@ -0,0 +1,203 @@
...
+env:
+  COZY_DESKTOP_DIR: "${{ github.workspace }}/cozy-desktop"
+  COZY_STACK_STORAGE: "${{ github.workspace }}/storage"
+  COZY_URL: "http://localhost:8080"
+  COZY_PASSPHRASE: "cozy"
+  GOPROXY: "https://proxy.golang.org"
+  GO111MODULE: "on"
+  NO_BREAKPOINTS: ${{ 1 }}
+  NODE_ENV: "test"
+  NPM_CONFIG_PROGRESS: ${{ false }}
```
nono, line 24: You have `NO_BREAKPOINTS` and `NPM_CONFIG_PROGRESS` on macOS but not on Linux. Is it expected?

.github/workflows/macos.yaml
```diff
@@ -0,0 +1,203 @@
...
+  COZY_DESKTOP_HEARTBEAT: ${{ 1000 }}
+  DISPLAY: ":99.0"
+
+jobs:
+  unit:
+    runs-on: macos-latest
+    name: Unit tests
+    strategy:
+      matrix:
+        fs: ['APFS']
```
nono, line 34: Is is expected to have only `APFS` here?

.github/workflows/macos.yaml
```diff
@@ -0,0 +1,203 @@
...
+        run: sudo Xvfb :99 -ac -screen 0 1024x768x8 &
+      - name: Unit tests
+        run: yarn test:unit
+
+  integration:
+    runs-on: macos-latest
+    name: Integration tests
+    strategy:
+      matrix:
+        fs: ['APFS']
```
nono, line 81: Idem

.github/workflows/macos.yaml
taratatach: It is expected for `NO_BREAKPOINTS` since it's related to Chokidar scenarios. It's not for the other variable but I should completely remove it as we use `yarn` instead of `npm`.

.github/workflows/linux.yaml
taratatach: That's taken directly from our Travis config. You added it yourself to speed up Travis builds. Do you think that's not relevant anymore?

.github/workflows/linux.yaml
taratatach: I used this syntax when I thought there might be a difference between string and number variables in GA. I should clean that up indeed.

.github/workflows/macos.yaml
taratatach: Yes it's expected. Unit and integration tests are not yet ready to be run on a HFS+ partition (mainly because some expectations need to differ).

.github/workflows/macos.yaml
taratatach: See https://github.com/cozy-labs/cozy-desktop/pull/1989#discussion_r532745561

.github/workflows/linux.yaml
taratatach: That's a good point. We probably should include it since some dependencies need to be recompiled when changing the Node version.

.github/workflows/linux.yaml
nono: I think it is not relevant anymore, but we should test to be sure.

taratatach: Removing the variable doesn't seem to change the build time. I'll get rid of it then.

nono (approved):
LGTM but there is still one red build on macOS

taratatach:
@nono yes, integration tests fail on my own mac with the same segfault error :/. I haven't found the reason yet.

PR #2005, opened 2021-01-08
core/merge: Accept trashing of already trashed doc

taratatach:
When we try to merge a doc trashing and that doc was already trashed
or deleted on the other side, we don't do anything as it looks like we
are already in sync.
However, this is ain't quite true. The sides metadata have changed and
since we use a deletion marker now, Sync will try to apply the first
deletion.
If the first deletion was done on the local PC, we'll try to trash a
doc that has since changed on the Cozy and we'll get a 412 error.
Now that we're blocking the Sync on this error, this is quite visible
and annoying (we can still get out of this situation by giving up on
the change).

By simply accepting the merge the second trashing request, we make the
remote revision will be updated when the remote watcher runs and the
synchronization will resume after a retry (either automatic or
manual).

Note: the situation is mostly the same for documents that were erased
from the Cozy (i.e. the Cozy trash was emptied after the documents
were deleted on the Cozy). It differs by the Merge methods called.

- [x] PR is not too big

### DIFF
--- a/core/merge.js
+++ b/core/merge.js
@@ -315,6 +315,8 @@ class Merge {
               metadata.markSide('remote', file, file)
               return this.pouch.put(file)
             } else {
+              // TODO: should we save the new metadata anyway to make sure we
+              // have up-to-date side infos?
               return
             }
           } else {
@@ -331,6 +333,8 @@ class Merge {
               metadata.markSide('local', file, file)
               return this.pouch.put(file)
             } else {
+              // TODO: should we save the new metadata anyway to make sure we
+              // have up-to-date side infos?
               return
             }
           }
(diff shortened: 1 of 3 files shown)


core/merge.js
```diff
@@ -275,6 +275,7 @@ class Merge {
           }
           return this.pouch.put(file)
         } else {
+          // TODO: nécessité de sauvegarder quand même le document ??
```
nono, line 278: You have some TODOs in french. Is it normal?

nono (approved):

core/merge.js
taratatach: Nope! Good catch. I'll rewrite it in English.

PR #2013, opened 2021-01-14
gui: Improve user action messages readability

taratatach:
A few improvements to the user actions displayed in the main window
mainly to improve their readability:
- adding room to display their content
- making document names stand out of the error message
- grouping action buttons
- adding a shortcut to display the affected document in the file
  manager

- [x] PR is not too big
  Split the message content in a different place.


gui/elm/Window/Tray/Dashboard.elm
```diff
@@ -296,3 +299,53 @@ removeCurrentAction model =
...
+
+
+decoratedName : Regex.Match -> Html Msg
+decoratedName match =
+    match
+        |> .submatches
+        |> List.head
+        |> Maybe.withDefault (Just "")
+        |> Maybe.map viewName
+        |> Maybe.withDefault (text "")
```
nono, line 332: Do we really need two `Maybe.withDefault`? I think we can remove the first one

nono (approved):

gui/elm/Window/Tray/Dashboard.elm
taratatach: We need both:
1. `.submatches` returns a `List (Maybe String)`
2. so `List.head` returns a `Maybe (Maybe String)`

The first one could be removed if `viewName` would take a `Maybe String` rather than a `String` as parameter but I'd prefer to avoid dealing with maybes in this function.

taratatach: I rewrote it with a switch case as I can pattern match `Just (Just str)` instead of using `Maybe.withDefault` twice.

PR #2021, opened 2021-01-27
core/sync: Retry failed changes that block Sync

taratatach:
When we failed to synchronize a change, we would block the Sync
process and subsequently check periodically if the change could be
synchronized via dedicated methods (one for each Sync error code).

This has a major flaw though: the dedicated method must always check
for all possible causes of the synchronization failures and ways to
fix them.
For example, a file upload to the remote Cozy could fail if the Cozy
is out of disk space and the situation can be solved in 3 different
ways:
- by removing files on the Cozy
- by increasing the Cozy disk space (by purchasing some)
- by removing files on the local disk, especially the file we're
  trying to Sync

We were not checking for the third solution and thus the user could be
stuck without understanding why if they ever chose it.

To simplify our retry mechanism, we decided to stop using those
dedicated methods and instead go through the whole Sync process again.
This way, any change made on the remote Cozy or the local disk will be
picked up and can influence the synchronization (or not).

- [x] PR is not too big

### DIFF
--- a/core/app.js
+++ b/core/app.js
@@ -334,7 +334,6 @@ class App {
       this.ignore,
       this.events
     )
-    this.sync.diskUsage = this.diskUsage
   }
 
   // Start the synchronization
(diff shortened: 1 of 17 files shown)


nono (approved):
Let's say it is OK (but I don't really know)

PR #2032, opened 2021-02-11
core/remote: Remove _type attribute from doc types

taratatach:
The `_type` attribute is used by `cozy-stack` to store the doctype of
CouchDB records. In our case, this doctype is always `io.cozy.files`
and since it is not stored in the CouchDB records themselves, the data
we receive from our calls to the remote changes feed for example does
not contain this attribute.

We don't need this attribute and should not expect its presence in
returned remote records via our Flow types nor add it ourselves since
it makes it harder to compare remote data in tests with docs fetched
from our helpers.

- [x] PR is not too big

### DIFF
--- a/core/remote/cozy.js
+++ b/core/remote/cozy.js
@@ -7,7 +7,6 @@ const autoBind = require('auto-bind')
 const OldCozyClient = require('cozy-client-js').Client
 const CozyClient = require('cozy-client').default
 const { FetchError } = require('cozy-stack-client')
-const _ = require('lodash')
 const path = require('path')
 const {
   AbortController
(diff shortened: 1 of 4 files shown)


nono:
> The _type attribute is used by cozy-stack to store the doctype of
CouchDB records. In our case, this doctype is always io.cozy.files
and it seems the data we receive from our calls to the stack via
cozy-client-js does not contain this attribute.

The `_type` attribute is not stored in CouchDB, so, if you query the changes feed, it is normal that the response has no `_type`.

nono (approved):

taratatach:
> The `_type` attribute is not stored in CouchDB, so, if you query the changes feed, it is normal that the response has no `_type`.

OK, it makes sense. I don't understand why we were adding it then.

PR #2039, opened 2021-02-19
core/local/atom: Fetch deleted local inode ASAP

taratatach:
When `@atom/watcher` sends us a `deleted` event, it does not contain
the inode (or fileid) of the deleted document. Besides, we cannot get
it from the filesystem as it does not exist anymore.

Until now we were fetching the supposedly deleted inode from PouchDB
(via the document's path) only when needed. However, in some
situations like the application of a remote move on the local
filesystem on Windows (on Windows we detect moves by grouping
`deleted` and `created` events), it might be too late to get the
inode of the moved document from PouchDB (i.e. it has been erased
already by Sync after the change application has completed).

To mitigate this situation we now fetch the inode of locally deleted
documents as soon as we get the event in case it would be modified or
moved in PouchDB before we actually need it.

- [x] PR is not too big

### DIFF
--- a/core/local/atom/add_infos.js
+++ b/core/local/atom/add_infos.js
@@ -59,21 +59,26 @@ function loop(
               path.join(opts.syncPath, event.path)
             )
           }
+
           if (event.stats) {
             // created, modified, renamed, scan
             event.kind = stater.kind(event.stats)
-          } else {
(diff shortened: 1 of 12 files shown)


nono (approved):
LGTM

PR #2040, opened 2021-02-19
core/local/atom: Fetch from PouchDB via local path

taratatach:
In a number of the local Atom watcher, we try to fetch the PouchDB
record associated with the events we're processing. We use for that
the events' paths as the record key.

However, the record itself has 3 different path values:
- a local path
- a remote path
- a "main" path, modified by Merge each time it's changed either
  locally or remotely (i.e. the document was renamed or moved)

This means that fetching the record from the local watcher via its
main path can return unexpected results (the wrong record, more than
one records or no records at all) if the document was moved or renamed
remotely and this information was merged but not fully applied.
We're usually in this situation when we're actually applying the
remote path change to the local filesystem and processing the events
fired by this application.

In some cases, it is crucial that those events are processed correctly
to not detect and propagate to the remote Cozy changes that were not
done by the user (e.g. when a file was moved then replaced or edited
on the remote Cozy, we could end up trashing the new doc on the Cozy).

By matching the local events' paths with the local path or PouchDB
records, we make sure the appropriate record is returned in these
situations if fetched before it is updated by Sync after the remote
change was successfully applied on the filesystem. This shouldn't be a
problem as the steps needing this kind of information come first (e.g.
`add_infos` which needs the record to find the inode of `deleted`
events).
We use the same request across the other steps to make sure we stay
coherent.

- [x] PR is not too big

### DIFF
--- a/core/pouch/index.js
+++ b/core/pouch/index.js
@@ -309,6 +309,22 @@ class Pouch {
     return matches.length ? matches[0] : undefined
   }
 
+  async byLocalPath(fpath) /*: Promise<?SavedMetadata> */ {
+    if (!fpath) {
+      return undefined
+    }
+
+    const params = {
+      key: byPathKey(fpath),
+      include_docs: true
+    }
+    const matches = await this.getAll('byLocalPath', params)
+    // TODO: Do we need to handle cases in which we have more than one match?
+    // This should probably not happen if we handle correctly id conflicts on
+    // Windows and macOS.
+    return matches.length ? matches[0] : undefined
+  }
+
   // Return all the files and folders in this path, even in subfolders
   async byRecursivePath(
     basePath,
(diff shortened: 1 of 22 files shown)


core/pouch/index.js
```diff
@@ -415,6 +436,41 @@ class Pouch {
...
+  // - the parent path
+  // - the document's name
+  // This allows us to either fetch documents via their full path or their
+  // parent path, recursively or not.
+  // The parent path of a document in the root folder will be '', not '.' as
+  // with Node's path.dirname() result.
+  async addByLocalPathView() {
+    const platform = JSON.stringify(process.platform)
+    const sep = JSON.stringify(path.sep)
+    const query = `function(doc) {
```
nono, line 451: I would have put the normalization per platform outside of the templated string:

```js
let normalized
switch (process.platform) {
case 'darwin':
    normalized = "doc.local.path.normalize('NFD').toUpperCase()"
    break;
case 'win32':
    normalized = "doc.local.path.toUpperCase()"
    break;
default:
    normalized = "doc.local.path"
}
const query = `function(doc) {
if ('local' in doc && 'path' in doc.local) {
    let normalized = ${normalized}
    // ...
}`
```

nono (approved):
It makes sense

core/pouch/index.js
taratatach: It could work. I'll try it out.

nono (approved):
I'm not sure to fully understand this new commit, but let's say it is OK.

PR #2043, opened 2021-02-26
core/sync: Handle 4xx remote errors

taratatach:
We specifically handle 4xx remote errors (i.e. Fetch errors with a 
status code between 400 and 499) so that they won't show up as an 
unreachable Cozy to the user.

When we can, we provide new errors messages to help the user figure out
how to fix the issue and resume the synchronization.

Some errors will be sent to Sentry so we can try to solve them before
too many users encounter them.

- [ ] PR is not too big

### DIFF
--- a/core/remote/errors.js
+++ b/core/remote/errors.js
@@ -182,11 +182,21 @@ const wrapError = (err /*: FetchError |  Error */) /*: RemoteError */ => {
           err
         })
       case 412:
-        return new RemoteError({
-          code: NEEDS_REMOTE_MERGE_CODE,
-          message: 'Known remote document revision is outdated',
-          err
-        })
+        if (sourceParameter(err) === 'If-Match') {
+          // Revision error
+          return new RemoteError({
+            code: NEEDS_REMOTE_MERGE_CODE,
+            message: 'The known remote document revision is outdated',
+            err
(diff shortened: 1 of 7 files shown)


core/remote/errors.js
```diff
@@ -171,42 +188,111 @@ const wrapError = (err /*: FetchError |  Error */) /*: RemoteError */ => {
...
-        })
+        if (status > 400 && status < 500) {
+          return new RemoteError({
+            code: UNKNOWN_INVALID_DATA_ERROR_CODE,
+            message:
+              'The data sent to the remote Cozy is invalid for some unhandled reason',
+            err
+          })
+        } else {
+          // TODO: Merge with UnreachableError?!
```
nono, line 251: `TODO` spotted

nono (approved):

core/remote/errors.js
taratatach: Yes indeed. But `UnreachableError` is only used when opening a Cozy Note and using a `RemoteError` instead means I also have to deal with `CozyDocumentMissingError` which should not be in the remote errors module…
I think this PR is large enough and we need to get this code out to fix users' problems so I'll deal with the refactoring later :).

PR #2103, opened 2021-06-01
gui: Disable Electron's HTTP cache

taratatach:
We continue experiencing network errors (e.g. `net::ERR_FAILED`) for
no apparent reasons even though we added a cache busting parameter to
every request.

We try a different approach by disabling Electron's HTTP cache at the
command line level.

We found this solution here: https://stackoverflow.com/questions/63562516/electron-throws-err-failed-when-making-requests-using-net

- [x] PR is not too big


taratatach:
This problem seems to be solved.

PR #2105, opened 2021-06-25
core/remote/cozy: Skip deleted docs on inital sync

taratatach:
The remote changes feed contains every change that was ever made on a
Cozy and also changes made on indexes (they're documents as well).
This means that a Cozy that has seen a lot of changes, even for
completely deleted documents, will have a very large changes feed and
thus will require many requests during the initial synchronization of
a new Desktop client.

To avoid losing time in fetching unnecessary documents and in hope of
decreasing the I/O consumption during this phase, we implement a
specific initial fetch that will only fetch the latest version of
remote documents and avoid downloading deleted ones.

- [x] PR is not too big

### DIFF
--- a/core/remote/cozy.js
+++ b/core/remote/cozy.js
@@ -7,6 +7,7 @@ const autoBind = require('auto-bind')
 const OldCozyClient = require('cozy-client-js').Client
 const CozyClient = require('cozy-client').default
 const { FetchError } = require('cozy-stack-client')
+const { Q } = require('cozy-client')
 const path = require('path')
 
 const { FILES_DOCTYPE, FILE_TYPE, DIR_TYPE } = require('./constants')
@@ -49,9 +50,8 @@ export type Reference = {
 }
 
 type ChangesFeedResponse = {|
-  pending: number,
   last_seq: string,
-  results: Array<{ doc: RemoteDoc|RemoteDeletion }>
+  remoteDocs: Array<RemoteDoc|RemoteDeletion>
 |}
 */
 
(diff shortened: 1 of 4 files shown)


core/remote/cozy.js
```diff
@@ -526,13 +529,83 @@ async function getChangesFeed(
...
+    } else {
+      const rows = await client.data.findAll(FILES_DOCTYPE, {
+        include_docs: true,
+        limit: batchSize,
+        startkey_docid: startDocId
+      })
+      const filteredDocs = filterDocs(rows)
+      // FIXME: why less than 2??? Use descriptive constant instead.
+      if (filteredDocs.length < 2) {
+        return docs.map(d => ({ doc: d }))
```
paultranvan, line 589: It's kind of a pity having to iterate over 10K docs just add the `doc` key :/ Is it really necessary? If yes, it should be more efficient to do it when we filter the docs after the fetch, so we only iterate once?

paultranvan: Note we also iterate over all the docs in [cozy-client-js](https://github.com/cozy/cozy-client-js/blob/4daaf6bb186cf17a51c672fca287b7aa63997880/src/data.js#L107), so it might be even better to fetch manually, just like [here](https://github.com/cozy/cozy-client/blob/812c763048b32568947e6e6c96757bea4409ef05/packages/cozy-pouch-link/src/remote.js#L10)

core/remote/cozy.js
```diff
@@ -526,13 +529,83 @@ async function getChangesFeed(
...
+      filteredDocs.shift() // Remove first element, already included in previous request
+      startDocId = filteredDocs[filteredDocs.length - 1]._id
+      docs = docs.concat(filteredDocs)
+
+      if (rows.length < batchSize) {
+        hasMore = false
+      }
+    }
+  }
+  return docs.map(d => ({ doc: d }))
```
paultranvan, line 600: Same remark about the additional iteration

PR #2118, opened 2021-07-29
core/merge: Handle moves on the same Pouch record

taratatach:
Historically, document moves were merged in PouchDB by erasing the
record with the source path and creating a new record with the
destination path as record _ids were mapped to the path.

This way of doing things creates a number of difficulties (e.g.
inconsistencies if one record can be modified/created and not the
other as PouchDB does not offer transactions) and is not necessary
anymore since we've moved to generated _ids.

This change should give us opportunities in the future like linking
documents to their parent directory via its PouchDB record _id instead
of using a path attribute that can change or point to a different
document.

The main difficulty was managing overwrites as we were previously
using the overwritten record's _id as the destination record _id and
can't anymore as the destination record _id is now the same as the
source record _id.
Therefore we now erase the overwritten record from the database, its
data being saved in the destination record's `overwrite` attribute.
We should be on lookout for potential issues arising from this change.

- [ ] PR is not too big

### DIFF
--- a/core/merge.js
+++ b/core/merge.js
@@ -88,42 +88,39 @@ class Merge {
   // the user's editor and renaming it could create issues (e.g. the editor does
   // not detect the renaming and we'd create a new conflict each time the open
   // file would be saved).
-  async resolveNoteConflict(doc /*: Metadata */, was /*: ?SavedMetadata */) {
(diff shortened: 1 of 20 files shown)


nono (approved):
I've waited for this change for so long. Nice :+1:

PR #2119, opened 2021-07-30
 core/sync: Prioritize specific changes over others

taratatach:
We know that we can't rely on the order provided by PouchDB's
changesfeed to synchronize changes in the order they were initially
performed because the feed keeps only 1 version of any given record.

To mitigate this, we introduce a dependency graph which allows us to
express dependency relations between changes independent from their
sequence in the feed.
The goal is to create as little dependencies as possible and keep
propagating changes in the order given by the feed though as we still
rely on the sequence number of propagated changes to decide where to
start fetching from the feed.

This means we also need to make sure marking a change as propagated
(i.e. marking its sequence number as the new starting point) does not
prevent propagating changes that would come afterwards.
To this end, we reschedule changes (i.e. update their records without
making any changes to increase their sequence in the feed) whenever a
prioritized change has a greater sequence number than the next change
to propagate.

- [ ] PR is not too big


core/sync/dependency_graph.js
```diff
@@ -0,0 +1,236 @@
...
+  }
+
+  // Expand the graph into an ordered list of Node objects and then map this
+  // list to a list of Change objects which can be used by Sync to propagate
+  // them.
+  toArray() /*: Change[] */ {
+    const orderedChanges = new Set()
+
+    for (const node of this.nodes) {
+      for (const dependency of orderedDependencies(node)) {
```
nono, line 165: What are the performances expectation? The algorithm has a pretty bad "worst case" with long chain of dependencies. Maybe we don't care, but I would like your opinion on that.

core/sync/index.js
```diff
@@ -64,13 +72,140 @@ const shouldAttemptRetry = (change /*: MetadataChange */) => {
...
+        return -1
+    }
+    if (opA.type === 'ADD' && opB.type === 'MOVE') {
+      // Handle child addtion after parent move
+      if (docA.path.startsWith(docB.path + sep)) return 1
+    }
+    if (opB.type === 'ADD' && opA.type === 'MOVE') {
+      // Handle child addtion after parent move
+      if (docB.path.startsWith(docA.path + sep)) return -1
+    }
```
nono, line 192: You don't have condition when both `opA` and `opB` are `'Add'` to ensure that the parent is created before the child? I think of this scenario:

1. the cozy-desktop is offline
2. we create the folder `A` on the cozy
3. we create the folder `B` on the cozy inside `A`
4. we rename the folder `A` to `C`
5. the desktop is now back online, and will see the creation of `C/B` before the creation of `C`

nono: And what about a file moved (opA) to a directory that is renamed (opB)?

core/sync/index.js
```diff
@@ -231,28 +364,217 @@ class Sync {
...
+              )
+              // Solve 3.
+              // We resolve the conflict on the remote Cozy to avoid local
+              // issues like permission issues on Windows.
+              await this.remote.resolveRemoteConflict(change.doc)
+            }
+            break
+          case remoteErrors.DOCUMENT_IN_TRASH_CODE:
+            delete change.doc.moveFrom
+            delete change.doc.overwrite
```
nono, line 499: [nit] I would have kept the `overwrite` in case trashing the file was a mistake and the user restore the file just after. We probably don't really care as it would be very rare to have the good conditions and timing to happen, but if keeping the `overwrite` doesn't cost us anything, I still think it would be more "correct".

test/scenarios/move_overwriting_dir/not_trashed/scenario.js
```diff
@@ -46,24 +44,36 @@ module.exports = ({
       'dst/dir/subdir/subsub/',
       'src/'
     ],
-    trash:
-      process.platform !== 'darwin' && runWithStoppedClient()
+    localTrash: [
+      // Why isn't the hierarchy preserved in the local trash on Windows?
```
nono, line 48: Is it something you want to fix in this PR?

taratatach: Probably not. I guess it's related to the fact that remote changes are fetched together and handled differently from the Windows local changes way.
Since the result is the same on all platforms and other platforms get this result when propagating local changes this does not seem too important to fix now.

core/sync/index.js
taratatach: I'm not sure to follow. If we remove the `moveFrom` attribute then the `overwrite` attribute will never be used by the Writers so I don't see any good reason to keep it.
On the other hand, I'm pretty sure keeping it could lead to unexpected behaviors.

core/sync/dependency_graph.js
```diff
@@ -0,0 +1,236 @@
...
+  }
+}
+
+// Expand the sub-graph represented by the given node and its dependencies into
+// an ordered list of nodes based on these dependencies which are directed
+// edges.
+function orderedDependencies(node /*: Node */) /*: Set<Node> */ {
+  const nodeDependencies = new Set()
+
+  if (node.dependentOn.length > 0) {
```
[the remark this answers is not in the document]
taratatach: You're right, we don't need it. I'll remove it.

core/sync/index.js
taratatach: The first scenario (i.e. both operations are `ADD`) is covered by the same operation conditional block: https://github.com/cozy-labs/cozy-desktop/pull/2119/files#diff-7a081eeb18a2cc3146ed8c339c9b93edf2feafda1c0c53e1a7c03c53d89aaf7eR162

The second scenario will also enter the same operation conditional block and if the file is moved to the directory after it has been renamed then we should propagate the folder renaming first.
Otherwise the order will stay unchanged.

core/sync/dependency_graph.js
taratatach: I haven't tried to propagate a large number of changes (I will though) but I don't expect great performances as the complexity of the algorithm isn't so good.
I don't expect our users to have lots of dependencies between their changes though and if it should happen, I prefer to make sure we can handle them even if it takes some time to sort them out (I don't expect the expansion to take more time than the synchronization itself).

taratatach: I have revisited the expansion algorithm to make sure each node is visited only once and remove a lot of loops that were unnecessary.
This makes the algorithm way easier to understand as well.

PR #2120, opened 2021-08-02
core/utils/fs: Check for doc presence when handling trash errors

taratatach:
The error message thrown by Electron's `shell.trashItem()` method is
localized on macOS. Therefore, we can't detect missing document error
by matching the error message against a crafted RegExp as this would
mean building a RegExp that matches against every possible language.

e.g. when deleting missing `dir/file.txt`, the message would be:
- in English, `The file « file.txt » doesn't exist`
- in French, `Le fichier « file.txt » n'existe pas`

The only way to be sure `shell.trashItem()` failed because the
document does not exist anymore is to check whether the document
exists or not (minus some potential race condition issues).
This is slightly less efficient but works on all platforms for
every language.

- [x] PR is not too big

### DIFF
--- a/core/utils/fs.js
+++ b/core/utils/fs.js
@@ -5,6 +5,7 @@
 const Promise = require('bluebird')
 const childProcess = require('child_process')
 const { shell } = require('electron')
+const fse = require('fs-extra')
 
 const logger = require('./logger')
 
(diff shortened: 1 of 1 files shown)


core/utils/fs.js
```diff
@@ -41,19 +42,15 @@ async function sendToTrash(fullpath /*: string */) {
   try {
     await shell.trashItem(fullpath)
   } catch (err) {
-    if (
-      err.message === 'Failed to move item to trash' || // Error message on Linux
-      err.message === 'Failed to parse path' || // Error message on Windows
-      /doesn’t exist/.test(err.message) // Error message on macOS
-    ) {
```
nono, line 48: I don't understand. The commit message says that the detection mechanism is changed only for macOS, but it looks like two lines for Linux and Windows are removed too.

taratatach: Right. This fix is mainly for macOS as the detection works fine on Linux and Windows but since the behavior will be different on all OS I should update the message.

taratatach: Also, I forgot to update the PR's message when I updated the commit's message. It must have been really confusing indeed.

PR #2128, opened 2021-08-11
core/remote: Find remote parents in local Pouch

taratatach:
When we're looking for a directory by path (usually the parent
directory of another document), we make a request to the remote Cozy,
even if the directory probably hasn't changed since the last time we
requested it, just to get its `_id` and use it as the `dir_id` of its
direct children.

This is inefficient both for the Desktop client and the remote Cozy
and slows down the entire synchronization.
This is inefficient because we probably already have the information
we're looking for in the local PouchDB.

Besides, when we look for the parent directory of a document during
its synchronization, we should reason about the information on which
previous steps like Merge were based on and changed.
For example, when adding a file to a directory on the local
filesystem, we should be looking for the remote _id of the folder
whose local path is the parent of our file and not for a remote folder
with the same path.

As a corollary of this change, we'll now throw an error when
propagating an action on a directory results in some kind of conflict
and won't link different directories together anymore.
Those errors will be caught by Sync and we can then decide what's the
best course of action to solve the problem.

- [x] PR is not too big

### DIFF
--- a/core/remote/index.js
+++ b/core/remote/index.js
@@ -7,7 +7,6 @@
 const autoBind = require('auto-bind')
 const Promise = require('bluebird')
 const path = require('path')
-const { posix, sep } = path
 
 const { isNote } = require('../utils/notes')
 const logger = require('../utils/logger')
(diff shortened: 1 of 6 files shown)


core/remote/index.js
```diff
@@ -404,13 +382,34 @@ class Remote /*:: implements Reader, Writer */ {
...
 
     const results = await this.remoteCozy.search({ dir_id: dirID, name })
     if (results.length > 0) return results[0]
   }
 
+  async findDirectoryByPath(
+    path /*: string */
+  ) /*: Promise<MetadataRemoteDir> */ {
+    // TODO: store a local copy of the root dir to limit requests to the Cozy?
+    if (path === '.') return this.remoteCozy.findDir(ROOT_DIR_ID)
```
nono, line 396: I don't understand why we need to get the root directory. Can't we recreate this document here? We know the id, we don't care of the revision, it is a folder, etc.

taratatach: I didn't want to hard code it but it's true that it's probably not gonna change much. And we would avoid quite some requests.

PR #2136, opened 2021-08-19
core/local/chokidar: Fix analysis defaultSorter

taratatach:
The `defaultSorter()` sort method used after the initial scan by the
Chokidar watcher to sort changes before merging them is supposed to
sort deletion changes before addition changes on the same path.

However, the way we compared those changes was flawed and in the
case the deletion change would come before the addition change
before sorting, our algorithm would switch places between them (i.e.
the "deleted path" of the addition change would be null and equal to
the "added path" of the deletion change).

We now make sure only non-null and normalized paths can be deemed
equal.

- [x] PR is not too big

### DIFF
--- a/core/local/chokidar/local_change.js
+++ b/core/local/chokidar/local_change.js
@@ -40,6 +40,7 @@ module.exports = {
   addPath,
   delPath,
   updatePath,
+  samePath,
   childOf,
   lower,
   isChildDelete,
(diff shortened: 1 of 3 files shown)


core/local/chokidar/local_change.js
```diff
@@ -245,6 +246,9 @@ function delPath(a /*: LocalChange */) /*: ?string */ {
 function updatePath(a /*: LocalChange */) /*: ?string */ {
   return isUpdate(a) ? a.path.normalize() : null
 }
+function samePath(p1 /*: ?string */, p2 /*: ?string */) /*: boolean */ {
```
Merkur39, line 249: Original ce "typage" 😁  👍

Merkur39 (approved):

core/local/chokidar/local_change.js
taratatach: J'aurais pu appeler la fonction `maybeSamePath` mais ça faisait un peu long pour pas grand chose ici.

PR #2138, opened 2021-08-20
core/sync: Manage excluded dir conflict errors

taratatach:
When a remote directory is excluded from a client's synchronization,
we lose track of it (or don't create it in the first place) in our
local PouchDB.
In this situation, if the user creates a local document with the same
path as an excluded remote directory, we will end up with a name
conflict when trying to synchronize the local document with the remote
Cozy.

We create here a special error for when the locally created document
is a directory to help the user understand the reason behind this
conflict and offer them to either rename the local directory with a
conflict suffix or link it with the excluded directory which will then
re-included into the synchronization.
For conflicts with moved directories or with files, we show a normal
conflict error offering only to rename the local document.

We also a script to help developers manage their client's exclusions
until a GUI is available to do it.

- [ ] PR is not too big


core/sync/index.js
```diff
@@ -676,6 +641,32 @@ class Sync {
...
+        since: seq,
+        filter: '_view',
+        view: 'byPath',
+        return_docs: false,
+        include_docs: true
+      }
+      this.pouch.db
+        .changes(opts)
+        .on('change', ({ doc }) => {
+          if (doc.path === expectedPath) resolve()
```
nono, line 665: I think we should add `else reject()` on this line. We call the changes feed with `limit: 1`, so if this change doesn't match, we won't have other changes, and the promise will never be resolved/rejected without the `else` branch.

taratatach: It's a `live` subscription to the changes feed so every change will be notified. The `limit: 1` option here makes sure we only receive on change at a time.

test/integration/differential_sync.js
```diff
@@ -0,0 +1,282 @@
...
+      .name('Photos')
+      .create()
+    remoteFile = await builders
+      .remoteFile()
+      .inDir(remoteDir)
+      .name('IMG_001.jpg')
+      .data('image')
+      .create()
+
+    // FIXME: This needs to be called after a first request to the remote Cozy.
```
nono, line 82: FIXME spotted

nono (approved):

PR #2152, opened 2021-09-21
core/sync: Skip trashing content of missing folder

taratatach:
Since v3.30.0, we make sure to synchronize the deletion of folders
before the deletion of their content to preserve their hierarchy in
the Trash.
However, an error is now notified to the user for each item of the
folder and blocks the synchronization until the remote trashing is
fetched and merged by the remote watcher.

Skipping the synchronization of the content deletion was done by
looking at the parent folder's record in PouchDB and checking if it
was marked for deletion but when the folder trashing is synchronized,
its record is erased from PouchDB.
We'll now also skip synchronizing its content deletion if its record
is missing from PouchDB.

- [ ] PR is not too big


nono (approved):
No idea, let's say it is OK for me

PR #2174, opened 2021-11-15
gui: Add partial sync configuration button

taratatach:
We add a button in the `Preferences` tab of the main window to open
the partial synchronization configuration for the Desktop's OAuth
client within the remote Cozy Settings app.

It will be opened in the user's own Web browser for 2 reasons:
1. it clearly shows the configuration can be done from the Settings
   Web app and thus any computer
2. the user is more likely to be logged into their Cozy in their own
   browser and thus not be required to log in again to configure the
   Desktop's partial synchronization

The button will only be shown if the
`settings.partial-desktop-sync.show-synced-folders-selection` flag is
enabled (either in the local configuration or on the remote Cozy).

- [x] PR is not too big


gui/ports.js
```diff
@@ -80,9 +80,23 @@ elmectron.ports.chooseFolder.subscribe(() => {
...
+    const partialSyncEnabled =
+      flags['settings.partial-desktop-sync.show-synced-folders-selection']
+    elmectron.ports.syncConfig.send({
+      address,
+      deviceName,
+      deviceId,
+      capabilities,
+      flags: {
+        partialSyncEnabled:
+          partialSyncEnabled != null ? partialSyncEnabled : true
```
nono, line 95: `true` as the default value?

nono (approved):

gui/ports.js
taratatach: Yes. The idea is to be able to remove the flag later without disabling the feature. The corrolary is that we'll have to set the flag for every context before releasing this code to make sure it's not enable by default.

nono: Good luck setting the flag for the self-hosted ;-)

Maybe you should start with `false`, and change the default value in a future release.

taratatach: Or I make sure the Settings app has the same behavior so they get selective synchronization by default?!

nono: Yes, but you need to release settings and cozy-desktop at the same time, which is what you wanted to avoid in the first place, no?

taratatach: Yes, that's why I want to use the flag. I don't see how I can avoid the release issue for self hosted users though. Do you have any idea?

taratatach: Maybe we could use the capabilities for that?

taratatach: Actually, I can see a way that's actually gonna be needed: define the minimum Desktop version for the Settings to show the configuration button.
But that means knowing in advance which version of the Desktop will be able to do selective synchronization.

nono: 1. We can make a release of cozy-desktop, with `false` as the default value here (version `x.y.z`)
2. We release settings, with partial sync enabled if at least one of those conditions is true:
    - the flag is activated
    - the cozy-desktop client is at least on version `x.y.z+1`
3. We make a new release of cozy-desktop that just swaps the default value to `true` (version `x.y.z+1`).

What do you think?

taratatach: That works only if we don't have to make an intermediate Desktop release without partial sync. That's why setting the expected version in advance is not fool proof.

taratatach: I'm gonna set the default value to `false` until we figure out the right way to plan the release.

PR #2191, opened 2022-01-19
gui: Clean on-boarding url without domains list

taratatach:
To help our users connect their Desktop client to their Cozy, we try
to handle as many URL formats as possible (e.g. a Cozy app URL instead
of the canonical one).
However, since Cozy can be self-hosted and can be configured to use
either nested or flat sub-domains, we have to know some details to
clean up all the possible user inputs.

We were previously using our own domain, `mycozy.cloud` as that
specific detail helping us determine how to clean up the URL but since
we now have partners with other domains and self-hosted Cozy can also
use flat sub-domains, we've decided to use a hard-coded list of know
core Cozy apps as the way to clean up the URL.

  This means that all our partners in the future will benefit from the
  cleanup but we'll have to maintain the list of apps when we publish
  new ones.

- [ ] PR is not too big


nono (approved):
I really hope that we won't have a case where a self-hosted use `drive.mydomain.com` as the URL for their Cozy.

taratatach:
@nono I've updated the implementation. It nos uses the new `cozy-client` `rootCozyUrl` helper instead of the hard-coded app names list.

nono (approved):

PR #2204, opened 2022-02-28
core/lochokidar/normalize_paths: Normalize only when needed

taratatach:
HFS+ filesystems automatically re-encode utf-8 characters of document
names using the NFD norm when they were encoded with the NFC norm in
the first place. On the other hand, most filesystems and,
specifically, the remote Cozy filesystem, encode utf-8 characters with
the NFC norm.
This means that documents created on these systems will be
automatically renamed when Desktop creates them on the local HFS+
filesystem, generating unnecessary name changes.

We implemented a re-normalization process to try and keep the previous
normalization of document names in our local PouchDB database but we
did not account for cases where both norms coexist in the same
document name. In these cases, the re-normalization wouldn't work
(because we could not detect the one norm used) and, worse, conflicts
were created.

We're now implementing a simpler version of the re-normalization which
consists of reusing the name of a document as it exists in PouchDB if
only its normalization has changed. When documents are renamed, we
won't try to save the new name using the old normalization anymore
since the name is changing anyway.
We shouldn't really have to deal with specific norms anymore with this
process.

- [x] PR is not too big

### DIFF
--- a/core/local/chokidar/normalize_paths.js
+++ b/core/local/chokidar/normalize_paths.js
@@ -36,19 +36,20 @@ const step = async (
       const parentPath = path.dirname(c.path)
       const parent =
         parentPath !== '.' ? await pouch.bySyncedPath(parentPath) : null
-      c.path = normalizedPath(
+      const normalized = normalizedPath(
         c.path,
         c.old ? c.old.path : undefined,
         parent,
         normalizedPaths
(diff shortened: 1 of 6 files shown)


nono (approved):
I don't know if it is a good idea or not. Let's try!

PR #2208, opened 2022-03-07
core/remote: Use most recent remote date

taratatach:
In some cases, at least when importing images from a mobile device via
  Drive Mobile's synchronization, we can have files with a creation date
  (`created_at` field) that is more recent than the latest modification
  date (`updated_at` field).

  This should be impossible since `cozy-stack` forbids these situations
  but we do see them in production.
  As a consequence, further metadata updates on the file, such as a
  move, will be rejected by `cozy-stack` due to the "invalid
  modification date".

  Since the user cannot fix this problem easily, we don't want to stop
  the Desktop synchronization when we encounter such files.
  This is why we'll use for the `updated_at` field value, in every
  request sent to `cozy-stack`, the most recent date between the
  following:
  - the local last modification date
  - the remote last modification date
  - the remote creation date

- [ ] PR is not too big


core/remote/index.js
```diff
@@ -506,8 +506,18 @@ function newDocumentAttributes(
...
+  const remoteCreationDate = doc.remote && doc.remote.created_at
+  const remoteModificationDate = doc.remote && doc.remote.updated_at
+  const remoteDate =
+    remoteCreationDate && remoteModificationDate
+      ? timestamp.maxDate(remoteCreationDate, remoteModificationDate)
+      : remoteModificationDate
+      ? remoteModificationDate
+      : remoteCreationDate
+      ? remoteCreationDate
+      : null
```
nono, line 518: I find this way to structure code hard to read. I would have made something like:

```js
let date = doc.updated_at

const remoteCreationDate = doc.remote && doc.remote.created_at
if (remoteCreationDate) {
  date = timestamp.maxDate(date, remoteCreationDate)
}

const remoteModificationDate = doc.remote && doc.remote.updated_at
if (remoteModificationDate) {
  date = timestamp.maxDate(date, remoteModificationDate)
}

return date
```

But you version works too. It is just a question of style.

nono (approved):

core/remote/index.js
taratatach: Your version is probably more readable indeed. I think I'll rewrite this part.

PR #2268, opened 2022-11-15
fix: Avoid PouchDB conflicts when merging moves

taratatach:
When a document being synchronized with the remote Cozy is moved on
the local filesystem, we fail to merge the move in PouchDB because we
fetch the source record before it is updated by Sync (i.e. to add the
remote info) and get a conflict error since we don't have the new
revision.

Fetching the source record (again) right before merging the move, when
Merge has the PouchDB lock, solves this problem.

This should not create new issues as Merge should be able to handle
potential situations where changes to the source record are more
substantial than adding remote metadata.

- [x] PR is not too big

### DIFF
--- a/core/local/chokidar/send_to_prep.js
+++ b/core/local/chokidar/send_to_prep.js
@@ -44,33 +44,30 @@ const onAddFile = (
   { path: filePath, stats, md5sum } /*: LocalFileAddition */,
   prep /*: Prep */
 ) => {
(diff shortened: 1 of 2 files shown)


taratatach:
I could not figure out a way to build an automated test since it is very time sensitive and our test helpers do not give enough control here.
I tested it manually though.

nono (approved):

PR #2269, opened 2022-11-25
fix: Force cozy-client-js stack version

taratatach:
When instanciating cozy-client-js, it requests the instance's
`/status` route to determine if it is talking to an old Cozy v2
(nodezy) or to a Cozy v3 (gozy).

Since there are no Cozy v2 anymore and Desktop is incompatible with
the obsolete nodezy, we'll force the instance version to 3 to avoid
a useless request to the status route.

- [ ] PR is not too big


sblaisot (changes requested):
I think the same change should be added to 

- `core/remote/registration.js:91`
- `dev/remote/change-dir-exclusions.js:140`
- `test/support/helpers/cozy.js:61`
- `test/unit/remote/cozy.js:860`
- `test/unit/remote/watcher.js:81`

sblaisot (approved):

PR #2278, opened 2023-02-08
chore: Migrate transifex config from v2 to v3

cballevre:
Transifex stops February 12 their API v2. Therefore, we are migrating the configuration to v3. If you were using a [transifex-client](https://github.com/transifex/transifex-client), you also must move to [Transifex CLI](https://developers.transifex.com/docs/cli) to be compatible with the v3 API.

### DIFF
--- a/.tx/config
+++ b/.tx/config
@@ -1,8 +1,8 @@
 [main]
 host = https://www.transifex.com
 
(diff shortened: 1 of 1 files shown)


cballevre:
🎉 🎉 🎉

PR #2298, opened 2023-04-28
feat: Fetch remote changes upon realtime events 

taratatach:
We now subscribe to `created`, `updated` and `deleted` realtime events
on `io.cozy.files` documents to detect as soon as possible when
documents were modified on the remote Cozy.

When we received one of these events, we'll enqueue a remote watcher
run request and use our classic mechanism to fetch and merge remote
changes.

Since fetching and merging remote changes consumes time and resources,
we'll limit the number of requests enqueued via a debounce mechanism:
- the first and last events will enqueue requests
- an event will be considered as the last of a group if we haven't
  received new events in the last 200 ms
- we'll enqueue a request at least every 5 s to avoid long idle
  periods when receiving lots of events (keep in mind that only one
  request is kept in the queue at any time)

- [ ] PR is not too big


gui/js/network/agent.js
```diff
@@ -0,0 +1,207 @@
...
+		// configured for a request. Defaults to a new `https.Agent()`
+		// instance with the proxy agent options passed in.
+		httpsAgent?: http.Agent;
+
+		// A callback for dynamic provision of proxy for url.
+		// Defaults to standard proxy environment variables,
+		// see https://www.npmjs.com/package/proxy-from-env for details
+		getProxyForUrl?: GetProxyForUrlCallback;
+	};
+*/
```
Peltoche, line 68: Should be removed?

gui/js/network/agent.js
```diff
@@ -0,0 +1,207 @@
...
+      log.debug('Proxy not enabled for URL: %o', url)
+      return secureEndpoint ? this.httpsAgent : this.httpAgent
+    }
+
+    log.debug('Request URL: %o', url)
+    log.debug('Proxy URL: %o', proxy)
+
+    // attempt to get a cached `http.Agent` instance first
+    const cacheKey = `${protocol}+${proxy}`
+    let agent = null //FIXME: this.cache.get(cacheKey)
```
Peltoche, line 146: FIXME

Peltoche (approved):

gui/js/tray.js
```diff
@@ -149,13 +115,57 @@ const init = (app, listener) => {
...
   }
+
   setStatus('idle')
 }
 
 const wasInitiated = () => {
   return tray != null
 }
 
+// old tray menu
```
nono, line 126: Why did you keep this commented code?

core/remote/watcher/realtime_manager.js
```diff
@@ -0,0 +1,160 @@
...
+    this.reconnectTimeout = null
+    this.realtimeLogsAdded = false
+
+    autoBind(this)
+  }
+
+  async start() {
+    log.debug('Subscribing to realtime events...')
+    try {
+      const realtime = await this.remoteCozy.realtime()
```
nono, line 67: In the future, it could also be nice to have a debounce on the server to avoid transferring many events in a short span of a time (like when a directory with many files is trashed).

nono (approved):

core/remote/watcher/realtime_manager.js
taratatach: Yes it would be nice to be able to subscribe to grouped events.

gui/js/tray.js
taratatach: Well, this has been around since forever so I just did not touch it.

Reading it again, I think I'd prefer reusing this menu definition rather than removing it. I'll talk about it during our next grooming.

gui/js/network/agent.js
taratatach: No, these are type definitions.

PR #2324, opened 2023-10-03
fix: Process Chokidar events batches sequentially

taratatach:
We buffer events emitted by Chokidar as the underlying FSEvents API,
on macOS, can add delay between events relating to the same document
and processing the first events without the latter ones can lead to
wrong assumptions about what changes were actually made.

Buffered events are then flushed 10 seconds after the last event was
buffered (i.e. each buffered event resets the 10 seconds delay).

Since this process is asynchronous and is only time-based, we can end
up flushing buffered events while some previous batch is still being
processed, creating situations where unexpected behavior can happen.

To prevent this, we now prevent any buffer flushes while a batch is
being processed and restore the flushing timeout once finished.
This can add a small delay to the next batch processing but since we
don't know if already buffered events could be flushed at that point
it seems safer.

- [x] PR is not too big

### DIFF
--- a/.github/actions/setup-dnsmasq/action.yaml
+++ b/.github/actions/setup-dnsmasq/action.yaml
@@ -1,0 +1,23 @@
+name: Setup cozy-stack
+author: Erwan Guyader
+description: Setup dnsmasq for .localhost domains on macOS
+runs:
+  using: composite
+  steps:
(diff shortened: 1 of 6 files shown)


taratatach:
Closing in favor of #2325.

PR #2350, opened 2024-10-21
feat: Disabling popover mode

sebn:
Recent GNOME versions don't have a systray anymore.

Users can use this extension to bring the systray icon back: https://extensions.gnome.org/extension/615/appindicator-support/

But instead of showing the Cozy tray window on click, the extension shows an intermediate menu with 2 entries instead:
- Show application (which actually toggles its visibility somehow?)
- Quit application (app gets a SIGSEGV signal)

Thus the current tray window behavior feels neither as a proper popover (needs 2 clicks to show up) nor as a proper window (no decorations / resizing / moving / mazimizing).

Here we introduce a new GUI configuration option named `popoverMode`, which is assumed to be enabled by default, unless explicitly disabled (i.e. set to false).

When disabled:
- The window has proper decorations, allowing it to be moved, resized, maximized & closed.
- The windows is hidden on close, same as on blur when the option is disabled, unless the application is actually exiting.

The new option replaces & improves the old visibleOnBlur option that was requested by some tiling window manager user(s).

- [x] PR is not too big
- [ ] it includes unit tests matching the implementation changes => there were no unit tests for this part of the code, I manually checked that I could show/hide/quit the app multiple time, with & without the option, but please note I didn't test it myself on Windows nor macOS.

### DIFF
--- a/doc/usage/linux.md
+++ b/doc/usage/linux.md
@@ -154,7 +154,8 @@ effects so you need to do a right-click on the icon then select the
 >
 > By default, the window behaves like a system tray / popover menu that appears
 > below the Cozy icon / app indicator, same as Windows & macOS. But it doesn't
-> work so well with recent GNOME versions or with tiling Window managers.
+> work so well with the GNOME appindicator extension (see above) or with tiling
+> window managers.
 >
 > You can set `"gui": {"popoverMode": false}` in your
 > `~/.cozy-desktop/config.json` to force the window to behave like a normal one
--- a/doc/usage/linux.md
+++ b/doc/usage/linux.md
@@ -150,11 +150,15 @@ effects so you need to do a right-click on the icon then select the
 > extension providing support for appindicator. You can find and install one
 > from https://extensions.gnome.org/ by searching for `appindicator`.
 
-> **Note for i3wm Users**
+> **Disabling the "popover" mode**
 >
-> You can set `"gui": {"visibleOnBlur": true}` in your
-> `~/.cozy-desktop/config.json` so the popover doesn't hide when focusing
-> another window.
+> By default, the window behaves like a system tray / popover menu that appears
+> below the Cozy icon / app indicator, same as Windows & macOS. But it doesn't
+> work so well with recent GNOME versions or with tiling Window managers.
+>
+> You can set `"gui": {"popoverMode": false}` in your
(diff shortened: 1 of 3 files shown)


doc/usage/linux.md
```diff
@@ -150,11 +150,15 @@ effects so you need to do a right-click on the icon then select the
...
 
-> **Note for i3wm Users**
+> **Disabling the "popover" mode**
 >
-> You can set `"gui": {"visibleOnBlur": true}` in your
-> `~/.cozy-desktop/config.json` so the popover doesn't hide when focusing
-> another window.
+> By default, the window behaves like a system tray / popover menu that appears
+> below the Cozy icon / app indicator, same as Windows & macOS. But it doesn't
+> work so well with recent GNOME versions or with tiling Window managers.
```
sebn, line 157:
```suggestion
> work so well with the GNOME appindicator extension (see above) or with tiling
> Window managers.
```

sebn:
@taratatach Looks like I made flow unhappy... My bad, I can fix it.

In the meantime, in case you're not convinced that's the right approach, or the timing is pretty bad, feel free to just close the PR for now, I know you're probably lacking the time to review it & I don't want to add too much on top of all your work in progress 🙈 

To be honest, even I have doubts about it...

I wonder whether implementing the initial popover design without challenging it was a good idea (not blaming anybody, I was working on it myself anyway). What I mean is that I don't think relying on electron to "simulate" a popover is a robust approach: it's hard to make it behave exactly like a native tray menu so in the end it feels "alien" on every desktop environment. Also it could break anytime depending on some desktop environment changes.

Maybe we could have used a separated native tray menu on each OS + a common electron window-based UI that would show up from some of the menu entry. But that would mean having a way for the tray menu to communicate with the sync core (e.g. to show the current status).

taratatach:
@sebn Hey! Sorry about the lack of communication. I was indeed trying to get the next version out with fixes for a few regressions from the last one.

I like the idea of disabling the popover, especially if it can be re-enabled via the config! I'd like to do a complete redesign of the Desktop interface to get a more modern one, with more features in the future and disabling the popover is part of that.

So thanks for the PR and don't hesitate to bring it back, especially if you can fix the flow errors :wink:.
Otherwise I'll try to reuse your code once I get the time.

PR #2366, opened 2024-12-12
deps: Upgrade electron to v33.2.1

taratatach:

- [ ] PR is not too big


ArchangeGabriel:
I guess you know that already, but 32 is going to be EOL on March 4th per https://www.electronjs.org/docs/latest/tutorial/electron-timelines, getting up to 34 that was released two weeks ago would be better I think.

PR #2407, opened 2025-09-01
fix: Use native APIs to manage auto-launch

### LINKED ISSUE #2460: On recent macOS versions, auto-launch can't be enabled
Our auto-launch management relies on a python script that can't be launched anymore because the python runner has moved in recent macOS versions.
The API the script uses might have changed as well.

taratatach:
## Summary

- manage auto-launch with Electron's native login-item APIs on macOS and Windows
- keep the existing `auto-launch` backend for Linux and AppImage packages
- preserve the public `isEnabled()` / `setEnabled()` contract used by onboarding and tray code
- retain the stable AppImage launcher name, hidden start, and desktop-entry path refresh
- add unit coverage for native status, approval, updates, failures, Linux/AppImage behavior, and both onboarding registration paths

## Root cause

The legacy `auto-launch` package manages macOS login items through AppleScript. This is no longer reliable on recent macOS versions, while the Electron version used by Desktop already exposes the native login-item APIs backed by macOS Service Management.

Fixes #2460

### DIFF
--- a/gui/js/autolaunch.js
+++ b/gui/js/autolaunch.js
@@ -4,67 +4,141 @@
  */
 
 const AutoLaunch = require('auto-launch')
+const { app } = require('electron')
 
 const log = require('../../core/app').logger({
   component: 'GUI'
 })
 
 const APP_NAME = 'Twake-Desktop'
-const opts = {
-  name: APP_NAME,
-  isHidden: true
+const NATIVE_PLATFORMS = new Set(['darwin', 'win32'])
+
+class LinuxAutoLauncher {
+  constructor({ AutoLaunchClass, appImagePath, logger }) {
+    const opts = {
+      name: APP_NAME,
+      isHidden: true
+    }
+
+    if (appImagePath) opts.path = appImagePath
+
+    this.autoLauncher = new AutoLaunchClass(opts)
+
+    if (appImagePath) {
+      // `auto-launch` replaces the configured name with the AppImage file name
+      // when a path is provided. Keep a stable name so updates do not change
+      // the autolaunch entry.
+      this.autoLauncher.opts.appName = APP_NAME
+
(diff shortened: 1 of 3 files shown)


gui/js/autolaunch.js
```diff
@@ -4,67 +4,141 @@
...
+      // `auto-launch` replaces the configured name with the AppImage file name
+      // when a path is provided. Keep a stable name so updates do not change
+      // the autolaunch entry.
+      this.autoLauncher.opts.appName = APP_NAME
+
+      // Refresh the desktop entry so its Exec path follows the current
+      // AppImage after an update or move.
+      this.ready = this.refresh().catch(err =>
+        logger.error('could not refresh AppImage autolaunch entry', { err })
+      )
```
taratatach, line 37: This is probably not necessary anymore since the AppImage filename doesn't include the version anymore thus the exec path does not change during updates anymore.

gui/js/autolaunch.js
```diff
@@ -4,67 +4,141 @@
...
+
+    return enabled
+  }
 }
 
-if (process.env.APPIMAGE) {
-  opts.path = process.env.APPIMAGE
+class NativeAutoLauncher {
+  constructor({ electronApp, logger, platform }) {
+    this.app = electronApp
```
taratatach, line 70: This seems unnecessary since we can (and do) import `app` in this module.
We can just simply call `app.getLoginItemSettings()` and such.

taratatach: Ah I see it's used to mock `app` in tests.

gui/js/autolaunch.js
taratatach: Might still be useful if the AppImage is moved though. Let's keep it.

PR #2448, opened 2026-07-06
fix: Show OAuth onboarding errors

shepilov:
## Summary
- surface OAuth callback and registration failures on the onboarding OAuth page
- stop the onboarding flow from continuing to sync configuration when OAuth registration fails
- add localized copy for the OAuth failure message across the shipped locale files

### DIFF
--- a/gui/locales/en.json
+++ b/gui/locales/en.json
@@ -217,6 +217,7 @@
   "OAuth Waiting for login": "Waiting for login",
   "OAuth Please check your default browser and log in your Twake account.": "Please check your default browser and log in your Twake account.",
   "OAuth You'll be redirected here once it's done.":"You'll be redirected here once it's done.",
+  "OAuth Could not login": "We could not complete the login. Please retry, or check that this staging account is linked to this Twake Workplace.",
   "OAuth If something went wrong, you can retry the login by clicking on the button below.":"If something went wrong, you can retry the login by clicking on the button below.",
   "OAuth Retry":"Retry",
   "OAuth Contact support":"Contact support",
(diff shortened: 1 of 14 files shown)


gui/locales/en.json
```diff
@@ -217,6 +217,7 @@
   "OAuth Waiting for login": "Waiting for login",
   "OAuth Please check your default browser and log in your Twake account.": "Please check your default browser and log in your Twake account.",
   "OAuth You'll be redirected here once it's done.":"You'll be redirected here once it's done.",
+  "OAuth Could not login": "We could not complete the login. Please retry, or check that this staging account is linked to this Twake Workplace.",
```
taratatach:
```suggestion
  "OAuth Could not login": "We could not complete the login. Please retry, or check that a Twake Workplace is associated with your account.",
```

This should be added to the other locales as well (I'll make a separate suggestion for French).

gui/locales/fr.json
```diff
@@ -217,6 +217,7 @@
   "OAuth Waiting for login": "En attente de connexion",
   "OAuth Please check your default browser and log in your Twake account.": "La page de connexion à votre Twake Workplace s'est ouverte dans votre navigateur.",
   "OAuth You'll be redirected here once it's done.":"Vous serez redirigé ici une fois connecté à votre workplace.",
+  "OAuth Could not login": "Nous n'avons pas pu terminer la connexion. Veuillez réessayer ou vérifier que ce compte de staging est rattaché à ce Twake Workplace.",
```
taratatach:
```suggestion
  "OAuth Could not login": "Nous n'avons pas pu terminer la connexion. Veuillez réessayer ou vérifier que ce compte est associé à une Twake Workplace.",
```

taratatach:
We need to add the new key to every locale (especially `es`). We can use the English value since we don't have any translator.

PR #2452, opened 2026-07-07
test: Avoid APFS integration timeout

shepilov:
## Initial problem

The APFS integration job was timing out in `test/integration/differential_sync.js` for the scenario where a local folder is created with the same path as an excluded remote folder and the user chooses to create a conflict. The failing test waited for `local-end` only after `helpers.local.side.start()` had resolved, but `start()` already waits for the initial scan to finish. The watcher emits `local-end` before resolving that `start()` promise, so the test could attach its listener too late and then wait forever until Mocha hit the 20s timeout.

The same race also existed after creating `Photos/` and `Photos/My Image.png`: the listener was registered after the filesystem writes, so a fast watcher could flush the events before the test started waiting.

## Changes

- Remove the redundant `local-end` wait after `helpers.local.side.start()`.
- Use a one-shot `once('local-end', ...)` listener.
- Register the local watcher completion promise before creating the local folder and file.

## Validation

- `./node_modules/.bin/eslint test/integration/differential_sync.js`
- `git diff --check origin/master...HEAD`

The targeted APFS integration test was not run locally because this checkout has no `.env.test`, and no local Cozy stack is running on `cozy.localhost:8080` / `127.0.0.1:8080`.

### DIFF
--- a/test/integration/add.js
+++ b/test/integration/add.js
@@ -323,7 +323,7 @@ describe('Add', () => {
       })
 
       context('and the directory is updated after its content is added', () => {
-        it('creates the directory and its content on the remote Cozy without errors', async () => {
+        it('creates the directory and its content on the remote Cozy without errors', async function() {
           // Create directory and its content
           const dir = await createDoc('local', 'dir', parent)
           const subdir = await createDoc('local', 'subdir', dir)
--- a/test/integration/move.js
+++ b/test/integration/move.js
@@ -7,6 +7,7 @@ const _ = require('lodash')
 const should = require('should')
 
 const { TRASH_DIR_ID } = require('../../core/remote/constants')
+const { MAX_SYNC_RETRIES } = require('../../core/sync')
 const { logger } = require('../../core/utils/logger')
 const Builders = require('../support/builders')
 const dbBuilders = require('../support/builders/db')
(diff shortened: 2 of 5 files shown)


test/integration/move.js
```diff
@@ -1158,6 +1159,19 @@ describe('Move', () => {
...
+    const exhaustMissingDocumentRetries = async dirPath => {
+      const docs = [
+        await helpers.docByPath(dirPath),
+        ...(await pouch.byRecursivePath(dirPath))
+      ]
+
+      await pouch.bulkDocs(
+        docs.map(doc => ({
+          ...doc,
+          errors: MAX_SYNC_RETRIES
```
taratatach: I'd suggest setting the counter to `MAX_SYNC_RETRIES - 1` to keep testing the retry behavior (while still speeding up the tests).

test/integration/add.js
```diff
@@ -323,7 +323,9 @@ describe('Add', () => {
       })
 
       context('and the directory is updated after its content is added', () => {
-        it('creates the directory and its content on the remote Cozy without errors', async () => {
+        it('creates the directory and its content on the remote Cozy without errors', async function() {
+          this.timeout(60000)
```
taratatach: This is a very long timeout. We should never have to wait that long.

shepilov:
```suggestion
```

test/integration/move.js
shepilov:
```suggestion
          errors: MAX_SYNC_RETRIES - 1
```

taratatach (approved):

PR #2453, opened 2026-07-07
test: Normalize HFS+ scenario paths

shepilov:
## Initial problem

The macOS HFS+ scenario job was failing on Unicode path assertions for remote-created paths such as:

- `test/scenarios/create_accentuated_file_in_accentued_dir/with_different_encodings/remote/`
- `test/scenarios/create_dir_with_utf8_character/remote/`

Those fixtures already define the expected HFS+ local paths with `String#normalize('NFD')`. If the HFS+ job is meant to verify filesystem behavior, the actual local filesystem tree must stay raw: normalizing it during assertion would only prove that JavaScript normalization works and could hide a local filesystem/helper mismatch.

## Changes

- Keep actual local scenario tree paths raw before comparing them with expected values.
- Add an HFS+-only assertion that every raw local path returned by `helpers.local.treeWithoutTrash()` is already NFD.
- Include Unicode code points in the HFS+ assertion message so NFC/NFD failures are readable in CI logs.
- Keep remote tree assertions strict, so remote filename spelling remains byte-for-byte checked.

## Validation

- `./node_modules/.bin/eslint test/scenarios/run.js`
- `yarn lint:flow`
- `git diff --check`

The HFS+ scenario job was not run locally because this checkout has no `.env.test`, and no local Cozy stack is running on `cozy.localhost:8080` / `127.0.0.1:8080`.

### DIFF
--- a/test/scenarios/run.js
+++ b/test/scenarios/run.js
@@ -464,12 +464,17 @@ async function verifyExpectations(
               fileid: pouchItem && pouchItem.local && pouchItem.local.fileid // undefined if not running on Windows
             }
       })
-      actual.localTree =
-        process.env.COZY_DESKTOP_FS === 'HFS+'
-          ? localTree.map(item =>
-              Object.assign({}, item, { path: item.path.normalize('NFD') })
-            )
-          : localTree
+      if (process.env.COZY_DESKTOP_FS === 'HFS+') {
+        localTree.forEach(({ path: fpath }) => {
+          fpath.should.equal(
+            fpath.normalize('NFD'),
+            `Expected HFS+ local path to be NFD: ${formatPathForUnicodeDebug(
+              fpath
+            )}`
+          )
+        })
+      }
+      actual.localTree = localTree
     }
     if (expectedRemoteTree) {
       // TODO: fetch tree with id and rev to compare them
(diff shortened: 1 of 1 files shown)


taratatach:
I'm definitely not satisfied with the AI's explanation for the problem. To me, on HFS+, paths are NFD encoded and that's why our expectations are normalized to NFD. By also normalizing the paths we get from the FS, we're just checking that the normalization function does its job.

shepilov:
> I'm definitely not satisfied with the AI's explanation for the problem. To me, on HFS+, paths are NFD encoded and that's why our expectations are normalized to NFD. By also normalizing the paths we get from the FS, we're just checking that the normalization function does its job.

nope, actually we are checking that sync function did therirs jobs

but yes, if we need to verify HFS+ filesystem then yes, it doesn't make sense

should HFS+ store names decomposed?  - Yes.
can our test assume fs.readdir() returns an NFD JavaScript string?  - Apparently no, not on this CI runner.

shepilov:
- COZY_DESKTOP_FS: HFS+
  - mounted as Apple_HFS
  - job name: Scenarios (HFS+)

  The debug assertion shows the important part: raw local paths returned by fs.readdir are not always NFD in this CI
  environment.

  Examples from the log:

  - Partages reçus/ contains U+00E7 (ç, NFC), while NFD would be c + U+0327
  - AccuséRéception.pdf contains U+00E9 twice (é, NFC), while NFD would be e + U+0301

taratatach:
So, we need to fix the CI environment. Otherwise these HFS+ jobs have no value.

shepilov:
As far as all the test suit that we are running for HFS+ is to test file sysstem behavious, we shouldn't mock it, and the integration tests and other scenarious we can check with other FS if they are fine

taratatach:
Replaced by https://github.com/cozy-labs/cozy-desktop/pull/2474

PR #2454, opened 2026-07-07
fix: Order child deletion before parent move

### LINKED ISSUE #2450: macOS stopped-client scenario creates conflict when a moved directory contains a replaced child file
### Problem

  The stopped-client scenario `test/scenarios/move_dir_and_replace_subfile/local/stopped` fails on macOS APFS/HFS+.

  The scenario starts with:

  - `src/`
  - `src/file` containing `initial content`

  While Twake Desktop is stopped, the user effectively does:

  - move `src` to `dst`
  - replace `dst/file` with a new file containing `new content`

  Expected final state:

  - `dst/file` exists locally and remotely
  - `dst/file` contains `new content`
  - the old `file` is in the remote trash
  - no conflict file is created

  Actual CI result:

  - `dst/file` is missing locally/remotely
  - `dst/file-conflict-...` is created instead
  - remote read for `dst/file` returns `404`

 ### Root Cause

  This is not just one folder move. Semantically, the stopped-client changes are:

  1. old child identity `src/file` was deleted/replaced
  2. parent folder moved from `src` to `dst`
  3. new child identity was created at `dst/file`

  Sync currently has dependency rules for some parent/child operations, but it does not enforce:

  ```text
  DEL(src/file) before MOVE(src -> dst)
```

  So the remote parent move can move the old child into dst/file before the new file is uploaded. Then the new upload sees an
  active file at the same path and is resolved as dst/file-conflict-....

  ### Proposed Fix

  Add a sync dependency rule in compareChanges:

  When two changes are on the same side:

  - one change is DEL
  - the other is a directory MOVE
  - the deleted document path is inside the move source path

  then the child deletion must be synchronized before the parent move.

  Example:

  DEL(src/file) -> MOVE(src -> dst) -> ADD(dst/file)

  This keeps the local dst/file untouched. The deletion is metadata/remote-id based: it trashes the old remote child identity,
  not the current local file.

  ### Tests

  Add unit coverage for compareChanges:

  - DEL(src/file) is ordered before MOVE(src -> dst)
  - symmetric argument order returns the opposite result

  Add/keep regression coverage proving the full order:

  ADD(dst/file), MOVE(src -> dst), DEL(src/file)

  is applied as:

  DEL(src/file), MOVE(src -> dst), ADD(dst/file)

### PR DESCRIPTION
**Pull request by shepilov**
## Problem

Fixes #2450.

The stopped-client scenario `test/scenarios/move_dir_and_replace_subfile/local/stopped` can create `dst/file-conflict-...` instead of preserving the replaced child at `dst/file`.

The scenario starts with `src/` and `src/file`. While Twake Desktop is stopped, the user moves `src` to `dst`, replaces `dst/file` with a new file, then starts sync. The final state should keep the new child as `dst/file` and trash the old child identity.

## Root cause

The sync dependency graph knew how to order some parent/child moves and additions, but it did not express that a deleted child from a directory move source must be applied before the parent directory move.

Without that dependency, sync could move the old remote child from `src/file` to `dst/file` as part of the parent directory move. The later upload of the new `dst/file` then saw an active file at the same path and was resolved as a conflict.

## Changes

- Add a same-side dependency rule in `compareChanges`: when a child deletion is inside a directory move source path, the deletion must run before the parent move.
- Add unit coverage for both argument orders.
- Add dependency-graph coverage for the full problematic order: `ADD(dst/file), MOVE(src -> dst), DEL(src/file)` is expanded as `DEL(src/file), MOVE(src -> dst), ADD(dst/file)`.

### DIFF
--- a/core/sync/index.js
+++ b/core/sync/index.js
@@ -170,22 +170,42 @@ const compareChanges = (
       else if (docA.path.startsWith(docB.path + sep)) return B_FIRST
     }
 
-    if (opA.type === 'DEL' && opB.type === 'MOVE' && docB.moveFrom != null) {
-      // Handle parent deletion after child move outside of parent
-      if (
-        docB.moveFrom.path.startsWith(docA.path + sep) &&
-        !docB.path.startsWith(docA.path + sep)
-      ) {
-        return B_FIRST
+    if (opA.type === 'DEL' && opB.type === 'MOVE') {
+      const moveFromB = docB.moveFrom
+      if (moveFromB != null) {
+        // Handle parent deletion after child move outside of parent
+        if (
+          moveFromB.path.startsWith(docA.path + sep) &&
+          !docB.path.startsWith(docA.path + sep)
+        ) {
+          return B_FIRST
+        }
+        // Handle child deletion before parent move
+        if (
+          metadata.isFolder(docB) &&
+          docA.path.startsWith(moveFromB.path + sep)
+        ) {
+          return A_FIRST
+        }
       }
     }
-    if (opB.type === 'DEL' && opA.type === 'MOVE' && docA.moveFrom != null) {
-      // Handle parent deletion after child move outside of parent
-      if (
-        docA.moveFrom.path.startsWith(docB.path + sep) &&
-        !docA.path.startsWith(docB.path + sep)
-      )
-        return A_FIRST
+    if (opB.type === 'DEL' && opA.type === 'MOVE') {
+      const moveFromA = docA.moveFrom
+      if (moveFromA != null) {
+        // Handle parent deletion after child move outside of parent
+        if (
+          moveFromA.path.startsWith(docB.path + sep) &&
+          !docA.path.startsWith(docB.path + sep)
+        )
+          return A_FIRST
+        // Handle child deletion before parent move
+        if (
+          metadata.isFolder(docA) &&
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: core/sync/index.js [hunk 1]
```diff
@@ -170,22 +170,42 @@ const compareChanges = (
...
+      const moveFromB = docB.moveFrom
+      if (moveFromB != null) {
+        // Handle parent deletion after child move outside of parent
+        if (
+          moveFromB.path.startsWith(docA.path + sep) &&
+          !docB.path.startsWith(docA.path + sep)
+        ) {
+          return B_FIRST
+        }
+        // Handle replaced child deletion before parent move
```
shepilov:
```suggestion
        // Handle child deletion before parent move
```

core/sync/index.js
```diff
@@ -170,22 +170,42 @@ const compareChanges = (
...
+    if (opB.type === 'DEL' && opA.type === 'MOVE') {
+      const moveFromA = docA.moveFrom
+      if (moveFromA != null) {
+        // Handle parent deletion after child move outside of parent
+        if (
+          moveFromA.path.startsWith(docB.path + sep) &&
+          !docA.path.startsWith(docB.path + sep)
+        )
+          return A_FIRST
+        // Handle replaced child deletion before parent move
```
shepilov:
```suggestion
        // Handle child deletion before parent move
```

taratatach (approved):

PR #2455, opened 2026-07-07
feat: Skip dependants of skipped changes

### LINKED ISSUE #2436: Alert user when changes are skipped during sync
Sometimes a change cannot be synchronized at all and we must skip it to unlock the synchronization of other changes.
When this is done automatically, we should alert the user so they're aware these changes were skipped.

taratatach:
A skipped change left its dependants unprotected: on the next cycle
  they would be attempted and either fail spuriously (e.g.
  `MISSING_PARENT`) or, worse, succeed partially and create an
  orphan. The multi-error loop makes this more visible since
  dependants are no longer shielded by the first-failure break.

  When a change is skipped (automatically via `handleSyncError`
  returning `'skipped'`, or by the user), the `syncBatch` loop now
  adds it to `blockedIds`. Dependents detected via
  `directPrerequisites` whose prereq has `doc.skipped === true` are
  themselves skipped (`skipChange` persists the flag) and emit a
  synthetic `SKIPPED_DEPENDENCY` alert. `blockedIds` propagates
  the skip transitively across the topologically ordered batch.

  The synthetic `SyncError` carries `prereqPath` as a dedicated
  field, transported by `makeAlert` to the GUI. `viewByCode`
  interpolates it into a localized message with the path wrapped
  in backticks. The restored `Util.DecorationParser` parses those
  backticks in `alertContent`, and `decoratedName` renders the path
  as a clickable chip (blue background, `title` with full path,
  `ShowInParent` on click).

  Fixes #2436 

- [x] PR is not too big

### DIFF
--- a/core/metadata.js
+++ b/core/metadata.js
@@ -166,7 +166,7 @@ export type Metadata = {
 
   trashed?: true,
   errors?: number,
-  skipped?: boolean,
+  skipped?: string,
   overwrite?: SavedMetadata,
   childMove?: boolean,
   incompatibilities?: *,
--- a/core/migrations/migrations.js
+++ b/core/migrations/migrations.js
@@ -409,5 +409,24 @@ module.exports = ([
         { concurrency: 10 }
       )
     }
+  },
+  {
+    baseSchemaVersion: 14,
+    targetSchemaVersion: 15,
+    description: 'Convert legacy skipped:true boolean to UserSkipped code',
+    affectedDocs: (docs /*: SavedMetadata[] */) /*: SavedMetadata[] */ => {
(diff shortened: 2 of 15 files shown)


core/metadata.js
```diff
@@ -166,7 +166,7 @@ export type Metadata = {
 
   trashed?: true,
   errors?: number,
-  skipped?: boolean,
```
shepilov, line 169: Will it work for the existing db? or it's not stored in the db?

taratatach: Ah crap, you're right, the introduction of `skipped` has already been released… I'll add a migration.

PR #2459, opened 2026-07-15
fix: Focus onboarding after SSO login

### LINKED ISSUE #2451: Focus the onboarding window when we receive OAuth credentials from the SSO
During an onboarding with an external SSO, we open the login page within the user's default browser. Once the login is done, the browser will try to open a special `cozy://` URL with the OAuth credentials as parameter with Desktop so that it can get them and finish the onboarding.
We should try to focus the onboarding window when we receive this URL so the user sees the login process worked and the onboarding can continue.

shepilov:
## Summary

- focus the onboarding window as soon as the SSO OAuth callback returns
- restore the window first when it is minimized
- add regression coverage for focus behavior and OAuth callback ordering

## Root cause

Desktop handled the `cozy://` callback and continued onboarding, but it did not bring the onboarding window back to the foreground after login in the user's browser.

Closes #2451

### DIFF
--- a/test/unit/gui/window_manager.js
+++ b/test/unit/gui/window_manager.js
@@ -1,0 +1,48 @@
+const sinon = require('sinon')
+
+const WindowManager = require('../../../gui/js/window_manager')
+
+describe('window_manager', () => {
+  describe('focus', () => {
+    const sandbox = sinon.createSandbox()
+    let windowManager
+    let win
+
+    beforeEach(() => {
+      win = {
+        isMinimized: sandbox.stub(),
+        restore: sandbox.spy(),
+        focus: sandbox.spy()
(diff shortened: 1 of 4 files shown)


test/unit/gui/window_manager.js
```diff
@@ -0,0 +1,48 @@
...
+      windowManager.win = win
+    })
+
+    afterEach(() => sandbox.restore())
+
+    it('does nothing when the window is closed', () => {
+      windowManager.win = null
+
+      windowManager.focus()
+    })
```
taratatach, line 31: There aren't any assertions in this test.

taratatach (approved):
LGTM. You can squash 👍

